PR #53, opened 2014-02-04
Add linenumbers to gui

vonloxley:
ln

### DIFF
--- a/languagetool-core/src/main/resources/org/languagetool/MessagesBundle.properties
+++ b/languagetool-core/src/main/resources/org/languagetool/MessagesBundle.properties
@@ -207,6 +207,8 @@ guiTagText = &Tag Text
 
 guiExamples = Examples:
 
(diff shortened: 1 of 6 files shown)


johnblommers:
IMHO languagetool needs the feature to display a line number column. Expecialy since the bottom pane refers to line numbers. The current implementaiton is incomplete in my view.

Indeed to be practial one should be able to scan the bottom pane, identify a diagnostic, and click  there to be sent to the offending text in context.

The wiggly underlines OTOH are spot on.

danielnaber:
Just checked the license of `TextLineNumber.java` to make sure it's okay, and it is: "We assume no responsibility for the code. You are free to use and/or modify and/or distribute any or all code posted on the Java Tips Weblog without restriction. A credit in the code comments would be nice, but not in any way mandatory.", source: https://tips4java.wordpress.com/contact-us/

PR #260, opened 2015-05-14
Bad prolonged soud mark

knok:
A prolounged soud mark "ー" is placed before Katakana/Hiragana character.
This rule comes from the following discussion:
https://sourceforge.net/p/languagetool/mailman/message/34107974/


danielnaber:
Thanks, I've merged it but then noticed that it also finds an error in `隣家ー`, i.e. the test `<example><marker>隣家</marker>ー</example>` fails. Could you have a look at that?

knok:
I made a new request as #261.

PR #264, opened 2015-05-18
support usage of "A" as an identifier

lindt:
- support for A as an identifier
   - issue was, that the "A" in the term "station A" has been seen as an indefinite article.
 - it has been recognized before as a "A_VS_AN"-A
 - added test
 - cleaned up matcher


danielnaber:
Thanks!

PR #271, opened 2015-05-23
Update grammar.xml

Mility:


Mility:
I think keep health is very important. you can do these things as more as possible every day.

danielclipnow:
I don't think "as more as possible" is correct, see e.g. see http://forum.wordreference.com/threads/as-more-as-possible.1912477/

danielnaber:
I don't think "as more as possible" is correct, see e.g. see http://forum.wordreference.com/threads/as-more-as-possible.1912477/

Mility:
Collect and master idioms as more as you can. Learning English idioms is one of the difficulties.

danielnaber:
I think it should be "as much as you can", but a native speaker should comment on that.

Mility:
Thanks so much.

Issue #280, opened 2015-06-26
Support for using .langaugetool.cfg with languagetool-server.jar


tingley:
Issue corresponding to [this thread](http://languagetool-user-forum.2306527.n4.nabble.com/Using-langaugetool-cfg-with-languagetool-server-jar-tc4642928.html) on the forums.
> I'm attempting to deploy LanguageTool in server mode for use by multiple users, with a single, centralized configuration.  I've done this successfully by configuring rules for multiple languages through the UI to create a ~/.languagetool.cfg, and then starting the server from the UI. 
>
> This works great, except for one thing: it requires the UI.  I'd like to do the same thing using languagetool-server.jar, so that I can wrap LT as a service and have it start automatically, etc. 
>
> However, when launched via the HTTPServer/HTTPSServer classes, it it looks like this configuration is ignored.  (The file is parsed, but ignored unless the internalServer flag is set -- see LanguageToolHttpHandler.java.) 

I have a patch for this, I'll open a pull request.

danielnaber:
Closing this issue, as the pull request has been merged.

PR #281, opened 2015-06-26
Add 'rulesFile' option to config properties

tingley:
First attempt at a fix for issue #280.

This adds a new option, `rulesFile`, to the server configuration file specified with `--config`.  The `rulesFile` option points to the location of a .languagetool.cfg-style rules file.

This is a little odd (the config file now points to another config file), but my thinking was that it was better to keep the server options consolidated in the `--config` file, rather than adding a new top-level switch.  I am also allowing the `rulesFile` param to point to an arbitrary file location, since that's more convenient for server deployment scenarios where we don't want to depend on things like the location of the home directory of the "user" running the service.

I'm happy to rework any portion of this if you'd like it to work a different way.


danielnaber:
Thanks, looks good. I'll merge in a few days once we're out of feature freeze for LT 3.0.

PR #302, opened 2015-08-09
Update grammar.xml

rg-software:
The proposed change includes 700+ rules for the Japanese language.

### DIFF
--- a/languagetool-language-modules/ja/src/main/resources/org/languagetool/rules/ja/grammar.xml
+++ b/languagetool-language-modules/ja/src/main/resources/org/languagetool/rules/ja/grammar.xml
@@ -10,35 +10,622 @@ Japanese Grammar and Typo Rules file for LanguageTool
 	<!-- 文法 -->
 	<!-- ============================== -->
 	<category name="文法">
-		<rule id="DOUSI_WO" name="動詞+を">
+		<rule id="SITAGATTE_END" name="したがって_END">
+			<pattern case_sensitive="no">
+				<token>したがって</token>
+				<token postag="SENT_END" />
+			</pattern>
+			<message>したがっては文の最後にきません</message>
+			<example type="correct"><marker>したがって</marker></example>
+			<example type="incorrect"><marker>したがって。</marker></example>
+		</rule>		
+		<rule id="KAIKETUNARU" name="解決なる">
+			<pattern case_sensitive="no">
+				<token>解決</token>
+				<token>なる</token>
+			</pattern>
+			<message>文法ミスがあります。<suggestion>解決になる</suggestion>の間違いです。</message>
+			<example type="correct"><marker>解決になる</marker></example>
+			<example type="incorrect"><marker>解決なる</marker></example>
+		</rule>		
+        <rule id="DOUSI_DEHA" name="動詞+では">
 			<pattern case_sensitive="no">
 				<token postag="動詞.*" postag_regexp="yes"/>
-				<token>を</token>
+				<token>で</token>
+				<token>は</token>
 			</pattern>
-			<message>のを</message>
-			<example><marker>見るのを</marker>ためらう</example>
-			<example correction=""><marker>見るを</marker>ためらう</example>
-		</rule>
-		<rule id="ENAI" name="得ない">
+			<message>文法ミスがあります。<suggestion>のでは</suggestion>の間違いです。</message>
+			<example type="correct"><marker>描くのでは</marker></example>
+			<example type="incorrect"><marker>描くでは</marker></example>
+		</rule>			
+        <rule id="MEISI_DESUMU" name="名詞+で住む">
 			<pattern case_sensitive="no">
-				<token>ざる</token>
-				<token>お</token>
-				<token>得</token>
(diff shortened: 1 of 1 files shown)


Shugyousha:
Many thanks for the pull request! The Japanese rules are in need of attention from native speakers...

There are a lot of whitespace issues in the code (for example, there are lines like '+               </rule>         '. These should be fixed. Also, the tests do not pass.

On Linux you can test the Japanese language rules using the following command line.

bash build.sh ja clean test

This should run through without any errors before we can merge the changes.

Please do not hesitate to ask if we can help you fix the failing tests!

rg-software:
Well, yes, I would appreciate help, since the rules file is all I have, and I am not really involved into LT development of any kind, so I just want to let the new grammar file go through.

1) To be honest, I don't know how tabs make XML ill-formed, but ok, I can fix it.
2) I run "testrules ja" from the binary LT distribution, but the message I get is not very helpful:

Exception in thread "main" junit.framework.AssertionFailedError: Japanese: Incor
rect suggestions: [] != [?????] for rule KAIKETUNARU[1]:[??, ??]:???? on input:
???? expected:<[]> but was:<[?????]>

It's rather difficult to debug "????". How can I check the actual problem?

danielnaber:
You could try `testrules.sh ja >result.txt` and then see if `result.txt` still has the correct characters.

rg-software:
Re: [languagetool] Update grammar.xml (#302)



Hi,

Вы писали 10 августа 2015 г., 15:51:13:





You could try testrules.sh ja >result.txt and then see if result.txt still has the correct characters.



I tried that, but unfortunately this part of the output is not done to ordinary stdout; some text instead goes to the file, but this particular fragment is still printed on the console.

danielnaber:
Have you also tried `testrules.sh ja >result.txt 2>&1`?

rg-software:
OK, it worked -- everything is in the textfile now. Unfortunately, I still see the question marks instead of Japanese characters.

I think the rule file was tested (maybe not properly enough), but it seems that testrules script reports errors on nearly every rule, and there is no way to debug it... We actually faced some other troubles with Japanese characters, but let's proceed step by step.

janschreiber:
@rg-software I think the main reason for the failing tests is that your incorrect examples lack the ``correction`` attribute. Line 29 of grammar.xml should read
``<example correction="解決になる"><marker>解決なる</marker></example>``
The new syntax does not use ``type="incorrect"``anymore. Instead, the existence of the ``correction`` attribute implies that the example is incorrect. This probably accounts for most of the errors.
(I hope I didn't mess up the Japanese, I just copied and pasted it.)

rg-software:
Well, these rules were indeed composed some months ago using perhaps LT 2.x
So, what should I do? Just remove type="incorrect" everywhere?

janschreiber:
Yes, and additionally insert the content of ``<suggestion>CORRECT</suggestion>`` as the value of the correction attribute in the example:
``<example correction="CORRECT"><marker>WRONG</marker></example>``
If there is no suggestion, use an empty correction attribute like so:
``<example correction=""><marker>WRONG</marker></example>``

danielnaber:
You should also use an XML editor to see if the file is valid, there are some whitespace-like characters that make the document invalid in some places.

rg-software:
I see. What about correct parts? Should I write as
<example><marker>解決になる</marker></example> to provide correct examples?

OK, I will see what I can do in the following days/weeks. I wasn't ready to face such complications :)

janschreiber:
Examples that have no correction attribute are considered correct:
``<example>This is correct.</example>``
``<example correction="completely wrong">This is <marker>complete wrn0g</marker>.</example>``

janschreiber:
Also, here is an online XML checker by w3schools.com that you might find useful: http://www.w3schools.com/xml/xml_validator.asp

Shugyousha:
On Mon, Aug 10, 2015 at 02:40:26AM -0700, rg-software wrote:
> I see. What about correct parts? Should I write as
> <example><marker>解決になる</marker></example> to provide correct examples?
> 
> OK, I will see what I can do in the following days/weeks. I wasn't
> ready to face such complications :)

I have started to convert all rules and hopefully will be able to find
all remaining issues in the next few days.

I will try to send a pull request your way when I am done.


Cheers,

Silvan

rg-software:
Thank you very much! I am so sorry I cannot contribute these weeks -- very busy times!

S> On Mon, Aug 10, 2015 at 02:40:26AM -0700, rg-software wrote:
>> I see. What about correct parts? Should I write as
>> <example><marker>解決になる</marker></example> to provide correct examples?
>> 
>> OK, I will see what I can do in the following days/weeks. I wasn't
>> ready to face such complications :)
S>  
S> I have started to convert all rules and hopefully will be able to find
S> all remaining issues in the next few days.
S>  
S> I will try to send a pull request your way when I am done.
S>  
S>  
S> Cheers,
S>  
S> Silvan
S>  
S> —
S> Reply to this email directly or view it on GitHub.
S>   
S>

Shugyousha:
I opened the pull request here: https://github.com/rg-software/languagetool/pull/2

If it is merged these rules should be able to be incorporated here as well.

Thanks again for all the work!

PR #318, opened 2015-09-29
Update grammar.xml

Mility:
Update "inconsistency"


danielnaber:
Actually I prefer the way it is now because day of week and date don't match, but we don't know if the day of week is wrong or the date, so it makes sense to underline both.

PR #324, opened 2015-10-15
Update grammar.xml

Mility:
Add a rule:UNCOUNTABLE_PLURAL

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
@@ -8233,6 +8233,15 @@ USA
     longer list but not checked thouroughly: 
     http://www.geocities.com/Athens/Olympus/7583/ncntlist.html
     -->
+        <rule id="UNCOUNTABLE_PLURAL" name="uncountable coun(no plural)">
+	    <pattern>
+		 <token inflected='yes' regexp='yes' postag='NNS'>accommodation|advice|access|baggage|bread|damage|equipment|garbage|luggage|money|cattle|sand|furniture|meat|food|news|pasta|progress|research|water|freedom|maturity|intelligence|travel|pollution|traffic|transport|patience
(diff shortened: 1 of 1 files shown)


danielnaber:
I suggest you get confirmation from native speakers for your rules. For example, "conscious of the damages" is a very common phrase (see Google) and it doesn't sound wrong to me.

Mility:
See reference: Sidorov, G., Gupta, A., Tozer, M., Catala, D., Catena, A., & Fuentes, S. (2013). Rule-based System for Automatic Grammar Correction Using Syntactic N-grams for English Language Learning (L2). CoNLL-2013, 96. 
page 98.

danielnaber:
The rule would still give a false alarm for `She was awarded $284,000 in damages.` (see http://english.stackexchange.com/questions/29203/damage-vs-damages)

Mility:
Oh. Could we add this exception?

danielnaber:
This was just one example, maybe there are more. Have you tested the rule against Wikipedia and Tatoeba sentences?

Mility:
Sorry, For some reasons, I don't have those big data(Wikipedia and Tatoeba) and the hardware environment to test rule, I only test it in ruleEditor.

dpelle:
> For some reasons, I don't have those big data(Wikipedia and Tatoeba)

Tatoeba sentences can be downloaded with:

  $ wget http://downloads.tatoeba.org/exports/sentences.tar.bz2 
  $ tar jxvf sentences.tar.bz2 

It contains sentences in many languages.
You extract sentences in one language only (English in example below) with:

  $ perl -F'\t' -ane 'print "@F[2 .. $#F]" if ($F[1] eq "eng")' sentences.csv > sentences-eng.csv

It currently contains 565,318 sentences in English. You can truncate it if that's too much for you.

Mility:
Thank you very much.

howlinghuffy:
From my point of view as a native speaker, both 'damages' and 'freedoms' are legitimate in some cases

PR #326, opened 2015-10-21
Added ability to disable categories from server query

howlinghuffy:
This change allows you to use the "disabledCategories" parameter when quering the languagetool server to disable entire categories of rules for the check.


danielnaber:
Thanks, could you maybe also write a test case for this?

howlinghuffy:
For sure, I will do that.

Also, my modifications currently only apply for regular server checks, not for checks in the bitext mode.  Should the command work for the bitext too?

I haven't implemented it as yet, as it would involve tinkering with some of the languagetool-core components rather than just the languagetool-server code.

danielnaber:
I must say I'm not familiar enough with the bitext mode. If turning off rules also works for bitext mode, I'd guess turning off categories should work as well?

danielnaber:
Implemented, including an `enabledCategories` parameter (useful with `enabledOnly=yes`).

PR #340, opened 2015-12-04
New rules for 'licence' when used as a noun in en-GB.

capncodewash:
New category, and two new rules to catch instances of the word 'licence' when it is used as a noun and inadvertently spelled with an 's' instead of a 'c'.

Reference for spelling usage in en-GB:
http://www.oxforddictionaries.com/definition/english/licence
http://www.grammar-monster.com/easily_confused/licence_license.htm

Thanks,

Graeme


capncodewash:
P.S. tests pass.

PR #341, opened 2015-12-07
脉搏/膊

Mility:
脉搏/脉膊


danielnaber:
Please run the tests (`testrules.sh`/`testrules.bat`) first, with this rule I get: ` Incorrect suggestions: [] != [脉搏] for rule BO1_BO2[1] on input: 脉膊跳得平稳。 expected:<[]> but was:<[脉搏]>`

Mility:
I have test this rule in my standalone version and the online ruleEditor of LT and it's right.

danielnaber:
That's not enough, you need to run `testrules.sh` or `testrules.bat`, these scripts run more checks.

PR #351, opened 2015-12-30
Supply alternate classloader to languagetool

reckart:
I have a case where I would like to supply languagetool with a specific classloader through which to create Language instances and to load resources. Specifically, I want to download the language JAR at runtime, create a classloader that includes the auto-loaded JAR and supply that to languagetool (JLanguageTool, Languages).

This branch includes changes where I have dabbled in doing this.

The main problem I was facing is that access to languagetool is mostly static and auto-detecting languages is static. So presently, the changes allow setting a global alternative classloader. For multi-threading on different languages that would not work well. 

If you find the idea useful, we might try to elaborate it to specify a classloader when getting a language instance, e.g. something like 

```
Languages.getLanguageForShortName("de", classLoader)
```

The changes are not for merging yet - the pull request is only for discussion at this point in time.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/Language.java
+++ b/languagetool-core/src/main/java/org/languagetool/Language.java
@@ -335,6 +335,7 @@ public abstract class Language {
     if (patternRules == null) {
       List<AbstractPatternRule> rules = new ArrayList<>();
       PatternRuleLoader ruleLoader = new PatternRuleLoader();
+      ruleLoader.setClassLoader(this.getClass().getClassLoader());
       for (String fileName : getRuleFileNames()) {
         InputStream is = null;
         try {
--- a/languagetool-core/src/main/java/org/languagetool/Languages.java
+++ b/languagetool-core/src/main/java/org/languagetool/Languages.java
@@ -18,16 +18,24 @@
  */
 package org.languagetool;
 
-import org.apache.commons.lang.StringUtils;
-import org.jetbrains.annotations.Nullable;
-import org.languagetool.tools.MultiKeyProperties;
-import org.languagetool.tools.StringTools;
-
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Constructor;
 import java.net.URL;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+
+import org.apache.commons.lang.StringUtils;
+import org.jetbrains.annotations.Nullable;
+import org.languagetool.tools.MultiKeyProperties;
+import org.languagetool.tools.StringTools;
+import org.languagetool.tools.Tools;
 
 /**
  * Helper methods to list all supported languages and to get language objects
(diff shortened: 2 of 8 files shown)


danielnaber:
You don't have to use the `Languages` class to create a language I guess... can't you just create the language using its constructor, like with `new English()`, or maybe via reflection? LT then shouldn't care about how the object has been created?

reckart:
There are at least two problems I would face with such an approach:

* I'd have to replicate the discovery mechanism that languagetool uses to find the available language classes and to resolve the short language name to the class name
* I'd only instantiate a specific language through my special classloader, but the main LT JAR remains on my root classpath. That means that all classes in the main LT JAR including the DefaultResourceDataBroker, RuleFilterCreator, and Tools classes wouldn't be able to access resources through my special classloader as they use `Class.forName()` or `this.getClass().getXXX()` methods.

I could try to get around the second problem by adding the main LT jar also to my special classloader and  implementing a local-first instead of the default parent-first lookup strategy (maybe limited to the LT packages). I might try that if you are not interested in opening up LT for alternative classloaders. However, I thought it would be a better idea to first reach out to you to check if a proper support for custom classloaders in LT woudn't be the better option.

danielnaber:
I'm not sure if LT should deal with low level concepts like classloaders. I think your use case is quite special, maybe too special for LT to support.

reckart:
Even if you don't wish to have support for custom classloaders, it would at least be useful if classes like DefaultResourceDataBroker, RuleFilterCreator, and Tools used the loader from the respective Language instance for which they are invoked instead of using their own loaders. That would allow to implement the strategy you mention above (https://github.com/languagetool-org/languagetool/pull/351#issuecomment-168194194). Of course from there, it would only be a very small step to actually support custom classloaders.

danielnaber:
You mean like `myLang.getClass().getResourceAsStream()` instead of e.g. `Tools.class.getResourceAsStream()`? BTW, the behavior in `Tools` shouldn't affect you, it's only used for tests and development tools (there was one other case which I fixed recently).

reckart:
https://github.com/languagetool-org/languagetool/pull/351#issuecomment-168673155 : yes, exactly. Thus, if the language class is discovered and instantiated through a custom classloader, it would also load all its resources through the same loader. I'd still have to reimplement the discovery mechanism, on my side.

danielnaber:
For `RuleFilterCreator` it looks easy, but the data broker has many calls, they would all need to be changed to include the language. Would you provide a pull request for that? Also, see the comment at https://github.com/languagetool-org/languagetool/blob/master/languagetool-core/src/main/java/org/languagetool/tools/Tools.java#L269-L271, so we'd need to check Grails (used on community.languagetool.org) after such a change.

reckart:
```
    // the other ways to load the stream like
    // "Tools.class.getClass().getResourceAsStream(filename)"
    // don't work in a web context (using Grails):
```

That's to be expected. Tools.class.getClass() would use the system class loader through which the JDK is loaded (the one for java.lang.Class) of which the loader for the Grails application would be a child. The system class loader has no access to the classes of the grails application.

It may take a while until I find the time to get back to this, but I opened another issue tracking this on my side (https://github.com/dkpro/dkpro-core/issues/750).

danielnaber:
Closing to clean up the list of pull requests, feel free to re-open when/if work continues on this issue.

PR #375, opened 2016-03-17
Handle OSGi bundled dictionaries to copy to temporary directories

gavinbunney:
When using LanguageTool within an OSGi bundle, the dictionaries were failing to load. This is due to the same issues as when they are contained within a JAR file - the URI returned isn't a file URI so fails to load.

The fix here is to handle OSGi bundle loading (protocol of `bundle`) the same as JAR files. Tests out ok :ok_hand:

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/HunspellRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/HunspellRule.java
@@ -210,9 +210,9 @@ public class HunspellRule extends SpellingCheckRule {
 
     final URL dictURL = JLanguageTool.getDataBroker().getFromResourceDirAsUrl(originalPath);
     String dictionaryPath;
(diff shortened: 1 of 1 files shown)


iuscl-ide:
This still doesn’t work for me, the prefix “bundle” was added, but on French language for example, Eclipse finds the dictionary with the protocol “bundleresource” and it doesn’t work
The correct code could be:
``` java
if ("jar".equals(dictURL.getProtocol()) || "vfs".equals(dictURL.getProtocol()) || "bundle".equals(dictURL.getProtocol()) || "bundleresource".equals(dictURL.getProtocol())) {
```

danielnaber:
@iuscl-ide Actually we committed this change not long ago, so it's in the source but not yet in the release available on Maven Central. It will be in 3.9, to be released end of September.

iuscl-ide:
OK, I have the last source (from GitHub) linked in Eclipse for debug and this is why I have the “bundle” one, sorry

PR #385, opened 2016-04-07
Enforce DependencyConvergence in parent pom

aldenquimby:
Enforce [dependency convergence](https://maven.apache.org/enforcer/enforcer-rules/dependencyConvergence.html) in parent pom. Runs during `validate` phase, which is the default for the plugin.

Changes needed to get convergence passing:
- downgrade `slf4j.version` from 1.7.9 to 1.7.6 so that it lines up with dependencies
  - only used for slf4j-nop, so I figured version doesn't matter
- exclude `morfologik-polish` from `morfologik-speller`
  - this was a bug in morfologik that will be fixed soon, see https://github.com/morfologik/morfologik-stemming/pull/78
- pull guava, commons-lang, junit up to parent dependencyManagement to resolve most convergence issues
- add special dependencyManagement to `languagetool-dev` because there are so many more conflicts there than any other package

See https://github.com/languagetool-org/languagetool/pull/342#issuecomment-205813554 for additional context

### DIFF
--- a/languagetool-commandline/pom.xml
+++ b/languagetool-commandline/pom.xml
@@ -77,7 +77,7 @@
                  see http://www.slf4j.org/codes.html#StaticLoggerBinder -->
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-nop</artifactId>
-            <version>1.7.9</version>
+            <version>${slf4j.version}</version>
         </dependency>
 
         <dependency>
--- a/languagetool-core/pom.xml
+++ b/languagetool-core/pom.xml
@@ -89,7 +89,6 @@
         <dependency>
             <groupId>commons-lang</groupId>
             <artifactId>commons-lang</artifactId>
-            <version>2.6</version>
         </dependency>
         <dependency>
             <groupId>net.java.dev.jna</groupId>
@@ -110,6 +109,12 @@
             <groupId>org.carrot2</groupId>
             <artifactId>morfologik-speller</artifactId>
             <version>${morfologik.version}</version>
+            <exclusions>
+                <exclusion><!-- see https://github.com/morfologik/morfologik-stemming/pull/78 -->
+                    <groupId>org.carrot2</groupId>
+                    <artifactId>morfologik-polish</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.carrot2</groupId>
--- a/languagetool-dev/pom.xml
+++ b/languagetool-dev/pom.xml
@@ -35,6 +35,83 @@
         </developer>
     </developers>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <!-- resolve convergence issue within org.apache.hadoop:hadoop-common -->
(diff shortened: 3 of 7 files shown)


aldenquimby:
@danielnaber here is a PR based on your comment on #342. The two main ways to fix convergence issues are:
- add exclusions to dependencies
- add dependencyManagement entry

Happy to rework this in either direction if you'd prefer one or the other. Exclusions can get pretty ugly if 3 or more dependencies have conflicting transitive dependencies

danielnaber:
Thanks. All Hadoop-related classes could maybe be moved to their own `languagetool-dev-hadoop` module, but that's an item for the long-term TODO list.

danielnaber:
This seems to have a side-effect that we cannot run tests inside modules anymore, as then the validation will fail (e.g. in `languagetool-dev` and `language-modules/de`). Is this by design?

aldenquimby:
No, that wasn't intentional. When I run `mvn validate` from within `languagetool-language-modules/de`, it works correctly - what error are you seeing?

danielnaber:
I see this result: https://gist.github.com/danielnaber/12c546f821a4c659a01335935b211682, my maven version is 3.2.5

aldenquimby:
Here is my output from running `mvn validate` within `languagetool-language-modules/de`:
```
[INFO] Scanning for projects...
[INFO]                                                                         
[INFO] ------------------------------------------------------------------------
[INFO] Building German module for LanguageTool 3.4-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- maven-enforcer-plugin:1.4.1:enforce (enforce-maven) @ language-de ---
[INFO] 
[INFO] --- maven-enforcer-plugin:1.4.1:enforce (enforce-dependency-convergence) @ language-de ---
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 1.088 s
[INFO] Finished at: 2016-04-08T12:19:46-04:00
[INFO] Final Memory: 12M/309M
[INFO] ------------------------------------------------------------------------
```

I think you might be having an issue specific to snapshots. You might have a cached version of languagetool-core in your remote or local repository, which is missing [this change](https://github.com/languagetool-org/languagetool/blob/71897be83645f6c745d92498787a28029c2ba17b/languagetool-core/pom.xml#L112-L117).

What do you get if you run the following from git root:
```
mvn validate -am -pl languagetool-language-modules/de
```

I think you could also fix this by running `mvn clean install -DskipTests -DskipITs` from the langaugetool-core directory, which would ensure your local repo has the correct snapshot version of core.

That said, it may make sense to either:
- move the enforcer to the `verify` phase, so that it never runs during normal development, and only runs after packaging (I think this is common)
- only turn on the enforcer on certain modules instead of globally for the project

danielnaber:
You're right, `mvn install` fixed the issue. Having validation only during packaging would be better, especially as it takes some time and we already suffer from our slow tests. Having it on for everything except `languagetool-dev` also makes sense. If you could create a pull requests for these changes that would be great.

Issue #200, opened 2014-10-12
don't store binary files in git


danielnaber:
Our git repo is currently 323 MB in size. As we check in binary files, it will keep growing. In the long run, we will probably need to find a way to keep the versioned binary files out of git. Possible solutions (which unfortunately make the workflow more complicated):
- https://github.com/jedbrown/git-fat
- http://git-annex.branchable.com/

danielnaber:
Current repo size: 387MB (`du -h` after a fresh clone)

danielnaber:
Possible solution: Have a separate git module for the resources of each language, e.g. `resources-en` for the English resources. Every module contains a Maven project with the binary files. This project is specified as a dependency by the language, e.g. instead of containing the binary data itself, Maven project `language-en` will specify `resources-en` as a dependency. To make sure that developers always have the latest binary data, they either need to run `mvn install` in all resource projects, or they set up their Maven to get the  snapshot artifacts from Jenkins or Travis, who should build them anyway.

Note that by 'resources' I mean large binary files, not `grammar.xml` or `disambiguaton.xml`.

What's nice about this solution is that it isn't too complicated. Drawbacks:

* Does not really solve the problem forever, as it only moves the growth to the `resources-xy` repos.
* Developer must call `mvn install` in the resources project to test new dictionaries locally.
* http://github.com/languagetool-org/ will list another 30 modules, most of which are not interesting to contributors (as they only work in their own language).

There's another issue that could be solved by this: Every 3 months, we release artifacts to Maven central of which a lot binary hasn't changed (the unmaintained languages). Nobody has complained about it, but it's not really nice. This could be solved by only having a new release of the resources artifacts when it has actually changed (new dictionary etc.)

danielnaber:
Git LFS might help: https://github.com/blog/1986-announcing-git-large-file-storage-lfs

dpelle:
Daniel Naber <notifications@github.com> wrote:

> Git LFS might help:
> https://github.com/blog/1986-announcing-git-large-file-storage-lfs
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/languagetool-org/languagetool/issues/200#issuecomment-91302062>
> .
>

My preference would be to build *all* binary dictionaries
by scripts.  So those big binaries that don't play well
with revision control would not need to be checked-in
anymore.

Regards
Dominique

danielnaber:
Current repo size: 418MB (`du -h` after a fresh clone)

[...]

PR #386, opened 2016-04-12
[ga] add basic support

jimregan:


jimregan:
Some notes:
Individual commits are here: https://github.com/jimregan/languagetool/tree/ga-basic
There's a pretty complete CG for Irish that I'll be converting. I've only added a single disambiguation rule so far, to have everything in place.
This will be maintained (by me) for at least 15 months. I don't expect to have a problem finding a successor.

danielnaber:
Hi Jim, thanks, this looks like a very clean patch. When you say you'll be maintaining this for some time, does it mean you have funding? Can you estimate how much time you'll be able to spend working on it?

About the patch itself, we have problems with large binary files in git (issue #200) - do you think you could move the binary files to their own module (both in git and Maven) that is then fetched as a dependency? Maybe with some Java code to access the files, so that users who need a simple tagger but not LanguageTool would also benefit? If that works, we could use that approach for other languages, too. It is more important the more often the binary files are going to change.

jimregan:
Not as clean as I would have liked, there are a few more changes to be made.

I have funding, though this is not part of my primary work. I'm not sure exactly how long I'll have to work on it, but there'll be a couple of days per month. On the other hand, I expect that some of my colleagues will be interested in adding rules, once the basics are in place, as it would tie in well with some of their projects.

I'll see what I can do with the binaries; in the meantime, I can redo the pull request without those pieces, and leave it open until it's solved.

Thanks for the quick answer!

jimregan:
replacing with #387

PR #387, opened 2016-04-12
[ga] add support (take 2, no binaries)

jimregan:


danielnaber:
Did you make progress with moving the dictionaries to their own project, i.e. does `languagetool-ga-dicts` already exist?

jimregan:
On 26 April 2016 at 12:42, Daniel Naber <notifications@github.com> wrote:

> Did you make progress with moving the dictionaries to their own project,
> i.e. does languagetool-ga-dicts already exist?
>

Yes, it exists: https://github.com/jimregan/languagetool-ga-dicts. It's
just a resource jar, and I've made the ga language module depend on it.
Haven't worked through pushing it to Sonatype or the like yet, because I'm
busy with assignments, but I should be able to pick it back up again after
May 10th.

danielnaber:
I think we should add some light-weight code to access the dictionary from Java, too. Thus the project would need a dependency on morfologik, but wouldn't need a dependency on LT. This way it's useful to more people. It also shouldn't have "languagetool" in its name then.

jimregan:
I guess I can see how that'd be more useful, on paper, but it's not going to be an easy sell for my boss.

danielnaber:
I thought more about this and I think your approach is okay the way it is: having the dictionaries usable without LT dependencies would be nice, but LT adds some words on the fly from a plain text list (for some languages at least). So that list would also need to be part of the new package, too. But then we'd have to re-release every time we change the text-only list which is exactly what I'd like to avoid. So, having a dictionary-only dependency seems okay for now.

danielnaber:
Closing old pull request, feel free to re-open or open a new one if there's progress.

jimregan:
Out of curiosity, what would need to be done for this to be merged?

danielnaber:
I'd basically like to see that a language is continually maintained before merging it. A second contributor would also be a great plus. Also, the recent commits didn't build in Travis and a `<repository>` had been added to the `pom.xml` - we can only accept dependencies at Maven central.

jimregan:
On 2 March 2017 at 08:24, Daniel Naber <notifications@github.com> wrote:

> I'd basically like to see that a language is continually maintained before
> merging it.
>
There's a kind of chicken-and-egg problem here, in that there's little
point in maintaining it if it's never going to be merged.


> A second contributor would also be a great plus.
>
Another chicken-and-egg problem, leading from the first. Without a second
user, there's very little chance of there being a second contributor.


> Also, the recent commits didn't build in Travis and a <repository> had
> been added to the pom.xml - we can only accept dependencies at Maven
> central.
>
AFAICT, the recent commits didn't build *because* of the <repository>, so
the dictionaries-only dependency is *not* ok.

If there's an option other than maintaining my own fork, I'd love to hear
it.

danielnaber:
I can offer to merge it (after the technical problems have been fixed) as "experimental", i.e. if we see it doesn't gain traction after 6 months or so it would be deactivated again. Deactivation means it would still be in git, but not part of a release.

PR #395, opened 2016-04-27
Added new rule for common spelling confusion

sbbic:
អសប្បាយ normally should be អរសប្បាយ


danielnaber:
Thanks for the patch. Could you run `testrules.sh` or `testrules.bat` (on Windows)? I get an error with our automatic example sentence validation.

sbbic:
Sorry about that! Yes, I can. I'll have re-install/setup - it's been years since I've added new rules!

PR #396, opened 2016-04-28
Fixed error of additional Khmer rule

sbbic:
Should be all good now.


danielnaber:
I still get an error (` Did not expect error in:`). The `example`s with and without `correction` seem to be the same? The one without `correction` not supposed to contain an error.

PR #397, opened 2016-04-28
Third try for Khmer addition

sbbic:
I think this is fixed now.


sbbic:
I read the last line of testrules.bat and thought everything passed, but if I scrolled up it was still failing. This should be fixed - you were right, I had incorrect value in the example. Thanks!

danielnaber:
Thanks, I've merged the new rule. It should become active tonight on languagetool.org.

sbbic:
Thanks Daniel! Sorry for so many errors! I'm a bit rusty!

PR #398, opened 2016-04-29
Removed starting slash to resolve Maven error when packaging on Windows.

oriain:
"[ERROR] OS=Windows and the assembly descriptor contains a *nix-specific root-relative-reference (starting with slash) /third-party-licenses"


danielnaber:
Thanks!

PR #424, opened 2016-06-18
Remove rules that are not appropriate for en-GB

cwd24:
These maybe more appropriate for a en-GB en-US bitext 
first floor - floor counting is different
hoover is a brand name, maybe the rule should be reversed.


MikeUnwalla:
@cwd24, thanks. I agree. 

@danielnaber, I am not sure about the correct option to choose after I select 'Merge pull request'. Thus, I did not merge the pull request. Please merge the pull request. Thanks. What is the correct option (create a merge commit or squash and merge)? 

After the merge, I will put a rule for 'hoover' in the default EN rules and a rule for 'ground floor' in the en-US rules.

danielnaber:
Both options are okay, but "Squash and merge" is what we usually use.

MikeUnwalla:
I added a rule to en-US for 'ground floor' (https://github.com/languagetool-org/languagetool/commit/f0f51a148bf4ac739810b5839db8a53c33b10627).
The 'hoover' rule is quite complex, so I want to think about it before I add a rule.

PR #425, opened 2016-06-28
Add folder pt-BR with grammar.xml

MatheusPoletto:
I start the pt-BR rules. Also i add folder with grammar.xml into ` languagetool\languagetool-language-modules\pt  ` please check if place is ok.

http://forum.languagetool.org/users/matheuspoletto

Best Regards, Matheus Otavio Poletto - from Brazil

### DIFF
--- a/languagetool-language-modules/pt/pt-BR/grammar.xml
+++ b/languagetool-language-modules/pt/pt-BR/grammar.xml
@@ -6,6 +6,21 @@ title="Easy editing stylesheet" ?>
 <!--
 Portuguese Brazilian - Grammar and Typo Rules for LanguageTool.
 Copyright (c) 2016 Matheus Otavio Poletto (https://github.com/MatheusPoletto).
+
+This library is free software; you can redistribute it and/or
+modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


danielnaber:
Thanks, some remarks:
* Please run `testrule.sh` to validate your rules. For example, the id `AGENTE` appears more than once, which isn't valid XML.
* Rules with a single word like "memsagem" or usually not useful, as they are detected by the spell checker already.
* For the `id`, please use only standard ASCII characters (e.g. not `Ã` as in `MANUTENSÃO`)

danielnaber:
Thanks. Sorry, one last thing: please add to the comment the LGPL license, like in https://github.com/languagetool-org/languagetool/blob/master/languagetool-core/src/test/resources/org/languagetool/rules/xx/grammar.xml#L9. Also, did you run `testrules.sh`? For me, it complains about `AGENTE_BR` because the error example actually isn't an error.

danielnaber:
Thanks, it has been committed. Unless there's another (unrelated) issue, these rules will become active on languagetool.org later today (around 22:30 CEST) and they will be part of the upcoming daily build at https://languagetool.org/download/snapshots/?C=M;O=D

PR #431, opened 2016-07-01
+15 rules to pt-BR grammar.xml

MatheusPoletto:
Hi, here are more 15 rules i wrote. Thx


danielnaber:
Thanks. Did you run `testrules.sh`? `DAQUI_BR` replaces `daqui` by `daqui a` but that will again match the same rule. I think you need to add a negation so `daqui` gets matched if the next token isn't `a`.

danielnaber:
Thanks, your changes have been merged.

PR #438, opened 2016-07-24
Added more text to test the default phrase in Tamil

muhelen:


danielnaber:
Thanks - any specific reason to add more text and specifically this text?

muhelen:
Yes, this will show more Grammar rules in Tamil on the front page.

danielnaber:
That's not the right place for that - I think it's https://github.com/languagetool-org/languagetool-website/blob/master/include/default_texts.php

PR #439, opened 2016-07-29
[ru] Update grammar.xml

kostyfisik:


danielnaber:
@yakovru Could you check and merge this?

kostyfisik:
I provided an exception to the "kakoe-to" rule as a TODO task and I can construct them quite a bit. However, in the current state the rule is already rather useful - it found many violations in the Wikipedia.

yakovru:
LanguageToool already has rule for words with "-то", "-либо", "-нибудь"

kostyfisik:
I was not able to locate this rule in grammar.xml. Could you, please, point me the line code?

yakovru:
LanguageToool already has Java rule for words "по-над", "по-за".

yakovru:
This rule is not in grammar.xml. This is Java rule. Words for rule are in 
https://github.com/languagetool-org/languagetool/blob/master/languagetool-language-modules/ru/src/main/resources/org/languagetool/resource/ru/compounds.txt

kostyfisik:
"Какая-то" was missed in the file. Rule definition seems to be more maintainable way.

kostyfisik:
"Разработка ПО за деньги" gives a false positive at current version.

kostyfisik:
"Можешь взять пиво какое либо водочки налей." current LT suggest to use "какое-либо", the correct case is to put a comma.  With word list there is no way to put an exception.

kostyfisik:
I can clean up compounds.txt to fit the rule from the last commit. Now it should be possible to provide exceptions.

yakovru:
I think this is bad idea.

kostyfisik:
Why not? Almost all words from the pattern below are in the list. Why some of them are missing is hard to understand. I believe that they have some exceptions, however, it is out of context at the moment. This is a single rule, however, this 60 words can be found here and there in 25к loc file....
For newcomers like me it is hard to understand the reason for this.

<pattern case_sensitive="no">
                 <token regexp="yes">какой|какая|какое|какому|какие|каким|кому|кто|что|такому|где|зачем|как|когда|куда|откуда|чего|отчего|почему|сколько</token>
                  <token spacebefore="yes" regexp="yes">то|либо|нибудь</token>
 </pattern>

yakovru:
All words are in compounds.txt now.   Move all the words in grammar.xml slows down the loading of the program.

Please make new changes in new pull request. I can not split pull request.

yakovru:
I commited your first rule to LT (https://github.com/languagetool-org/languagetool/commit/104b9107248ef2078d95f4d9384c307115dcd9bd)

PR #461, opened 2016-08-07
Should cover all #447

kostyfisik:


yakovru:
Examples of this rules are triggered of existing rules.

kostyfisik:
I have checked is once more. 
<pattern>
                    <token>в</token>
                    <token regexp="yes" case_sensitive="yes" >в[бвгджзклмнпрстфхчшщ].*</token>
                </pattern>
will not trigger on pronouns.

kostyfisik:
Corrected. Added and exception for "СвДП"

PR #519, opened 2016-08-30
Standards of typography (rules of Imprimerie Nationale)

jaihrem:
Suggest "unbreakable space" instead of "thin unbreakable space" before colon (:) according to the "Lexique des règles typographiques en usage à l'Imprimerie nationale" (p.149).

### DIFF
--- a/languagetool-language-modules/fr/src/main/java/org/languagetool/rules/fr/QuestionWhitespaceRule.java
+++ b/languagetool-language-modules/fr/src/main/java/org/languagetool/rules/fr/QuestionWhitespaceRule.java
@@ -86,9 +86,9 @@ public class QuestionWhitespaceRule extends FrenchRule {
             fixLen = 1;
             break;
           case "»":
-            msg = "Le guillemet fermant est précédé d'une espace fine insécable.";
+            msg = "Le guillemet fermant est précédé d'une espace insécable.";
             // non-breaking space
-            suggestionText = " »";
+            suggestionText = " »";
             fixLen = 1;
             break;
           case ";":
@@ -98,14 +98,14 @@ public class QuestionWhitespaceRule extends FrenchRule {
             fixLen = 1;
             break;
           case ":":
-            msg = "Deux-points sont précédé d'une espace fine insécable.";
+            msg = "Deux-points sont précédé d'une espace insécable.";
             // non-breaking space
-            suggestionText = " :";
+            suggestionText = " :";
             fixLen = 1;
             break;
         }
       } else {
-        // Strictly speaking, the character before ?!;: should be an
+        // Strictly speaking, the character before ?!; should be an
         // "espace fine insécable" (U+202f).  In practise, an
         // "espace insécable" (U+00a0) is also often used. Let's accept both.
         if (token.equals("?") && !prevToken.equals("!")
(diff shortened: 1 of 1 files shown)


danielnaber:
Pinging @dpelle

danielnaber:
Pinging @dpelle again, as I don't speak French

danielnaber:
@dpelle, can you check this pull request?

f-knorr:
I think this pull request should be merged. Here is what the French wikipedia says regarding this issue

> En typographie française, le deux-points est généralement séparé par une espace insécable du caractère qui le précède, comme c'est le cas pour tous les signes en deux parties. Pour plus de détails, voir l'article dédié au sujet.

and

> le deux-points — appellation typographique considérée plus puriste que « les deux-points » ( : ), précédé logiquement d’une espace insécable, parfois d’une espace fine insécable ;

dpelle:
> I think this pull request should be merged

I disagree. I don't think that this pull request should be merged.

"séparé par une espace insécable" doesn't necessarily mean the Unicode character U+00A0.  For it to be precise, it would need to say which Unicode character to use.

Here http://www.loriskumo.com/petit-guide-typographique-de-base/  says "Avant chacun de ces caractères (? ! ; : ) on doit placer une espace fine" and it goes on giving the &#8239;  HTML character.

See also Wikipedia: https://fr.wikipedia.org/wiki/Espace_fine_ins%C3%A9cable says "

So I what LanguageTool recommends looks correct.

jaihrem:
@dpelle in the page  http://www.loriskumo.com/petit-guide-typographique-de-base/ you can find a comment about the "espace fine insécable" that refer to http://disq.us/url?url=http%3A%2F%2Fwww.orthotypographie.fr%2Fvolume-I%2Fechelle-extrait.html%3ARDqPUOxSnU6m_mNEkQJyjJRmQ7U&cuid=1522866

But in fact, the rule is different between "France" and "Suisse Romande". That's probably why you find 2 kind of answers.

I let you see which French you really want to check :)

jaihrem:
By the way, Wikipedia's page tells : "Le code typographique français recommande une espace fine insécable devant les signes de ponctuation doubles (point-virgule, point d’interrogation, point d’exclamation, sauf deux-points[1])"
That's precisely the colon exception!

Hope that can help you!

dpelle:
> By the way, Wikipedia's page tells : "Le code typographique français recommande
> une espace fine insécable devant les signes de ponctuation doubles (point-virgule,
> point d’interrogation, point d’exclamation, sauf deux-points[1])"
> That's precisely the colon exception!

Oh I see, you're right about the colon. I checked other sources too
and I now agree it should be a U+00A0 before colon.  So thanks
for the feedback.

But the pull request also changes the space for the French « ».
I still don't really agree with this one. U+202F looks preferable
although both are OK.

See what this commercial French grammar check say:

https://www.druide.com/fr/enquetes/pour-des-espaces-ins%C3%A9cables-impeccables

Which space to use in French typography should ideally be
configurable.  Warning about all rules can be too nitpicky for
many users. It's not easy to have custom configuration forms
in LanguageTool unfortunately.  The Grammalecte French grammar
checker has a nice configuration box.  See:

https://framablog.org/wp-content/uploads/2015/05/ss-formateur.png

jaihrem:
@dpelle according to Wikipedia (https://fr.wikipedia.org/wiki/Espace_ins%C3%A9cable)

"Le deux-points et les guillemets-chevrons constituent une exception en ce que le code typographique français recommande d'insérer devant le deux-points, ou entre les guillemets et le texte qu'ils renferment, non une espace fine insécable mais une espace insécable (espace-mot insécable)[6]. L'usage actuel en PAO française tend toutefois à généraliser l'usage de l'espace fine insécable dans tous les cas, aussi pour le deux-points et les guillemets[9],[10]."

The correct form is U+00A0 but U+202F is tolerable.

dpelle:
> The correct form is U+00A0 but U+202F is tolerable.

That's an interpretation I think, and it says that "L'usage actuel en PAO française tend toutefois à généraliser l'usage de l'espace fine insécable dans tous les cas". 

In any case, LT accepts both U+00A0 or U+202F which is the right thing to do, at least by default if we can't configure preferences, to avoid warning about debatable mistakes.  It suggests U+202F in case space was U+0020.  You would prefer to suggest U+00A0 instead for "deux-points" and "guillemets". I'm not sure if it's best, but I'll accept that.

Better would be to be configurable, so user can ensure consistency according to his preference, as in Grammalecte, but I don't see how to do that in LanguageTool unfortunately.

PR #576, opened 2016-10-16
Fix typo in a Spanish rule description

fitojb:
The short description had “esiste” instead of “existe”, which is an embarrassingly prominent typo. I detected it while using the LibreOffice extension.


danielnaber:
Thanks! More contributions are very welcome, unfortunately not that much is happening for Spanish LT (see https://languagetool.org/languages/ for the level of activity).

PR #582, opened 2016-11-01
Create COPYING.LESSER

Z-Fikar:
I see that LanguageTool is under LGPL 2.1 or later license but there was no license file

Adding a COPYING.LESSER file makes that more accessible
to humans and computers.

### DIFF
--- a/COPYING.LESSER
+++ b/COPYING.LESSER
@@ -1,0 +1,165 @@
+                   GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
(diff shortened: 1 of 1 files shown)


danielnaber:
The license is at https://github.com/languagetool-org/languagetool/blob/master/languagetool-standalone/COPYING.txt, it gets copied to the top-level directory for the ZIP we distribute.

Z-Fikar:
I haven't open that yet. It is good then.

but you know, GitHub will not show it because it can not read wether the repo has license or not. So people have to find it by themselves, like me

Anyway, Thank you for fast response. 👍

PR #589, opened 2016-11-05
languagetool-core: Convert CRLF to LF

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>


stweil:
I suggest using uniform LF line endings (instead of some files using CRLF). If there is consensus, I can send a larger pull request for the remaining java files and also for other files.

danielnaber:
I think all IDEs should handle this transparently, so it shouldn't matter. Also, it makes the git "Annotate" feature more difficult to use, as the latest change in these files will always be the whitespace change.

stweil:
It makes the git "Annotate" feature a little bit more difficult to use: `git annotate -w` ignores any whitespace changes. And it matters because many Linux tools don't expect CRLF line endings. Nor do developers working on Linux. The current mixture of many files with LF and some files with CRLF is especially bad.

danielnaber:
I'm working on Linux and I've never even noticed there are different line endings. I assume everybody contributing to Java code will use an IDE that handles this automatically.

stweil:
Maybe you have non default git settings which enforce LF? Then you won't notice different line endings of course.

Two simple examples what happens with CRLF: vim-tiny shows ^M line endings (that's how I noticed them). grep won't find the usual patter looking for a word at end of line. So `git grep "checker$"` will only find files with LF endings, but not `languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/HunspellRule.java` or other files with CRLF.

PR #594, opened 2016-11-06
Update README.md

yobels:
This link contain page not found https://sourceforge.net/projects/tattoi.u/files/Trigram%20Text/ ,The alternative way from http://www.sttmedia.com/syllablefrequency-filipino that contain trigram data

### DIFF
--- a/languagetool-standalone/README.md
+++ b/languagetool-standalone/README.md
@@ -347,7 +347,7 @@ The words for the Tagger Dictionary were taken from the Philippine Literature
 Domain of Dalos D. Miguel's Comparative Analysis of Tagalog POS Taggers.
 The Tagger Dictionary and the Tagset are made available under LGPL.
 The Trigram Training Data is available at: The Trigram Training Data is available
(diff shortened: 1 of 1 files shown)


jayvdb:
I dont see Tagalog at http://storage.googleapis.com/books/ngrams/books/datasetsv2.html

yobels:
my apologize, Sir, i'm too concern about trigram and forget about the
tagalog itself, so it wasn't correct, am i still have permission to fix up
my pr or i am not passed Mr. John? i am trying to find the correct link
again or find another pr.

danielnaber:
The new URL seems to be 3gram data for the character level. Also, they don't seem to be under a free license. I'm not even sure what the original comment in the README (with the broken link) was referring to.

jayvdb:
It comes from rev6041 , by user [tattoi](https://sourceforge.net/u/tattoi/profile/), which may be @tattoi.

I cant find it online anywhere now.  Perhaps this URL should be removed from the readme?

PR #614, opened 2016-11-23
DNM: Convert DOS EOL to Unix EOL

jayvdb:
Do not merge.
Ran `dos2unix` over the repository.
Tests all pass, but there are a few aspects that need to be discussed first.

---
This change is [Reviewable](https://reviewable.io/reviews/languagetool-org/languagetool/614)


jayvdb:
I have raised https://github.com/languagetool-org/languagetool/issues/615 about the BOMs in some of these files.  `dos2unix` has a `--keep-bom` option, which can be used to not modify these existing BOM, if that would be preferable.

PR #626, opened 2016-12-05
#8 Extend pattern rule test for short messages

agutisan:
- Extend the pattern rule test to make sure that the short message
(<short> in the grammar XML file), if any, is actually shorter than the
other error message (<message>).
- Leave the test initially deactivated (@Ignore).

### DIFF
--- a/languagetool-core/src/test/java/org/languagetool/rules/patterns/PatternRuleTest.java
+++ b/languagetool-core/src/test/java/org/languagetool/rules/patterns/PatternRuleTest.java
@@ -18,20 +18,44 @@
  */
 package org.languagetool.rules.patterns;
 
-import org.junit.Test;
-import org.languagetool.*;
-import org.languagetool.databroker.ResourceDataBroker;
-import org.languagetool.rules.*;
-import org.languagetool.rules.spelling.SpellingCheckRule;
-import org.languagetool.tagging.disambiguation.rules.DisambiguationPatternRule;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


danielnaber:
Thanks. Have you checked what happens when you activate the rule? Maybe we should make it a warning first, so we can activate the rule and fix the places it warns about?

agutisan:
There are many rules which would need to be fixed. I think it's a good idea to make it a warning first.

danielnaber:
Thanks, I've merged it and extended the output a bit, so it's active now.

languagetool-core/src/test/java/org/languagetool/rules/patterns/PatternRuleTest.java
```diff
@@ -90,24 +89,23 @@ public void testSupportsLanguage() {
...
       MultiThreadedJLanguageTool languageTool = new MultiThreadedJLanguageTool(lang);
       for (AbstractPatternRule rule : getAllPatternRules(lang, languageTool)) {
-        failIfShortMessageSmallerThanErrorMessage(rule);
+        warnIfShortMessageSmallerThanErrorMessage(rule);
       }
     }
   }
 
-  private void failIfShortMessageSmallerThanErrorMessage(AbstractPatternRule rule) {
+  private void warnIfShortMessageSmallerThanErrorMessage(AbstractPatternRule rule) {
```
danielnaber: Yes, it has already been fixed in a later commit.

agutisan: Yes, sometimes it's hard to come up with good method names...

On this case the we're using the >= operator:
sizeOfShortMessage >= sizeOfErrorMessage
So if we'd like to be very precise we should probably name it something like
warnIfShortMessageLongerThanOrEqualInSzeToErrorMessage
but that would be a little too verbose ;-)

PR #632, opened 2016-12-12
Fix typo 'statistic suggests' to 'statistics suggest'

krwang:
Update to all instances of `statistic suggests` to read `statistics suggest`. If there is a preference for alternatives such as `a statistic suggests`, I'm open to thoughts. However, I do think `statistics suggest` makes the most sense in the context of the suggestion.


danielnaber:
Thanks, I've manually added this, as we use an external translation tool and don't edit the files directly.

PR #655, opened 2017-01-05
Convert glossaries into terminology checks for translation Quality Assessment

czojo26:
Hi, I try to do the task from the site: http://wiki.languagetool.org/missing-features .
In the main directory I added an example file, so you can test my code.
Can you tell me what should I change with my code so the task can be finished?


languagetool-core/src/main/java/org/languagetool/CsvToXml.java
```diff
@@ -0,0 +1,132 @@
...
+      DOMSource source = new DOMSource(doc);
+      StreamResult result = new StreamResult(new File("file.xml"));
+
+      // Output to console for testing
+//            StreamResult result = new StreamResult(System.out);
+      transformer.transform(source, result);
+
+      System.out.println("File saved!");
+
+    } catch (IOException | ParserConfigurationException ex) {
```
danielnaber: Please don't ignore exceptions

languagetool-core/src/main/java/org/languagetool/CsvToXml.java
```diff
@@ -0,0 +1,132 @@
...
+
+      Element example00 = doc.createElement("example00");
+      rootElement.appendChild(example00);
+
+      Element example01 = doc.createElement("example01");
+      rootElement.appendChild(example01);
+
+      example00.setAttribute("lang", "en");
+
+      example01.setAttribute("lang", "ru");
```
danielnaber: Not sure, what exactly is this code supposed to do, why is `ru` hard coded here? Should this be a test case instead?

languagetool-core/src/main/java/org/languagetool/CsvToXml.java
```diff
@@ -0,0 +1,132 @@
...
+import javax.xml.transform.stream.StreamResult;
+
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+
+public class CsvToXml {
+
+  public static void main(String[] args) {
+    Scanner scanner = new Scanner(System.in);
+    System.out.print("Choose interface (To choose command line type 'cl', to choose UI type 'ui'): ");
```
danielnaber: I think you should use `args` instead of asking the user. Also, a main method should either start a UI or not, I don't think it makes sense for this to be an option.

danielnaber:
I'm closing this, as there was no further feedback. Feel free to reopen if the comments are addressed.

PR #674, opened 2017-02-24
Update segmentation rules for the German language

janissl:
Added 'Tag', 'Jahrestag', 'Spieltag' to the German segmentation rules for keeping together e.g. 20. Tag, 40. Jahrestag, 9. Spieltag.


danielnaber:
Thanks!

PR #678, opened 2017-02-28
Update German segmentation rules

janissl:
Group similar items, fix a typo


danielnaber:
Thanks, I've added those changes manually.

PR #680, opened 2017-03-08
[ga] replacement pull request for Irish

jimregan:
Replacement for #387 
No binaries, no means to add the binaries easily, no will to jump over any more hurdles. If anyone else does, this is at least a cleaner starting point. Feel free to close.


languagetool-language-modules/ga/pom.xml
```diff
@@ -0,0 +1,85 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.languagetool</groupId>
+    <artifactId>languagetool-parent</artifactId>
+    <version>3.6-SNAPSHOT</version>
```
dpelle: Shouldn't that be 3.7-SNAPSHOT?  Since current stable release is 3.6.

languagetool-language-modules/ga/src/main/resources/org/languagetool/resource/ga/disambiguation.xml
```diff
@@ -0,0 +1,379 @@
...
+                <token postag="Verb:.*" postag_regexp="yes">raibh</token>
+            </marker>
+        </pattern>
+        <disambig action="filterall"></disambig>
+    </rule>
+    -->
+    <rule id="NUM_DIG_ORD" name="6ú">
+        <pattern>
+            <marker>
+                <token regexp="yes">([0-9]+)ú</token>
```
dpelle: (nitpick) useless capture with parenthesis. Regexp engines can often be slower when capturing.

languagetool-language-modules/ga/src/main/resources/org/languagetool/resource/ga/irish_tags.txt
```diff
@@ -0,0 +1,3717 @@
...
+Adv:Loc
+Adv:Q
+Adv:Q:Past
+Adv:Temp
+Adv:Temp:Gen:Sg
+Art:Gen:Sg:Def:Fem
+Art:Pl:Def
+Art:Sg:Def
+CC:Cop:Pro:Q
+CC:Noun:Masc:Com:Pl
```
dpelle: Have you considered using single letters like:

N:M   instead of "Noun:Masc"

For the language I maintain, I try to use single letters: it makes writing regexp easier in some cases and faster.
For example, in French, I want a masculine or epicene nouns, can wrote:

N [me]

instead of something like:

Noun:(maculine|epicene)

If I want to match a noun (N) or adjective (J) without caring about gender I can write:

[NJ] . 
(the dot matches any gender m, f or e for masculine, femine or epicene)

languagetool-language-modules/ga/src/main/resources/org/languagetool/rules/ga/grammar.xml
```diff
@@ -0,0 +1,384 @@
...
+            <rule id="FEAR_NOT_FEARR" name="fear not fearr">
+                <pattern>
+                    <token inflected='yes' postag='(?:C[UMC]:)?Verb:.*' postag_regexp="yes"></token>
+                    <token>an</token>
+                    <marker>
+                        <token>fearr</token>
+                    </marker>
+                </pattern>
+                <message>An “fear” a bhí i gceist agat?</message>
+                <example correction='bhí an fear ann'>bhí an <marker>fearr</marker> ann</example>
```
dpelle: The string in correction='...' looks wrong.  I assume that this will not pass the automatic tests.

I think it should be:

<example correction='fear'>bhi and <marker>fearr</marker> ann</example>

And the rule should also have the suggestion tag somewhere like this for example:

<suggestion>fear</suggestion>

languagetool-language-modules/ga/src/main/resources/org/languagetool/resource/ga/irish_tags.txt
jimregan: I have to maintain the upstream analyser this dictionary is derived from, and it really makes no sense for me to go back and forth between two different tagsets. 
Also, single character tagging schemes are appallingly unreadable :)

languagetool-language-modules/ga/src/main/resources/org/languagetool/rules/ga/grammar.xml
jimregan: It worked when I opened the first pull request, but a lot has changed in the year or so since then. I'll take a look.

languagetool-language-modules/ga/pom.xml
jimregan: I'd changed that but not bothered to commit. Until I can get the maven gpg plugin to sign while deploying to sonatype there are no dictionary binaries, so it's not usable.

PR #687, opened 2017-03-25
[WIP] Rule conversion

jimregan:
I've been getting the CG conversion to work again. The first thing is that the GUI depended on an unavailable jar, so I removed it. ISTR that this was removed because it didn't support generating rules that use <regexp> -- is that relevant to converting AtD rules?


jimregan:
Oh, also, there are a few things that need to be fixed in CG conversion to make it usable:
NullPointerException when PREFERRED-TARGETS is used, and it hangs on rules that use NOT.

jimregan:
Mixing surface forms and tags like this:
```
REMOVE (Verb) IF (-1 ("<nach>" Cop));
```
causes java.lang.ArrayIndexOutOfBoundsException

TiagoSantos81:
This project is great. Any chance of making a dump of the Portuguese AtD section conversion when this is done? I am not familiar with AtD but it is very likely that it has some rules that can complement our present set.

jimregan:
If I can find where they are hidden, sure

jimregan:
Ok, it seems to be in [this file](http://www.polishmywriting.com/download/atd_lang20140430.tgz) from [here](https://open.afterthedeadline.com/download/multi-lingual-atd/)

jimregan:
It seems that for Portuguese they only have context-sensitive spell checking, so no rules.

TiagoSantos81:
I know that Portuguese is supported, although I am not sure if that code has been open-sourced.
https://open.afterthedeadline.com/download/multi-lingual-atd/
http://www.polishmywriting.com/download/atd_lang20140430.tgz

TiagoSantos81:
> 
> Ok, it seems to be in this file from here
>  @jimregan
>      Contributor
> jimregan commented 3 hours ago
> It seems that for Portuguese they only have context-sensitive spell checking, so no rules.

Sorry, I was replying in line. You had found that already. So it is only n-gram rules... not useful. I had explored that before but I though the other pattern rules could be encoded in some way.
Anyway, many thanks for looking into it. Now I can "close my AtD file".

TiagoSantos81:
One thing...  what idioms besides English have pattern rules? 
I see model.bin for all alternative languages, but I could never find the source for those binaries.

jimregan:
As far as I can tell, model.bin only contains an ngram model, against which spelling suggestions are ranked. Most languages have only that, and a configuration for running LanguageTool, if available. English seems to be the only language with rules.

TiagoSantos81:
I believe that n-gram models doesn't work that way.
Either way, if it is just an encoded n-gram database for statistical grammar corrections, it would be great to find a way to connect it to LT. This would allow a computationally cheap n-gram equivalence for many other languages. And license wise it is great. 
Unfortunatelly, I am already so filled with plans that I should not even think on the possibility of reverse engineering that.

danielnaber:
Hi Jim, is this in a state that it should be merged?

jimregan:
Not yet. I've not really looked at the AtD parts -- I'm still not sure if it's useful -- and there are a couple of problems that come up with the conversion of the CG that comes with irishfst: for one, the special tags for sentence boundaries, ``<<<`` and ``>>>``, aren't treated as such. I'd like to get it done by September, but if not, I'll have plenty of time to do it then.

danielnaber:
I'm cleaning up old PRs, please re-open if there's new actitivy.

PR #718, opened 2017-06-01
[ru] rule to check verb conjugation

evlEeena:
Hi,
I've noticed that there are not any rules for verb conjugation in Russian. For example, if I write "It's sitting". LT consideres it as right sentence. So, I've tried to create a rule, but I checked my rule only by the test class, that I've also added. Is it possible to check code using application?

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/hunspell/pt_BR.aff
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/hunspell/pt_BR.aff
@@ -1,15 +1,16 @@
 SET ISO8859-1
 TRY áàãâéêíóõôúüçesianrtolcdugmphbyfvkwjqxz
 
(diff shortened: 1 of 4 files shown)


danielnaber:
Hi Elena, thanks for your contribution. As you've already added your rule to `Russian.java`, it should be available in LT in your local build already. Didn't that work for you? Or maybe I misunderstood your question. @yakovru Can you check the rule?

yakovru:
Yes, it works. But still need to test it on a large number of examples. It will take some time.

evlEeena:
@danielnaber,
Oh, sorry, it was my fault, I've had an NPE error in my class and I haven't noticed it before sending the comment. Now it fixed and I'm able to run app after build and check my rules. 
Thanks

yakovru:
Hi Elena, thanks for your contribution.  Can you add header to files RussianVerbConjugationRule.java and RussianVerbConjugationRuleTest.java ?

PR #720, opened 2017-06-04
Add short code to languages page

adamvoss:
I would find having the short code available on the [languages page](https://languagetool.org/languages/) helpful, so I added it.


danielnaber:
I think this way it can confuse people, e.g. the language code used for German is usually not `de`, but one of `de-DE`, `de-AT`, or `de-CH`.

PR #727, opened 2017-06-28
Update SRX rules for the German language

janissl:
Added 'Konferenz' and corresponding compounds.


danielnaber:
Thanks, I've added this (although have removed "Spiel", as "Spielkonferenz" seems to be rare).

janissl:
I agree. It looks that such compound has been only used in some books related with games and usually has no ordinal number before.

PR #743, opened 2017-07-17
add some new rules for recomanded writing to german grammar.xml

FredKruse:
Add some new rules for Duden-recommandations to German grammar.xml
Merge the rules from f-knorr with the rules in the new category (use a little bit more compact formulation for simular rules).

### DIFF
--- a/languagetool-language-modules/de/src/main/resources/org/languagetool/rules/de/grammar.xml
+++ b/languagetool-language-modules/de/src/main/resources/org/languagetool/rules/de/grammar.xml
@@ -9116,60 +9116,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
             <message><suggestion>bekanntermaßen</suggestion> im Sinne von 'bekanntlich' wird zusammengeschrieben.</message>
             <example correction="Bekanntermaßen"><marker>Bekannter Maßen</marker> sind viele einfach depressiv.</example>
         </rule>
-        <rule id="IM_STANDE" name="im Stande (imstande) sein" default="off">
-            <pattern>
-                <token>im</token>
-                <token>Stande</token>
-            </pattern>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/de/src/main/resources/org/languagetool/rules/de/grammar.xml
```diff
@@ -34572,15 +34518,121 @@ Andere Projekte</example>
...
-		<example correction="fantasievolle" type="incorrect">Er lieferte eine <marker>phantasievolle</marker> Beschreibung.</example>
+                <example correction="fantasievolle" type="incorrect">Er lieferte eine <marker>phantasievolle</marker> Beschreibung.</example>
             </rule>
             <rule>
+                <antipattern>
+                    <token regexp="yes">.*pho[nt]ograph.*|graphe(misch|matisch|ologisch).*</token>
+                </antipattern>
                 <pattern>
-                    <token regexp="yes">geograph.*</token>
+                    <token regexp="yes">.*graph.*</token>
```
janschreiber: This is dangerous because it matches 'Photograph' without replacing the first 'Ph'.

FredKruse: The rule includes an antipattern which excludes .*pho[nt]ograph.* and other exceptions.

FredKruse: Hm, the asterisks are missing in the comment

danielnaber: I've commented this rule out for now, it complained about "Graph" which is the recommended spelling according to http://www.duden.de/rechtschreibung/Graph_Darstellung_Mathematik. Could you also check the other new matches at https://languagetool.org/regression-tests/20170717/result_de_20170717.html?

PR #768, opened 2017-08-11
Fix an exception rule for English in the SRX file

janissl:
Added a missing backslash in an afterbreak regex.


danielnaber:
Thanks!

PR #772, opened 2017-08-14
Deletion of many entries from "replace.txt"

Discostu36:
My first pull request,  I hope I did everything correctly.

Many words should be removed from this file because they can be used in British English, only the meaning is different:

"bathroom" does exist in BE but it has a slightly different meaning (may not have a toilet)
https://en.oxforddictionaries.com/definition/bathroom

"braid" does exist in BE when not talking about hair but about silk or cotton
https://en.oxforddictionaries.com/definition/braid

"bureau" does exist in BE when talking about a desk with drawers
https://en.oxforddictionaries.com/definition/bureau

"busy" does exist in BE, the exception is a "busy phone line"
https://en.oxforddictionaries.com/definition/busy

"carnival" does exist in BE, but it desribes a festival instead of a fun fair
https://en.oxforddictionaries.com/definition/carnival

"cookies" do exist in BE when talking about a web browser
https://en.oxforddictionaries.com/definition/cookie

"curb" does exist in BE but it does not describe the edge of the pavement
https://en.oxforddictionaries.com/definition/curb

"elevator" is used in BE in some technical and biological contexts
https://en.oxforddictionaries.com/definition/elevator

"flashlight" is used in BE when talking about signals and lighthouses
https://en.oxforddictionaries.com/definition/flashlight

"stocks" can have various meanings, many of them used in BE
https://en.oxforddictionaries.com/definition/stock

"suspenders" has a different meaning in BE but is used
https://en.oxforddictionaries.com/definition/suspenders

"yard" is used in BE, it just doesn't describe a garden 
https://en.oxforddictionaries.com/definition/yard

There may be even more but I'm not a native speaker so I only suggested clear cases.


MikeUnwalla:
@janschreiber , thanks Jan. I agree with all the changes from @Discostu36.

@danielnaber. Help! I merged the pull request. As you can see, one check failed. As best I can tell, that failure is not related to the pull request. But, what should I do? Revert the merge commit?

janschreiber:
Apparently the tests are failing for the sentence "Where's the bathroom on the Enterprise?" and possibly others. The error message points to [StandaloneDisambiguationRuleTest](https://github.com/languagetool-org/languagetool/blob/master/languagetool-standalone/src/test/java/org/languagetool/tagging/disambiguation/rules/StandaloneDisambiguationRuleTest.java), but this file doesn't contain the actual tests.

danielnaber:
> But, what should I do? Revert the merge commit?

No, I'll fix it.

danielnaber:
Each Java rule has an example sentence, and that one failed now that "bathroom" doesn't trigger a match anymore. I've changed the example sentence, it's now: "We can produce <marker>drapes</marker> of any size or shape from a choice of over 500 different fabrics.", the correction is using "curtains" instead of drapes. Please let me know if that's not an appropriate example.

PR #773, opened 2017-08-14
[eo] Additional rules.

Mbodin:
There was a typo in an example sentence, in the Esperanto file. The sentence itself is marked as wrong, but the typo is not in the marked place, so this is just a typo.


Mbodin:
I also added several rules in-between, including the one I have suggested in http://forum.languagetool.org/t/eo-proposed-rule/2010/

janschreiber:
Thank you for the pull request!
@dpelle Should we pull this?

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
```diff
@@ -1961,59 +1962,91 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+                <token postag='V.* tr.*' postag_regexp='yes'></token>
+                <marker>
+                    <token postag='T.* nak.*' postag_regexp='yes'></token>
+                </marker>
+                <token postag='O.* akz.*' postag_regexp='yes'></token>
+            </pattern>
+            <message>Aspektas ke ‘<match no="2"/>’ rilatas al ‘<match no="3"/>’, sed ‘<match no="2"/>’ ne akuzativas: probable devintus esti ‘<suggestion><match no="2"/>n</suggestion>’.</message>
+            <short>Mankas akuzativo</short>
+            <example correction='tiun'>Mi ne ŝatas <marker>tiu</marker> frazon.</example>
+            <example>Mi ne ŝatas tiun frazon.</example>
```
dpelle: This correct example at line 1992 is not needed, since it matches the correction of the example above it at line 1991.

LT checks both that:
* the incorrect example triggers the error in the rule
* and that the incorrect example after correction no longer triggers the rule.

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
```diff
@@ -1961,59 +1962,91 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+            <example correction='tiajn'>Kiu apogas <marker>tiaj</marker> ideojn?</example>
+            <example>Kiu apogas tiajn ideojn?</example>
```
dpelle: Same remark: the correct example at line 1994 is not needed.

Note there might be other useless correct examples.  This is because LT did not check the corrected sentences in earlier versions.

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
```diff
@@ -4047,6 +4080,16 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
         <example correction="simplan|facilan">Li donis <marker>malkomplikan</marker> respondon.</example>
       </rule>
     </rulegroup>
+    <rulegroup id="HUMIDA" name="humida → malseka">
+      <rule>
+        <pattern>
+          <marker><token inflected="yes" regexp="yes">(ne|mal)?humid((umi|igi|iĝi)|(a|o|eco|ega|eta)j?n?)</token></marker>
+        </pattern>
+        <message>La vorto ‘humida’ estas ne estas oficiala esperanta vorto, kaj la vorto ‘malseka’ normale signifas ekzakte la saman. Ĉu vi intencis ‘<suggestion><match no="1" regexp_match="humid" regexp_replace="malsek"/></suggestion>’?</message>
```
dpelle: General remark: instead of quotes like 'humida', I suggest to use «humida» to be consistent with the other existing rules.

I'd also suggest to write it «\1» since the word is not always humida, it could be humideco for example, so saying humida may be confusing.

But I don't really agree with this rule.  Humida is in PIV and ReVo  (with a slightly different meaning than malseka).  According to ReVo  http://www.reta-vortaro.de/revo/art/humid.html#humid.0a  the word humida is should only avoided outside of a scientific context.  But that advice is quite fuzzy.

I even find «humida» in the dictionary of the Acdemy of Esperanto: http://www.akademio-de-esperanto.org/akademia_vortaro/index.html
So I don't think we can say that it's not an official Esperanto word.  I'd suggest to either:
* remove this rule
* or add context to indicate it as an error if we can find pattern where it's clearly a mistake (with a link to ReVo or perhaps other source of information about it).

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
```diff
@@ -1961,59 +1962,91 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+    <rulegroup id="AKUZATIVO" name="ĝenerala problemo pri akuzativo">
+        <rule id="AKUZATIVA_TIKORELATIVO_RILATA_AL_NEAKUZATIVA_SUBSTANTIVO" name="Akuzativa “Ti”-korelativo rilata al neakuzativa substantivo.">
+            <pattern>
+                <token postag='V.* tr.*' postag_regexp='yes'></token>
+                <token postag='T.* akz.*' postag_regexp='yes'></token>
+                <marker>
+                    <token postag='O.* nak.*' postag_regexp='yes'></token>
+                </marker>
+            </pattern>
+            <message>Aspektas ke ‘<match no="2"/>’ rilatas al ‘<match no="3"/>’, sed ‘<match no="3"/>’ ne akuzativas: probable devintus esti ‘<suggestion><match no="3"/>n</suggestion>’.</message>
```
dpelle: Not sure if there is any official rule about this, but «aspektas ke» is very rare somehow, whereas «Ŝajnas, ke» is the common idiom.  Looking in tekstaro.com (without case sensitive search), I find:
* only one match for «aspektas, ke» but it's not in the same sense as in above sentence
* and I find 218 matches for «ŝajnas, ke» 
So it's wiser to say «Ŝajnas, ke».
If there is a link which justifies that «Aspektas¸ ke» is incorrect, we could also add a rule to detect it.

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
```diff
@@ -2066,7 +2099,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
       <example correction="">Mi <marker>vi</marker> vidas kaj vin aŭdas.</example>
       <example correction="">Mi <marker>vi</marker> ankoraŭ ne tre bone vidas.</example>
       <example>Mi <marker>vin</marker> ankoraŭ ne tre bone vidas.</example>
-      <example correction="">Mi <marker>vi</marker> ne bene vidas.</example>
+      <example correction="">Mi <marker>vi</marker> ne bone vidas.</example>
```
dpelle: Thanks, that mistake must have happened when I was learning Italian!

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
```diff
@@ -1961,59 +1962,91 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+                </marker>
+            </pattern>
+            <message>Aspektas ke ‘<match no="2"/>’ rilatas al ‘<match no="3"/>’, sed ‘<match no="3"/>’ ne akuzativas: probable devintus esti ‘<suggestion><match no="3"/>n</suggestion>’.</message>
+            <short>Mankas akuzativo</short>
+            <example correction='frazon'>Mi ne ŝatas tiun <marker>frazo</marker>.</example>
+            <example>Mi ne ŝatas tiun frazon.</example>
+            <example correction='ideojn'>Kiu apogas tiajn <marker>ideoj</marker>?</example>
+            <example>Kiu apogas tiajn ideojn?</example>
+        </rule>
+        <rule id="NEAKUZATIVA_TIKORELATIVO_RILATA_AL_AKUZATIVA_SUBSTANTIVO" name="Neakuzativa “Ti”-korelativo rilata al akuzativa substantivo.">
```
dpelle: I don't think we should be so verbose in rule IDs. It's just an ID without meaning (never shown to users, except in command line for example where it can be used to disable a rule).  The meaning is in the name="…".

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
```diff
@@ -4047,6 +4080,16 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
         <example correction="simplan|facilan">Li donis <marker>malkomplikan</marker> respondon.</example>
       </rule>
     </rulegroup>
+    <rulegroup id="HUMIDA" name="humida → malseka">
+      <rule>
+        <pattern>
+          <marker><token inflected="yes" regexp="yes">(ne|mal)?humid((umi|igi|iĝi)|(a|o|eco|ega|eta)j?n?)</token></marker>
```
dpelle: Since you wrote inflected="yes", the rule should work without having to put j?n?  at the end of the regexp.

The regexp does not look right.  Better would be:

<token …>(ne|mal)?humid(um|ig|iĝ|ec|et|eg)[ao]</token>

dpelle:
One more remark: it's best to put unrelated changes in different pull requests (not for this PR, but for future PR). I could have immediately merged trivial changes like bene → bone it was in a separate PR.

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
Mbodin: I have to admit that I did not really have put large care in the error message itself. So sure, « Ŝajnas, ke » perfektly suits me.
I mainly speak Esperanto at home with my wife, so my Esperanto tends to drift from the official one ☺
I will make a patch for this.

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
Mbodin: Sure. I used the automatic generator tool to build some parts of this rule… so the ID was generated automatically. I am fine with any ID: the important part is the rule ☺

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
Mbodin: I wasn’t aware of this. Thanks! I will change this.

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
Mbodin: ☺
I saw it by chance… so I repported it ☺

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
Mbodin: I really wasn’t sure how to do this. Thanks. Was does “inflected” exactly matches?

languagetool-language-modules/eo/src/main/resources/org/languagetool/rules/eo/grammar.xml
Mbodin: Good point for the quotation mark. I will change this.

We can remove this rule. It was just a general thought that I had that I was starting to hear “humida” at places were “malseka” should generally be used. It is probably out of the scope of the tool to know whether we are in a scientific context or not, so removing this rule seems acceptable.

Mbodin:
Oh, sorry about this. I am not really used to the pull request tool of Github. I guess that I should have created a separate branch or something like this: when pushing to the main branch, the pull request was automatically updated. I will probably perform the requested changes today.
Ĝis baldaŭ!

dpelle:
I have not pull the pull request as-is because I did not agree with all changes, and they were too many changes at once. The pull request also included unrelated merged (instead of rebasing).

But I have:
* checked-in the fix for the typo  (bene → bone), see 9fae36b73b479ed275e7cb6760475c3a040cd1f2
* added a rule HUMIDA.  I did not use the rule in this pull request because it would give false positive without checking the context.  "Humida vento" should be correct for example. Rule HUMIDA was added in b83e59765c93e38bf6cf43ef57a3696491b58910
* I have not looked yet at the remaining change for the rule AKUZATIVO in this pull request. I'll have a look later.

Mbodin:
Hi,
I am not familiar with all the mecanics behind pull request. So yes, I did things unoptimally. However, all the commits before the merges are just copies: I think that you can ignore every commit after https://github.com/languagetool-org/languagetool/pull/773/commits/16cb4291e08890ebffc10245f7aa976ca6fdf998.

I have agreed to remove my rule about “humida”, as I feel that I lack the expertise about this particular word: I didn’t know that it was official.

About the accusative rule, I think that it can be made better. Initially, it only matched patterns such as “tiu* vorto*” where one star was an accusative mark and the other not. As-is, I think that it generates false-alarms. So I specialised it to ensure that there is a transitive verb before: accusative is probable in such a position. This rule works, I think, as it conflicted with a non-correct sentence of LT’s database (the corresponding sentence has been corrected in between in Tatoeba). I think that we can make this rule more generic, for instance also checking for patterns like “ĉiuj* vortoj*” or “sia[j]* vorto[j]*”, as well as looking for other situations in which accusative is expected or not expected.

Hoping this helped,
Martin.

dpelle:
>I have agreed to remove my rule about “humida”, as I feel that I lack
> the expertise about this particular word: I didn’t know that it was official.

Note that I did not remove it completely.
I instead made a rule HUMIDA that signal an error depending on the context, thanks to your idea.

So:
* "humida vango" or "humidaj okuloj" (etc.) will trigger an error (because what is meant here is wet i.e. malseka)
* "humida vento", "humda aero" (etc.) will not trigger an error because it's about vapor in the air, and humida is then a correct technical word.

Regarding the rules for accusative, they have become complex. They need to take into account plenty of cases which are not so trivial.  I'll see what can be done there when I find time.

danielnaber:
I think we can close this PR? If there are issues left, it's probably best to re-open a new one.

PR #787, opened 2017-09-15
Improve GermanTagger.java

nopepper:
Make GermanTagger more accurate and add support for recognizing dash-separated words.


languagetool-language-modules/de/src/main/java/org/languagetool/tagging/de/GermanTagger.java
```diff
@@ -1,207 +1,244 @@
...
+    }
+  }
+
+  //Removes the first part of dash-linked words (SSL-Zertifikat -> Zertifikat)
+  private static String sanitizeWord(String word) {
+    String result = word;
+    String[] splitWord = word.split("-");
+    if (splitWord.length > 1) {
+      for (int i = splitWord.length - 1; i >= 0; i--) {
+        if (!splitWord[i].trim().equals("") && !splitWord[i].trim().equals(null)) {
```
f-knorr: I think the 2nd condition is not necessary because `trim` cannot return a null-value. (If it could, the 1st condition would probably throw a NullPointerException. Therefore, I would recommend writing `!"".equals(splitWord[i].trim())`).

f-knorr: If splitWord[i] can take null, add `splitWord[i] != null`

languagetool-language-modules/de/src/main/java/org/languagetool/tagging/de/GermanTagger.java
```diff
@@ -1,207 +1,244 @@
...
+    for (TaggedWord taggedWord : taggedWords) {
+      List<String> allButLastPart = compoundParts.subList(0, compoundParts.size() - 1);
+      String lemma = String.join("", allButLastPart)
+                   + StringTools.lowercaseFirstChar(taggedWord.getLemma());
+      result.add(new AnalyzedToken(word, taggedWord.getPosTag(), lemma));
+    }
+    return result;
+  }
+
+}
```
f-knorr: Can you also add some unit-test to GermanTaggerTest.java?

PR #789, opened 2017-09-18
German tagger improvement for dash-linked words

nopepper:
Sorry about the new pull request. I seem to have messed up the old one.
Here is the modified class with the requested changes and unit test.

### DIFF
--- a/languagetool-language-modules/de/src/main/java/org/languagetool/tagging/de/GermanTagger.java
+++ b/languagetool-language-modules/de/src/main/java/org/languagetool/tagging/de/GermanTagger.java
@@ -136,8 +136,14 @@ public class GermanTagger extends BaseTagger {
               if (word.split(" ").length == 1) {
                 String wordOrig = word;
                 word = sanitizeWord(word);
+
+                //Tokenize, start word uppercase if it's a result of splitting
                 List<String> compoundedWord = compoundTokenizer.tokenize(word);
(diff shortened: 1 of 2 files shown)


f-knorr:
Hi @nopepper ! Thank you! I think that's a great contribution to LT.

f-knorr:
Hi @nopepper ! I have reverted the pull request due to a lot of errors in the regression test (see https://languagetool.org/regression-tests/20170918/result_de_20170918.html) I suggest we wait until after the release of the new LT-version (end of september) and then we see what has to be done to turn of these false positives (maybe it's enough to add an exception to DE_CASE rule for adjektives containing '-')

f-knorr:
I think it is pretty easy to define antipatterns so that DE_CASE-rule does not trigger for the tagged compounds. The tagging needs some refinement, too: I think compounds with a leading number (1930-er or 5-fache) should be ignored.

nopepper:
Thanks, noted.

f-knorr:
I have added some test for DE_CASE-rule (eedc0e6) and an exception which would turn off errors for upper-cased compounds (6982a5e) such as UV-bestrahlt etc.

f-knorr:
When it comes to some nouns, your approach will create false tagging:

- Das Entweder-oder, das Teils-teils, das Sowohl-als-auch, das Als-ob
- Ich habe Du-weißt-schon-wen getroffen

PR #800, opened 2017-09-25
[de] Make GermanTagger recognize and tag dash-linked words (fixed)

nopepper:
The bugs mentioned in the last pull request should be fixed now.
Note: By default the program will mark uppercase linked words that it hasn't recognized as Nouns (SUB). This may need to be removed later if it causes problems.

Examples:
IT-Unternehmen: POS same as 'Unternehmen'
Entweder-oder: [-]
1930-er: [-]
6534-289332... : [-]
Medizinisch-Korrekt: POS same as 'korrekt'


nopepper:
^ The failure seems to come from a faulty rule in grammar.xml

f-knorr:
I don't think the rule is faulty: Before your pull request the word "HTML-Backend" was tagged "UNKNOWN". Now it seems to get tagged by "SUB" which is not a typical tagging. It should by something like "SUB:NOM:SIN:NEU"

f-knorr:
I have just seen in your comment that you describe this behavior and that it was intended. I think if we cannot provide a valid "SUB"-tagging, no tagging should be provided at all.

nopepper:
I understand. However, "Backend" gets an UNKNOWN tag when analysed
regularly. Linked words that start with uppercase are typically nouns, so
they're marked as SUB when nothing else can be said about them. This
feature wouldn't be hard to disable though.

nopepper:
Hello @f-knorr, I have removed the faulty tagging. Everything should work properly now.

f-knorr:
Looks good to me. Great job! Let's see what the nightly diff says...

f-knorr:
Hi @nopepper ! 
I have just found another problem that is probably related to your pull request. The problem is that the position of errors are shifted due to the tagging of hyphenated compounds as non-hyphenated words.
Take, for example:
_Beim Wiederaufbau wurde das original-gotische Achsfenster mit reichem Maßwerk ausgebaut und an die südliche Polygonseite gesetzt._
LT does not know the word _Polygonseite_ which should be marked as an error. However, marked is "südliche Pol". The reason probably is that _original-gotische_ is tagged as a form of "gotisch". When I remove the "-" in "original-gotisch", everything is fine. Can you please create another pull request that either fixes the problem or reverts the previous changes?

languagetool-language-modules/de/src/main/java/org/languagetool/tagging/de/GermanTagger.java
```diff
@@ -138,17 +137,21 @@ public AnalyzedTokenReadings lookup(String word) throws IOException {
...
                 word = wordOrig;
+
                 if (linkedTaggerTokens.size() > 0) {
-                  readings.addAll(getAnalyzedTokens(linkedTaggerTokens, wordOrig, compoundedWord));
+                  if(wordStartsUppercase){ //Choose between uppercase/lowercase Lemma
+                    readings.addAll(getAnalyzedTokens(linkedTaggerTokens, word));
+                  }
+                  else{
+                    readings.addAll(getAnalyzedTokens(linkedTaggerTokens, word, compoundedWord));
+                  }
```
f-knorr: sollte hier word durch wordOrig ersetzt werden?

nopepper:
Hello @f-knorr ,
I unfortunately couldn't recreate the bug on my own, but I've addressed the lemma issue and now the words should be properly tagged. Please let me know if this fixes the problem on your side.

Pull request:
https://github.com/languagetool-org/languagetool/pull/810

f-knorr:
Hi @nopepper ,

thanks for you pull request. My example is no longer reproducible as I have added the formerly unknown 'Polygonseite' to the dictionary.

PR #808, opened 2017-10-04
Some spaces should be non-breakable.

Mbodin:
In the previous state of the program, it correctly detected an issue in “test : test”, but suggested to replace to “test : test” itself instead of replacing the space by a non-breakable space.

I think that we should add some tests for this case.

I have no idea what Git did to make this branch so complicated (I just did “git pull”, “git checkout -b whitespace-fr”, did my change, then “git commit -a”, “git pull --rebase” and “git push”). Please only consider my only commit in this branch.


danielnaber:
Sorry, it seems this PR somehow got lost. If you create a new one that contains only relevant changes, we'll have a look. This one seems to contain a lot of unrelated changes in its diff.

PR #821, opened 2017-10-26
Add vocab from corpus of technical terms

GiPfi:


danielnaber:
Thanks. A lot of these terms seems to be recognized already? Could you check using https://community.languagetool.org/analysis/index?lang=de which terms need to be added?

GiPfi:
Sure! So whichever word already has a complete analysis of token, lemma and PoS tag doesn't need to be added?

danielnaber:
I think so, or can you think of a reason why words with a complete and correct analysis still need to be added? Many of the compounds are probably only detected because they can be split and their last part is known, but that seems okay.

GiPfi:
Well it may depend on the purpose. Looking at it from the perspective of a jwordsplitter user, it may be useful to have even words in the list the parts of which are already known, e.g. when you have a compound of 3 words and 2 of these combine into a compound of its own right like "Metallschlagbohrer" -> "Schlagbohrer". Or would those cases be covered by other mechanisms?

danielnaber:
From a grammar perspective, it doesn't matter wheter it's `Metall+schlag+bohrer` or `Metall+schlagbohrer`, as both `schlagbohrer` and `bohrer` have the same gender etc. (by definition, as `schlagbohrer` also ends in `bohrer`). So I suggest you use https://community.languagetool.org/analysis/index?lang=de to see if a reading is missing, and we can then analyze why.

GiPfi:
Okay. Could you explain briefly what kind of words you include in `jwordsplitter`'s `additions.txt`. We were wondering if that would be a more appropriate place to keep these compound words 🤔

danielnaber:
`additions.txt` in jwordsplitter doesn't contain compounds, but potential compound parts. E.g. `uniform` is not `Uni`+ `Form`, but it can be a part of a compound like `Arbeitsuniform`. It's just an extension of the main dictionary in binary format.

GiPfi:
hmm so what would be the difference between `Uni`+`Form` in `Arbeitsuniform` and `Schlag`+`Bohrer`as part of `Hammerschlagbohrer`?

danielnaber:
`Uniform` is not a compound, it doesn't make sense to split it. For Schlagbohrer it makes more sense, as a Schlagbohrer is a kind of Bohrer.

GiPfi:
I see 😄

GiPfi:
Would you mind us adding a `jwordsplitter` constructor which loads additional words from a file? Like this it would be possible to enhance the dictionary without modifying the it.

danielnaber:
There's already a very similar constructor, but it doesn't load additional words but a new list of words. Maybe it's more robust and transparent to use that?

GiPfi:
hm generally that's fine. Is there an option to use both the embedded dictionary AND an additional list of words. If I understand it correctly, only one or the other is possible atm, is it?

danielnaber:
I'm cleaning up old PRs. I understand that this PR isn't needed anymore, please re-open if that's not correct.

PR #833, opened 2017-11-18
String rework

comradekingu:
I don't know about "whitespace" vs "space", nor what the policy on full stops is.

### DIFF
--- a/languagetool-core/src/main/resources/org/languagetool/MessagesBundle.properties
+++ b/languagetool-core/src/main/resources/org/languagetool/MessagesBundle.properties
@@ -1,5 +1,7 @@
 # English translation of LanguageTool
-# Copyright (C) 2006 Daniel Naber (http://www.danielnaber.de), Marcin Milkowski
+# Copyright © 2006 Daniel Naber (http://www.danielnaber.de), Marcin Milkowski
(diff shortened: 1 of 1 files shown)


danielnaber:
Thanks, but I hesitate to make these small changes to the English text, as it triggers a re-wording of all translations. Could you maybe provide a pull requests that only fixes real errors and skips the small stuff (like uppercase and typography)?

comradekingu:
@danielnaber No translator has ever complained about that. And there is a good chance translations will improve as a result. Weblate fortunately handles "fuzzy" strings correctly, so translations are just marked fuzzy for review, nothing else happens.

danielnaber:
We're not using Weblate but transifex. It might *mark* them as fuzzy, but via the API we'll get the English string again until the translation has been updated.

comradekingu:
@danielnaber https://hosted.weblate.org/projects/languagetool am i missing something?
Yes, Transifex handles fuzzy strings so badly it takes time to get into.

danielnaber:
https://hosted.weblate.org/projects/languagetool is not in use

comradekingu:
@danielnaber Was there an issue with this?

PR #838, opened 2017-11-19
Install script

InnovativeInventor:
I added an install script to help everyone in building and installing languagetool. It also downloads java and maven automatically if they are not detected.

Here are is the usage options:
```
Usage: install.sh <option> <package>
Options:
   -h --help                   Show help
   -o --override <OS>          Override automatic OS detection with <OS>
   -b --build                  Builds packages from the bleeding edge development copy of LanguageTool
   -p --package <package>      Specifies package to install when building (default all)
   -c --command <command>      Specifies post-installation command to run (default gui)

Commands:
   GUI                         Runs GUI version of LanguageTool
   commandline                 Runs command line version of LanguageTool
   server                      Runs server version of LanguageTool

Packages (only if -b is specified):
   standalone                  Installs standalone package
   wikipedia                   Installs Wikipedia package
   office-extension            Installs the LibreOffice/OpenOffice extension package
```


install.sh
```diff
@@ -0,0 +1,212 @@
...
+
+detect_maven() {
+    if ! [ -x "$(mvn -v)" ]; then
+        install_maven
+    fi
+}
+
+detect_java() {
+    if ! [ -x "$(java -version)" ]; then
+        install_maven
```
TiagoSantos81: Possible typo here. I believe you meant `install_java`.

install.sh
```diff
@@ -0,0 +1,212 @@
...
+    fi
+}
+
+postinstall_command () {
+    if [ "$command" = GUI ] || [ "$command" = GUI ]; then
```
TiagoSantos81: The || has the same argument. Maybe `standalone`?
I haven't tested it, but seems to be another great tool. Many thanks.

danielnaber:
Thanks, but I'm not sure yet. First, it still contains `LanguageTool-3.9` so it would fail with the next version. The other question is, what is the use case of this? Java developers will have Java and Maven installed anyway. And for users it seems limited by the fact that it won't work on Windows and only on Debian-based Linux, is that correct?

install.sh
InnovativeInventor: Thanks. I just changed it to `install_java`.

InnovativeInventor:
I meant to change `LangaugeTool-3.9` to `$RELEASE`, but forgot to. Thanks for telling me.

The script doesn't just install Java and Maven, but it also builds it. I've also tested it on MacOS. In the future, Fedora and other operating systems can be supported (its easy to add support for them). This script makes it easy for people who see the LanguageTool on GitHub to build it or install it with only two lines.

InnovativeInventor:
In addition, if you want to install LanguageTool from the command line, it is extremely easy to do with an install script. This is important for people who are running headless servers with LanguageTool. I think it would be a nice resource to have.

install.sh
```diff
@@ -0,0 +1,212 @@
...
+    echo "Java is not installed"
+    echo "Installing java . . ."
+
+    if [[ "$OSTYPE" == "linux-gnu" ]]; then
+        sudo apt-get update
+        sudo apt-get install java -y
+
+    elif [[ "$OSTYPE" == "darwin"* ]]; then
+        if ! [ -x "$(brew -v)" ]; then
+               /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
```
TiagoSantos81: Tried this script moments ago on Mint 18.3.
It ends up not detecting JavaJDK when -d is used.
Needs: `sudo apt-get install openjdk-8-jdk`.
Also #JAVA_HOME was not defined, but I am not sure if this is solved after openjdk-8-jdk installation.

Still not thoroughly tested but, when ready, it will be a time-saver.

install.sh
```diff
@@ -0,0 +1,212 @@
...
+    echo 'An tool for installing or building for LanguageTool.'
+    echo 'Usage: install.sh <option> <package>'
+    echo 'Options:'
+    echo '   -h --help                   Show help'
+    echo '   -o --override <OS>          Override automatic OS detection with <OS>'
+    echo '   -b --build                  Builds packages from the bleeding edge development copy of LanguageTool'
+    echo '   -p --package <package>      Specifies package to install when building (default all)'
+    echo '   -c --command <command>      Specifies post-installation command to run (default gui)'
+    echo
+    echo 'Commands:'
```
TiagoSantos81: It would be easier to follow if "Commands" instructions are shown after "Package" instructions, like in the former list.

install.sh
```diff
@@ -0,0 +1,212 @@
...
+    -h|--help)
+    help=YES
+    shift # past argument
+    ;;
+esac
+done
+
+display_help() {
+    echo
+    echo 'An tool for installing or building for LanguageTool.'
```
TiagoSantos81: `An tool for installing or building for LanguageTool.`
Shoud read:
`A tool for installing or building LanguageTool.`

README.md
```diff
@@ -21,7 +21,7 @@ For more technical details, see [our wiki](http://wiki.languagetool.org).
 #### Scripted installation and building
 To install or build using a script, simply type:
 ```
-curl -L https://git.io/vFHpu | sudo bash <options>
+curl -L https://bit.ly/2kfvAha | sudo bash <options>
```
TiagoSantos81: The static link that existed before was an old version, but it is a potential security problem to have a script that requires with administrative permissions to retrieve code dynamically. That way it is impossible to review this code.

If the script is included, why should a web version be retrieved?

README.md
```diff
@@ -18,6 +18,33 @@ for issues to get started.
 
 For more technical details, see [our wiki](http://wiki.languagetool.org).
 
+#### Scripted installation and building
+To install or build using a script, simply type:
+```
+curl -L https://bit.ly/2kfvAha | sudo bash <options>
```
TiagoSantos81: If the script is included, there is no reason to retrieve the on-line version. Besides, it would have to be a static version to be possible to be reviewed.

install.sh
```diff
@@ -0,0 +1,364 @@
...
+            echo "   Please install maven yourself or override automatic OS detection with -o <OS> See help for more details."
+    fi
+}
+
+install_java() {
+    echo "Java is not installed"
+    echo "Installing java . . ."
+
+    if [[ "$OSTYPE" == "linux-gnu" ]]; then
+        add-apt-repository ppa:webupd8team/java -y
```
TiagoSantos81: Adding PPA is a major decision on behalf of the user. Webupd8 team has a good reputation, but all packages should come from the distribution repository to avoid breakages or security issues.

install.sh
```diff
@@ -0,0 +1,364 @@
...
+        apt update
+        apt install oracle-java8-installer -y
+
+    elif [[ "$OSTYPE" == "darwin"* ]]; then
+        if ! [ -x "$(brew -v)" ]; then
+               /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
```
TiagoSantos81: Retrieving a script from the Internet, may be contentious.Why not upload this script as well? 
Notice that homebrews with administrative permissions may cause the same concerns as PPAs.

I don't own a macOS, so macOS section needs to be reviewed by more people and can take some time.
Probably the best idea is to upload a base version concerning Linux (commenting out this else if) and after that, make another pull request with the macOS/darwin changes.

install.sh
```diff
@@ -0,0 +1,364 @@
...
+    quiet=YES
+    shift # past argument
+    ;;
+esac
+done
+
+display_help() {
+    echo
+    echo 'Script version 2.0.'
+    echo 'An tool for installing or building LanguageTool.'
```
TiagoSantos81: An -> A

install.sh
```diff
@@ -0,0 +1,364 @@
...
+        exit 1
+    fi
+}
+
+build_or_install_loud () {
+    # Build or install loudly
+    if [ "$build" == YES ]; then
+        build
+        echo "Your build is done."
+        echo "Post-installation commands are not availble for the build option. Contributions are welcome."
```
TiagoSantos81: availble -> available

install.sh
```diff
@@ -0,0 +1,364 @@
...
+        java -jar "$RELEASE-$version"/$cmd.jar $file
+    fi
+}
+
+build_or_install_quiet () {
+    # Build or install loudly
+    if [ "$build" == YES ]; then
+        build_quiet
+        echo "Your build is done."
+        echo "Post-installation commands are not availble for the build option. Contributions are welcome."
```
TiagoSantos81: available -> availble

README.md
InnovativeInventor: So do you want me to revert back to git.io to create static links? We could also just place it on the langaugetool website and have this script fetch it from there.

install.sh
InnovativeInventor: I added a dialog to ask the user if they want to add webupd8

README.md
InnovativeInventor: I do see how using `curl | sudo bash` can be worrying for some paranoid folks, so maybe we could sign it with our pgp keys and/or put the checksum up on github so those people can check the script themselves. Something like pipethis: https://github.com/ellotheth/pipethis
could also be used.

README.md
InnovativeInventor: So they should first do git clone, then execute the install script? If you are doing git over https, I'm not sure how much of a difference security-wise that is.

InnovativeInventor:
@TiagoSantos81 I've added a checksum to mitigate risks. We could use PGP or something if you want. At the end, you have to get Languagetool from the internet, so I'm not really seeing the security issues. I've also added dialogs to ensure that PPAs and homebrew are only added if the user agrees or runs the script with -a.

TiagoSantos81:
@InnovativeInventor Many thanks for addressing the concerns. In my opinion, I think it is enough to hard link the curl to the Languagtool repository script, once it is integrated. Is is easier and safe once, reviewed.
That can be done after merging the script.
I would prefer to see several iterations of `sudo apt-get install openjdk-8-jdk`, then the PPA, but I won't nag you more with this. I haven't tested it again, though.

@danielnaber Do you agree with this script inclusion at this stage? Does it need more changes?

danielnaber:
@TiagoSantos81 Feel free to commit if you think you can audit this script and can maintain it. I can neither test nor maintain it.

TiagoSantos81:
` Feel free to commit if you think you can audit this script and can maintain it. I can neither test nor maintain it.`
Thank you, @danielnaber. I was hoping that a more experienced person would review this. This way I have to be more picky.

@InnovativeInventor Considering that there won't be second opinions, I have to insist on the changes I've requested. 
- No need for PGP signing, but the bash script will be changed to the version in LanguageTool repository.
- Remove the darwin part, since I can't review it, and add it later as a separate pull request.
- Use non-opinative user messages (line 27). 

   The matter with PPAs and scripts by curl has to do with the review process and human error. Something that happens to everybody, and that has happened before on this script:
https://github.com/languagetool-org/languagetool/pull/838/commits/ee29a530d2efea2d15c07e0c96c530bb188a56a7 , https://github.com/languagetool-org/languagetool/pull/838/commits/ac5d668a1bb3a951fc8a2d49b6288a1344563d7a) and th_quietat continues to happen (leftovers on `.gitignore`, line 125, the removal should be verbose except on install input, build_quiet() == build(), typos: commnd->command,  no text is specified to check -> no text was specified to be checked).

- For java JRE and JDK `sudo apt-get install default-jre default-jdk` should be sufficient for any ubuntu based distro. Those are the one's I can test.

I still haven't tested this last improved version, but looks good apart from the points I mentioned.

PR #855, opened 2017-12-07
Neuralnetworkrule cleaned

gulp21:
Add neural network based rules, as discussed in https://forum.languagetool.org/t/neural-network-rules/2225

* Add support for single and two layer networks for confusion pair disambiguation
* Add parameters and configuration options for choosing the path to a word2vec model and neural network configurations


languagetool-core/src/main/java/org/languagetool/rules/neuralnetwork/NeuralNetworkRule.java
```diff
@@ -0,0 +1,182 @@
...
+      String token = tokens[i].getToken();
+      if(getSubjects().contains(token)) {
+        final String[] context = getContext(tokens, i);
+        final float[] y = classifier.getScores(context);
+        final Suggestion suggestion = getSuggestion(y);
+        if(!suggestion.matches(token)) {
+          if (!suggestion.isUnsure()) {
+            ruleMatches.add(createRuleMatch(tokens[i], suggestion, y));
+          } else {
+            System.out.println("unsure: " + getMessage(suggestion, y) + Arrays.toString(context));
```
danielnaber: We probably shouldn't print to STDOUT...

languagetool-core/src/main/java/org/languagetool/rules/neuralnetwork/NeuralNetworkRuleCreator.java
```diff
@@ -0,0 +1,38 @@
...
+  private NeuralNetworkRuleCreator() {}
+
+  private static final String CONFUSION_SET_FILENAME = "neuralnetwork/confusion_sets.txt";
+
+  public static List<Rule> createRules(ResourceBundle messages, Language language, Word2VecModel word2vecModel) {
+    List<ScoredConfusionSet> confusionSets;
+    try(InputStream confusionSetsStream = new FileInputStream(word2vecModel.getPath() + File.separator + CONFUSION_SET_FILENAME)) {
+      confusionSets = ScoredConfusionSetLoader.loadConfusionSet(confusionSetsStream);
+    } catch (IOException e) {
+      System.err.println("Error: " + CONFUSION_SET_FILENAME + " not found for " + language.getShortCode() + ".");
```
danielnaber: Maybe just throw an exception?

languagetool-gui-commons/src/main/java/org/languagetool/gui/Tools.java
```diff
@@ -205,6 +202,11 @@ public static void configureFromRules(JLanguageTool langTool, Configuration conf
         langTool.disableCategory(new CategoryId(categoryName));
       }
     }
+    if(config.getEnabledRulesOnly()) {
```
danielnaber: Is this an unrelated bugfix?

danielnaber:
@gulp21 Are you working on the remarks I had? If not, we could also merge the PR now and fix the issues after merge.

languagetool-gui-commons/src/main/java/org/languagetool/gui/Tools.java
gulp21: I forget to mention that in the summary: I have added the possibility to only enable the rules which are explicitly mentioned in the configuration file. Afaik that was not possible before. I'm using that for the demo page where only neural network rules are enabled.

languagetool-core/src/main/java/org/languagetool/rules/neuralnetwork/Dictionary.java
```diff
@@ -0,0 +1,35 @@
...
+      String[] kv = entry.split(": ");
+      put(kv[0].substring(1, kv[0].length() - 1), Integer.parseInt(kv[1]));
+    }
+  }
+
+  Integer safeGet(String key) {
+    if (containsKey(key)) {
+      return get(key);
+    } else {
+      System.out.println(key + " unknown");
```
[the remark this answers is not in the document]
gulp21: Yeah, can be removed. We don't have a logging system in LT, do we?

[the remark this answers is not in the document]
danielnaber: No, it was once planned, but for now, we assume that we're a library: either everything is fine, or if it's not, we can throw an exception instead of logging that something might be wrong.

languagetool-core/src/main/java/org/languagetool/rules/neuralnetwork/NeuralNetworkRuleCreator.java
gulp21: I think that throwing an exception (which is not caught anywhere else) implies that a user always needs all language models if they use any language model. E. g. I could not use word2vec for German only, and don't use it for English, because there would be an uncaught exception when loading the rules for English.

danielnaber: I see. In that case, I suggest you test for the existence of the file and print a warning if it's not found. If it is found, the IOException should be thrown (as IOException or in a RuntimeException), as then there's another reason for it than "file not found".

gulp21:
I think all points are now addressed.

danielnaber:
Thanks, this is merged now. Could you maybe also add a few sentences to `CHANGES.md`?

PR #862, opened 2017-12-15
[sr] Updated Serbian dictionaries; introduced EkavianSerbian and JekavianSerbian

strn:


danielnaber:
Thanks, could you have a look here, as the test cases fail:

```
[ERROR]   RuleTest.testJavaRules:55->testExamples:80->testIncorrectExamples:102 Did not get the expected rule match for the incorrect example sentence:
Text: Двије сам зивједзе видјела.
Rule: MORFOLOGIK_RULE_SR_HR
Matches: [] expected:<1> but was:<0>
```

strn:
Unable to resolve problem, will sync with main tree and create another pull request.

languagetool-language-modules/sr/src/main/java/org/languagetool/language/JekavianSerbian.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+import java.util.ResourceBundle;
+
+
+/**
+ * Class modelling Serbian Jekavian dialect
+ *
+ * @since 4.0
+ */
+public class JekavianSerbian extends Serbian {
```
danielnaber: I only noticed this now, but this class has no `getName()` method, is that on purpose?

strn: Yes, it is on purpose - no one should use "JekavianSerbian" class directly. This class models Jekavian dialect, that is basic for Serbian language spoken in Montenegro, Bosnia and Herzegovina and Croatia. Dialects are mostly the same, but there are differences between them.

danielnaber: Can it be made abstract then maybe, to prevent users from using it directly?

strn: I made the class abstract; it will be visible in next pull request.

PR #869, opened 2017-12-29
Lo fr 171229

OswaldLuc:
Validity extension of gender and number agreement rules in French.
Not only at the beginning of a sentence, but also after a punctuation mark.


danielnaber:
@dpelle, could you have a look?

danielnaber:
I've merged your changes, thanks! Sorry it took some time.

danielnaber:
Could you please check https://languagetool.org/regression-tests/20180108/result_fr_20180108.html - it's the list of new matches after your changes have been applied. Could you check if these are valid matches or false alarms?

PR #879, opened 2018-01-14
Refinement of last request (lo-fr-171229 #869)

OswaldLuc:
The last request was about :
Validity extension of gender and number agreement rules in French. 
And this one is an answer to :
[ https://languagetool.org/regression-tests/20180108/result_fr_20180108.html ](https://languagetool.org/regression-tests/20180108/result_fr_20180108.html)


danielnaber:
Thanks!

danielnaber:
Here's the result: https://languagetool.org/regression-tests/20180114/result_fr_20180114.html - it's in diff format, i.e. a minus at start of line means that match has been removed.

PR #888, opened 2018-02-04
[WIP] Start adding Hebrew support

amire80:
This is a very naïve first attempt to add Hebrew support. It probably needs seriously cleanup before merging, but appears to work already with two simple rules.

This was done with excellent help from @jaumeortola at the Wikimedia Hackathon in Olot (precursor to the Wikimedia Hackathon 2018 in Barcelona).


danielnaber:
I'm cleaning up old PRs, please re-open if there's new actitivy.

PR #900, opened 2018-02-13
[es] add Wikipedia rule

davidemdot:
Hello,

This PR adds a Wikipedia rule for the Spanish module.


danielnaber:
Thanks, could you also add a small test, like `PortugueseWikipediaRuleTest`?

Issue #58, opened 2014-02-06
Visualisation of false friends per language


PanderMusubi:
I need a visualisation of false friends in the same way of
https://github.com/languagetool-org/languagetool/blob/master/languagetool-core/src/main/resources/org/languagetool/rules/false-friends.css
but show only the rulegroups in which a certain language is used.

The list of false friends is getting longer and longer so being able to filter on rulegroups only related to the language you are working on is practical.

Ideas for implementation are (one or more):
- parameter attached to the URL when viewing false-friends.xml in a web browser such as ?lang=all (default), ?lang=en, ?lang=de, etc.
- adding a HTML 1-of-n select that has the values all (default), en, de, nl, etc. on top of the view

PR #901, opened 2018-02-14
Add HTML file for allowing to filter false friends by language (issue #58)

davidemdot:
Hey,

This PR is a tentative solution for the issue #58, labeled as "easy fix".  It doesn't remove the possibility of accessing to the `false-friends.xml`, but it adds a new file (`false-friends.htm`) that allows to filter by language, attaching a parameter to the URL in this way: `false-friends.htm?lang=en`. If there is no parameter provided, it shows all the elements.

Please tell me if this fit into the idea.

### DIFF
--- a/languagetool-core/src/main/resources/org/languagetool/rules/false-friends.htm
+++ b/languagetool-core/src/main/resources/org/languagetool/rules/false-friends.htm
@@ -1,0 +1,63 @@
+<!-- [False Friends - filter by language]
+
+  Allowing to filter false friends by a certain language,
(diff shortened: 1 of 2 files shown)


danielnaber:
Thanks! Maybe you could add some documentation directly to the HMTL file, to make the parameter feature less hidden?

davidemdot:
Sure! Hope everything is OK now.

danielnaber:
Thanks! I've moved the comment from being a comment to the HTML body, so everybody opening the file can see it.

Issue #652, opened 2017-01-02
PatternRules ignore postag in <match> when rule contains <or>


f-knorr:
When writing some new rules for English, I wrote the following rule:
```
<rule>
    <pattern>
        <or>
            <token postag="SENT_START"/>
            <token regexp="yes">what|which|who|whose|where|when|whether|how</token>
        </or>
        <token postag="DT"/>
        <token postag="NNS"><exception postag="NN|NN:.*" postag_regexp="yes"/></token>
        <marker>
            <token postag="VBZ"><exception>is</exception></token>
        </marker>
    </pattern>
    <message>You should probably use <suggestion><match no="4" postag="VBP"/></suggestion>.</message>
    <example correction="bark">The dogs <marker>barks</marker> loudly.</example>
    <example correction="bark">He asked whether the dogs <marker>barks</marker> loudly.</example>
</rule>
```
The test for this rules fails because the insertion of an inflected matched token (`<match no="4" postag="VBP"/>`). The reason for this seems to be the `<or>` within the rule: The PatternRuleHandler seems to treat the iinflection correclty only for the first token in the or-group. (see `suggestionMatches.clear();` in PatternRuleHandler.java)

bebras123:
Also see: https://forum.languagetool.org/t/wrong-match-suggestion-when-changing-postag-in-rules-with-or/2189

oserikov:
Working on fix.

oserikov:
Seems that I've fixed the problem with or patterns, it passes the case stated in https://forum.languagetool.org/t/wrong-match-suggestion-when-changing-postag-in-rules-with-or/2189.

But your case fails:
```
java.lang.AssertionError: English: Incorrect suggestions: [bark] != [Bark] for rule AGREEMENT_SENT_START[1] on input: The dogs barks loudly. 
Expected :[bark]
Actual   :[Bark]
```
The word is correctly corrected, but is also capitalized. May that be caused by `SENT_START`? Or here is another bug discovered?

f-knorr:
Great job, @oserikov ! Closed this issue.

danielnaber:
@f-knorr Would you also like to merge the PR? (#920)

PR #920, opened 2018-02-24
bugfix: issue #652

oserikov:
Fixed the bug mentionned [here](https://github.com/languagetool-org/languagetool/issues/652) and [here](https://forum.languagetool.org/t/wrong-match-suggestion-when-changing-postag-in-rules-with-or/2189).

It looks like @jaumeortola used a walkaround to deal with the strange behavior of OR-grouped patterns (sometimes the corrections were wrongly capitalized because of the bug) when writing rules for catalan.  I've removed these walkarounds (the rules containing the walkarounds looked unrelated to the capitalization of corrections), but need somebody (maybe @jaumeortola) to check that part.

When I was testing the bugfix I used to replace the english grammar rules with the rules stated in bugreports, but I've removed that now and dont know how to prove that the bug is fixed. Can somebody suggest any way to demonstrate that the fix is ok?


danielnaber:
@f-knorr As you originally reported this, maybe you can check this RP?

Issue #63, opened 2014-02-13
Create a JUnit test that would check if the messages and titles of rules in LanguageTool are correct


milekpl:
There may be typos in messages and rule titles, in particular in XML pattern rules. Create a check that would run LanguageTool on these messages. Note: some rule titles quote the error that they match, so the match of that very rule should be ignored.

Requires: knowledge of Java

czojo26:
Hi, I am Michał and I would like to contribute to this project. This is the first problem I would like to solve,
but I am not sure where this JUnit should be placed. In which module such test would fit?

danielnaber:
Thanks for your interest in LT. You could place the test in `languagetool-standalone`, like this test: https://github.com/languagetool-org/languagetool/blob/master/languagetool-standalone/src/test/java/org/languagetool/JLanguageToolTest.java#L40

janschreiber:
Maybe this excellent idea can be expanded to run checks on the (supposedly) correct example sentences as well?

ImmortalTurtle:
Hello! May I take this issue?

danielnaber:
> Hello! May I take this issue?

Sure! Let us know here or [on the forum](https://forum.languagetool.org/) if you have questions.

ImmortalTurtle:
My questins are rather concerning the understanding of LT basics.
There are rules in languagetool-core, which extend "Rule" class. What is title of a rule here? What does "in particular in XML pattern rules" mean? I want to understand what does **title** and **message** mean applied to **Rule** class.
And the task is to apply all rules to titles of each of these rules?

There's a lot of documentation, it would be wonderful if you could guide me what to explore firstly.
Any information would be appreciated. Thank you

danielnaber:
`title` is a string shown in the configuration dialog where users can enable/disable rules. `message` is the string (often with a variable) that is shown to the user when a potential error is found.

> And the task is to apply all rules to titles of each of these rules?

Basically yes, but as `title`s are not complete sentence, we need to see if it makes sense. Maybe some rules are just not useful for this use case.

ImmortalTurtle:
All right, I can tell a little about my intermediate results. I analysed descriptions of English rules with the rules themselves.

There are cases, when this approach finds errors, for example:
Phrase: *Space character at the begin of paragraph*
Rule: *a/the + infinitive*

[...]

PR #935, opened 2018-03-08
Check rules' messages by LanguageTool: issue #63

ImmortalTurtle:
Solving issue #63. 
*Firstly*, get message from `Rule` - run it on the first `IncorrectExample`
*Secondly*, run `LanguageTool`, i.e. all enabled Rules, on this message

Notice, that some matches are ignored:
 - If `RuleMatch` was created by the same `Rule` that we are checking - it doesn't count
 - If `RuleMatch` is in brackets or quotes - it doesn't count *but this is customizable*

Disabled by default rules are configurable, I disabled some that are the most frequent, but they still might carry useful information and suggest some reasonable fixes.
Some (probably most) of these matches don't make a lot of sense, but some certainly do.

I can start fixing Rules that this test found mistakes in, should I?

### DIFF
--- a/languagetool-standalone/src/test/java/org/languagetool/JLanguageToolTest.java
+++ b/languagetool-standalone/src/test/java/org/languagetool/JLanguageToolTest.java
@@ -18,6 +18,7 @@
  */
 package org.languagetool;
 
+import org.junit.Ignore;
 import org.junit.Test;
 import org.languagetool.language.*;
 import org.languagetool.markup.AnnotatedText;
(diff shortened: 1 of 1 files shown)


danielnaber:
Thanks, I've already used this to find errors in the German `grammar.xml`. The only change I made was adding `{...}` around single-line `if` statements, according to our coding style.

> I can start fixing Rules that this test found mistakes in, should I? 

Yes, please. Also, extending the unit test to mark the position of the error would be nice.

ImmortalTurtle:
>Also, extending the unit test to mark the position of the error would be nice.
Sure, will do. Should I do it in this PR or create another one? I'm not very experienced in this.

Also I'm not sure about editing the rules: should I just create pull request w\o issue?

danielnaber:
> Sure, will do. Should I do it in this PR or create another one? I'm not very experienced in this.

Please create a new one.

> Also I'm not sure about editing the rules: should I just create pull request w\o issue? 

Yes, no issue is needed for small changes.

Issue #746, opened 2017-07-20
SynthDictionaryBuilder: write xyz_tags.txt to better location


danielnaber:
Currently it's written to temp, which causes it to be ignored. See http://forum.languagetool.org/t/reverse-postag-lookup-please-help/1933/8. It should probably be written to the same directory as the `*.dict` file.

gatitoneku:
Hello, i'm new to open-source. Can i try to do this one?

danielnaber:
> Can i try to do this one?

Sure!

gatitoneku:
Hmm weird, i also tested the master branch and it failed the English module. Is it because of the recent commits?

danielnaber:
> Hmm weird, i also tested the master branch and it failed the English module. Is it because of the recent commits?

It's not because of your change I think, something else broke the build.

PR #944, opened 2018-03-14
updated tags file location

gatitoneku:
Attempted fix for #746, write in in the same directory as the .dict file instead of temp

### DIFF
--- a/languagetool-tools/src/main/java/org/languagetool/tools/SynthDictionaryBuilder.java
+++ b/languagetool-tools/src/main/java/org/languagetool/tools/SynthDictionaryBuilder.java
@@ -58,7 +58,9 @@ final class SynthDictionaryBuilder extends DictionaryBuilder {
   }
   
   File build(File plainTextDictFile, File infoFile) throws Exception {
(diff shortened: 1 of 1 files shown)


danielnaber:
Thanks. I think there should be a slightly cleaner way that works without `substring`, maybe `File.getParent()`?

danielnaber:
I assume you've not just changed the code but also testes manually that it works, haven't you? In that case I'd merge your change without testing it myself.

gatitoneku:
Yeah, i've tested running SynthDictionaryBuilder and it generated the tag file in the same location as the dict file.

danielnaber:
Thanks, I've merged your change now.

gatitoneku:
Oops i forgot to delete() the directory after usage.  Should i make a PR again?

danielnaber:
> Oops i forgot to delete() the directory after usage. Should i make a PR again?

Sorry, I'm not familiar with the flow now - which directory? There's no reason to delete the result, I guess? :-)

gatitoneku:
I used ```File outputDirectory = new File(outputFilename).getParentFile();``` to get the directory of the output files. What i meant was I forgot to delete the ```File``` object.

PR #949, opened 2018-03-17
Added collective noun verb agreement

thecuriousvector:
[Collective Noun Verb Agreement.pdf](https://github.com/languagetool-org/languagetool/files/1821291/Collective.Noun.Verb.Agreement.pdf)

Kindly, review the changes @danielnaber.

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
@@ -6859,47 +6859,47 @@ USA
             <example>You kids must be hungry.</example>
             <example>You guys wait here.</example>
         </rule>
-        <rule id="COLLECTIVE_NOUN_VERB_AGREEMENT_VBP" name="a fleet of ships are (is)">
+        <rule id='COLLECTIVE_NOUN_VERB_AGREEMENT_VBP' name='COLLECTIVE_NOUN_VERB_AGREEMENT_VBP'>
+ 		 <pattern>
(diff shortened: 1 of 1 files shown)


danielnaber:
Thanks. I've checked against more sentences, could you check whether these matches are false alarms and, if so, whether they can be avoided?

```
In pure water the majority of molecules are H2O, but...
Small percent of Afghans are also fluent in Arabic...
The majority of employees work within the service sector...
References to the concept of atoms date back to ancient Greece and India.
An overwhelming majority of atoms combine to form various compounds...
...orders the creation of the Badge of Military Merit to honor soldiers wounded in battle.
The lone pair of electrons repel more strongly than bond pairs...
...who also built a lesser temple of his own.
```

thecuriousvector:
@danielnaber Please review the changes
Initially, only the pattern of the form <NN of NNS -> was being checked. So, even if a word had multiple tags it allowed matching. Now, the pattern matching is made strict. This eliminated the following false errors:
__Small percent of Afghans are also fluent in Arabic...
...orders the creation of the Badge of Military Merit to honor soldiers wounded in battle.
The lone pair of electrons repel more strongly than bond pairs...
...who also built a lesser temple of his own._

Adding "majority" as an exception eliminated these two:
_The majority of employees work within the service sector...
An overwhelming majority of atoms combine to form various compounds..._
(Though this sounds like overfitting I couldn't find a better way. Also added "pair" as an exception. Ideally, the number of false errors associated with such numbered collective nouns should be reduced because "NN:UN" is not matched in the pattern anymore)

Unresolved agreement:
In the below example, "date" is a verb associated with the subject "References" rather than the collective noun. To resolve this, I would need dependency parsed tags in addition to POS tags. 
_References to the concept of atoms date back to ancient Greece and India._

Let me know in case of any further issues. I'll revisit.

danielnaber:
Thanks, I've merged your PR.

PR #956, opened 2018-03-20
#Issue 947 Please allow to reduce the list of available languages 

LAVEEN:


languagetool-standalone/src/main/java/org/languagetool/gui/LanguageComboBoxModel.java
```diff
@@ -44,7 +46,7 @@ static LanguageComboBoxModel create(ResourceBundle messages,
 
     LanguageComparator comparator = new LanguageComparator(messages, extLangSuffix);
     LanguageComboBoxModel model = new LanguageComboBoxModel();
-
+    languageComboBoxModel=model;
```
danielnaber: Please format with spaces around the `=`.

languagetool-standalone/src/main/java/org/languagetool/gui/PopUpRemoveLanguage.java
```diff
@@ -0,0 +1,89 @@
...
+          comboBoxModel.setSelectedItem(newElement);
+          comboBoxModel.removeElement(selectedElement);
+
+        }
+
+      }
+
+      @Override
+      public void mouseClicked(MouseEvent arg0) {
+        // TODO Auto-generated method stub
```
danielnaber: The TODOs should probably be removed?

languagetool-standalone/src/main/java/org/languagetool/gui/RightClickListener.java
```diff
@@ -0,0 +1,39 @@
...
+ * USA
+ */
+package org.languagetool.gui;
+
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
+
+class RightClickListener extends MouseAdapter {
+  public void mousePressed(MouseEvent e) {
+    if (e.isPopupTrigger())
```
danielnaber: Please add {} also for `if` blocks with only one statement.

languagetool-standalone/src/main/java/org/languagetool/gui/PopUpRemoveLanguage.java
```diff
@@ -0,0 +1,89 @@
...
+import java.awt.event.MouseListener;
+
+import javax.swing.JMenuItem;
+import javax.swing.JOptionPane;
+import javax.swing.JPopupMenu;
+
+class PopUpRemoveLangauge extends JPopupMenu {
+
+  private static final long serialVersionUID = 1L;
+  JMenuItem removeOption = new JMenuItem("Remove");
```
danielnaber: Does "Remove" appear in the GUI? Then it needs to be translatable.

languagetool-standalone/src/main/java/org/languagetool/gui/PopUpRemoveLanguage.java
```diff
@@ -0,0 +1,89 @@
...
+      @Override
+      public void mousePressed(MouseEvent e) {
+        LanguageComboBoxModel comboBoxModel = LanguageComboBoxModel.languageComboBoxModel;
+        int itemCount = comboBoxModel.getSize();
+        Object selectedElement = comboBoxModel.getSelectedItem();
+
+        int index = comboBoxModel.getIndexOf(selectedElement);
+        // If only one language is left,we cannot remove it
+        if (itemCount == 1) {
+          JOptionPane.showMessageDialog(null, "Selection Cannot Be Removed");
```
danielnaber: This needs to be translatable.

danielnaber:
I have only looked at the code and not actually tested this - do the changes get saved to the configuration? I guess they should? Also, can languages be re-added?

languagetool-standalone/src/main/java/org/languagetool/gui/PopUpRemoveLanguage.java
LAVEEN: yes it should be removed.Will do that.

languagetool-standalone/src/main/java/org/languagetool/gui/RightClickListener.java
LAVEEN: ok

LAVEEN: @danielnaber Language once added can't be removed.

danielnaber:
Please indent according to http://wiki.languagetool.org/code-style. Are the changes saved now? Sorry, I should have mentioned you only need to translate the English "messages" file, any other languages files will be overwritten from Transifex.

LAVEEN:
@danielnaber I have updated all the properties files by adding the messages in their respective language,is that ok.

danielnaber:
You mentioned the settings are saved now, didn't you? Can you point me to the code that does this? Also, how can languages be re-added.

languagetool-standalone/src/main/java/org/languagetool/gui/PopUpRemoveLanguage.java
```diff
@@ -0,0 +1,78 @@
...
+public class PopUpRemoveLanguage extends JPopupMenu {
+
+  private final ResourceBundle messages;
+  private final JMenuItem removeOption;
+  private final LanguageAdapter languageAdapter;
+
+  public PopUpRemoveLanguage() {
+    languageAdapter = (LanguageAdapter) LanguageComboBoxModel.languageComboBoxModel.getSelectedItem();
+    messages = JLanguageTool.getMessageBundle(languageAdapter.getLanguage());
+    removeOption = new JMenuItem(messages.getString("guiRemoveOptionText"));
```
LAVEEN: Text Value are taken from there resource bundle

languagetool-standalone/src/main/java/org/languagetool/gui/PopUpRemoveLanguage.java
```diff
@@ -0,0 +1,78 @@
...
+    LanguageAdapter languageAdapter = (LanguageAdapter) comboBoxModel.getSelectedItem();
+    ResourceBundle messages = JLanguageTool.getMessageBundle(languageAdapter.getLanguage());
+    int itemCount = comboBoxModel.getSize();
+    Object selectedElement = comboBoxModel.getSelectedItem();
+
+    int index = comboBoxModel.getIndexOf(selectedElement);
+    // If only one language is left,we cannot remove it
+
+    if (itemCount == 1) {
+      JOptionPane.showMessageDialog(null, "<html>" + messages.getString("guiRemoveWarningMsg") + "</html>", "",
```
LAVEEN: Text Value are taken from there resource bundle

languagetool-standalone/src/main/java/org/languagetool/gui/PopUpRemoveLanguage.java
```diff
@@ -0,0 +1,78 @@
...
+    removeOption = new JMenuItem(messages.getString("guiRemoveOptionText"));
+    add(removeOption);
+    removeOption.addMouseListener(new OnClickListener());
+  }
+}
+
+class OnClickListener implements MouseListener {
+  @Override
+  public void mousePressed(MouseEvent e) {
+    LanguageComboBoxModel comboBoxModel = LanguageComboBoxModel.languageComboBoxModel;
```
LAVEEN: ComboBoxModel will hold all the settings.So,the set of languages are taken from here and are removed from here only.If you want all the languages that are deleted,we can just find set A - set B where set A=the original list which is unmodifiable and which we get at Languages.get() and set B is the set coming from  by iterating the elements in comboBox [model.getElementAt(model.getSize())].

danielnaber:
> I think you mean to say whether we are storing the deleted languages and original languages,but we are not saving either of them.

These two changes will be needed before we can merge this: 1) store the user's changes in the configuration so user's changes are kept after a restart. 2) make it possible to re-add languages

LAVEEN:
Ok i will make those changes.Thanks for the review

LAVEEN:
Ok i will make those changes

danielnaber:
I think this was never finished, was it? Please re-open if the final comments are addressed.

Issue #872, opened 2018-01-04
Readability: word counts are not correct for section numbers and similar


MikeUnwalla:
In this example, the number of words is 15, but LT gives a warning:

![image](https://user-images.githubusercontent.com/5816160/34565181-aea0f70a-f151-11e7-893d-ba6d584713eb.png)

danielnaber:
That's because `8.5.a` etc. gets split into several tokens. Maybe we should ignore tokens that are just a number?

MikeUnwalla:
> Maybe we should ignore tokens that are just a number?
If it is not possible to change the rule such that these tokens are treated as 1 token, I suggest that we live with the false warnings. (I think false positives are less bad than false negatives.)

danielnaber:
> If it is not possible to change the rule such that these tokens are treated as 1 token

Somehow that's possible, but it seems complicated, and the next user will come up with yet another way to write numbers (`8-5-c` or whatever) that won't be covered then.

f-knorr:
I suppose the rule counts each token as one word. I think a more intuitive approach to count the spaces (regexp "\s+") and to interpret each (sequence of) space(s) a word (delimiter).
With this approach the above examples would all count as one word.

Demfier:
@danielnaber @f-knorr -  I would like to take up this issue.

If you don't mind, can you please guide me on which part of the codebase to touch here? This is my first Java project and it would be really great if you can guide me through it. From what I saw in the codebase, I thought one of the [tokenizers](https://github.com/languagetool-org/languagetool/tree/master/languagetool-core/src/main/java/org/languagetool/tokenizers) has to be changed here?

Thank you.

f-knorr:
A good starting point point is the rule that causes the undesired bahavior: `LongSentenceRule`
The rules "sees" an error, when a match is added to ruleMatches:
```
    if (tokens.length < maxWords + 1) {   // just a short-circuit
      return toRuleMatchArray(ruleMatches);
    } else {
      for (AnalyzedTokenReadings aToken : tokens) {
        String token = aToken.getToken();
//        pos += token.length();  // won't match the whole offending sentence, but much of it
        if (!aToken.isSentenceStart() && !aToken.isSentenceEnd() && !NON_WORD_REGEX.matcher(token).matches()) {
          numWords++;
        }
      }
    }
    if (numWords > maxWords) {
      RuleMatch ruleMatch = new RuleMatch(this, sentence, 0, pos, msg);
[...]

PR #964, opened 2018-03-25
Issue #872 : Readability word counts are not correct for section num…

LAVEEN:
Issue occured due to words like **8.6.a** which gets tokenized into three tokens.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/tokenizers/WordTokenizer.java
+++ b/languagetool-core/src/main/java/org/languagetool/tokenizers/WordTokenizer.java
@@ -270,7 +270,7 @@ public class WordTokenizer implements Tokenizer {
         }
       }
 
-      if (count == noOfMatch && executed == true)
+      if (count == noOfMatch && executed)
         continue;
       
       // Last Token Case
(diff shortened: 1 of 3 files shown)


languagetool-core/src/main/java/org/languagetool/tokenizers/WordTokenizer.java
```diff
@@ -217,5 +218,104 @@ private boolean urlEndsAt(int i, List<String> l, String urlQuote) {
...
+            List<String> backList = tokens.subList(index + 6, tokens.size());
+            frontList.addAll(backList);
+            tokens = frontList;
+            executed = true;
+          }
+
+        }
+      }
+
+      if (count == noOfMatch && executed == true)
```
f-knorr: Little remark: Comparing a boolean variable to true or false (`executed==true`) is redundant. Either it is true, or it is false.

LAVEEN: Yeah ,you are right.Thanks for pointing this

danielnaber:
I'm closing old PRs that have conflicts, please open a new one if you want this to be merged.

PR #978, opened 2018-04-14
[zh] Update zh translations

gensitu:
* Separate into zh-hans and zh-hant
* Correct punctuation error


danielnaber:
Thanks, have you tried that this actually works, by testing the LibreOffice add-on with this change?

danielnaber:
What's the advantage of having separate files like `desc_en.txt`?

gensitu:
I submitted new commits to make use of LibreOffice's extension documentation which states that descriptions must be linked with `src` and cannot be added within `description.xml`. The problem with the current setup (including the description with the extension name) is that the extension version is hidden because the name is too long.

gensitu:
I have tested this with several language packs and it seems to work.

danielnaber:
Have you also tested this with OpenOffice?

gensitu:
Also tested with OpenOffice. The LO extension schema is identical to the AOO extension schema, so everything works the same.

PR #988, opened 2018-04-25
[de] don't accept "Auftragsport"

simfi:


danielnaber:
Thanks, but is this a misspelling of a close word?

PR #1002, opened 2018-05-22
Chinese Module update

hyousi:
- Separate Chinese Language to Simplified Chinese(zh-CN) and Tradition Chinese(zh-TW).
- Substitute [ictclas4j](https://code.google.com/archive/p/ictclas4j/) by [HanLP](https://github.com/hankcs/HanLP).
- Maintain the `grammar.xml` and create one for zh-TW.


danielnaber:
The test fail on the CI server because it doesn't have the data. Can you change the tests so that they work on small data (less than 10MB) which can then be included in git? Or if that's not possible, skip the tests if the data isn't there and print a warning?

languagetool-language-modules/zh/src/main/resources/META-INF/org/languagetool/language-module.properties
```diff
@@ -1 +1 @@
-languageClasses=org.languagetool.language.Chinese
+languageClasses=org.languagetool.language.Chinese, org.languagetool.language.SimplifiedChinese, org.languagetool.language.TraditionalChinese
```
danielnaber: org.languagetool.language.Chinese can probably be removed here, can't it? If everybody always uses Simplified or Traditional, there's no need for it?

hyousi:
Sorry...The data is used for segmenting sentences and tagging words. Without these data, even the basic function can't work.

TiagoSantos81:
Looking at this merge request, it is noticeable that the module is being build from scratch instead of being a cumulative work. Look at the commit summary: `+9,126 −17,065`, and the files changed.
The former work should be integrated in the Simplified variant and adapted/corrected to the new tokenizer if needs. 
Otherwise, this is just a regression and it is disrespectful to the former contributers.

TiagoSantos81:
Another thing is, the new rules in zh-CN are simple replace rules. You can made them more easily with a Java rule extending  `AbstractSimpleReplaceRule`, then add the replacement table to the chosen database file.
Notice that most of the already existing rules in `./zh/grammar.xml` use regular expressions to match multiple situations

danielnaber:
@TiagoSantos81 I think the new rules are more compact due to changed tokenizing and that's the reason many lines were removed. @hyousi Can you confirm?

TiagoSantos81:
@danielnaber No they aren't. And the record from former contributors is being erased. 
Look at the header of the grammar.xml files[ in our repo](https://github.com/languagetool-org/languagetool/blob/master/languagetool-language-modules/zh/src/main/resources/org/languagetool/rules/zh/grammar.xml)  and the header of pulled version of [Simplified Chinese, zh-CN](https://github.com/hyousi/languagetool/blob/838973c8d7df3a99ea68725ff8cc75ed495e02e1/languagetool-language-modules/zh/src/main/resources/org/languagetool/rules/zh/zh-CN/grammar.xml) or the new [Traditional Chinese, zh-TW](https://github.com/hyousi/languagetool/blob/838973c8d7df3a99ea68725ff8cc75ed495e02e1/languagetool-language-modules/zh/src/main/resources/org/languagetool/rules/zh/zh-TW/grammar.xml). Also look at the XML rules. They are token rules that match strings, not postags, so the new tokenizer has no effect. I don't even get why it is needed given the rules that were posted.

danielnaber:
@hyousi Please add these lines from the old file: https://github.com/languagetool-org/languagetool/blob/master/languagetool-language-modules/zh/src/main/resources/org/languagetool/rules/zh/grammar.xml#L5-L9

hyousi:
@danielnaber @TiagoSantos81  Thanks for your advice.  I have added the lines now.  I removed many lines because I changed the tokenizer. The tokenizer actually does two things the same time. First, it segments words from a sentence. Then, it tags each word. 
Since Chinese is a little different from western languages. It doesn't have morphological information and blank between each word. So it is a challenge for Chinese to segment words and tag words. 
Here is an example from the `grammar.xml` which may make it more clear.

晶莹 - Word(词语) - sparkling and crystal-clear
剔透 - Word(词语) - bright and limpid
晶莹剔透 - Chinese Idiom(成语) - crystal-clear
A sentence`这晶莹剔透的露珠太美丽了。` with no errors can be split into multiple cases.
1. 这/晶莹/剔透/的/露珠/太/美丽/了/。 
2. 这/晶莹剔透/的/露珠/太/美丽/了/。  <- most common 
3. 这/晶莹剔透/的/露珠/太美丽了/。
It is clear that tokenizing a word depends on both its front characters and its back characters.
When the sentence has an error. `这晶萤剔透的露珠太美丽了`.Tokenizing may be changed.
In the past LT's tokenizer :  
这/晶/萤/剔透/的/露珠/太/美丽/了/。
In my case: 
这/晶萤剔透/的/露珠/太/美丽/了/。
According to different segments, POS is different, And different POS may cause different segment. 

In my GSoC proposal, the idea of improving Chinese is to keep the wrong sentences' segments and tag like the correct one as much as possible.  Then correct the sentences.

hyousi:
@TiagoSantos81 Thank you again. 
There are 5 categories in the `grammar.xml`.  词语错误 and 成语错误 are something like word misspelling in English. And the rest of them are something like "grammars". But they are not totally the same. In the next deliverable, I will submit a rule to replace most of the 词语错误 and 成语错误 rules in the grammar table. 
But some of grammar rules should be kept because what I have said above. 
> According to different segments, POS is different, And different POS may cause different segment.
I need the POS to detect errors too.

hyousi:
@TiagoSantos81  Thank you. I got it. Sorry, this is the first time I use git and plan to join a community (I am a sophomore now...). Everything is new to me, and I am willing to learn it.  
The most of rules using `<regexp/>` are temporary in my opinion. I will fix them next. Except this, is there anything else I need to notice?

TiagoSantos81:
> Thank you. I got it. Sorry, this is the first time I use git and plan to join a community (I am a sophomore now...). Everything is new to me, and I am willing to learn it.
You are doing great. Sorry if I was harsh on the comments, but I was seeing this slipping by.
> The most of rules using <regexp/> are temporary in my opinion. I will fix them next. Except this, is there anything else I need to notice?
I haven't noticed anything else I should mention. Just build upon what was done and remember to reference former contributors, and no matter how much it is done, you will already be pushing the envelop.

languagetool-language-modules/zh/src/main/java/org/languagetool/language/SimplifiedChinese.java
```diff
@@ -0,0 +1,15 @@
+package org.languagetool.language;
```
danielnaber: Please add our standard license header here as in other *.java files (but with your name). Same for all other new *.java files.

languagetool-language-modules/zh/src/main/java/org/languagetool/rules/zh/TraditionalChineseAmbiguityRule.java
```diff
@@ -0,0 +1,55 @@
...
+public class TraditionalChineseAmbiguityRule extends Rule {
+
+  @Override
+  public String getId() {
+    return "TC_AMBIGUITY_RULE";
+  }
+
+  @Override
+  public String getDescription() {
+    return "A rule detects ambiguities in words.";
```
danielnaber: This description should mention TC and SC, I guess.

languagetool-language-modules/zh/src/main/resources/org/languagetool/rules/zh/zh-CN/grammar.xml
```diff
@@ -0,0 +1,7314 @@
...
+                </pattern>
+                <message>
+                    您的意思是"托运"吗？
+                </message>
+                <example correction="">我要<marker>拖运</marker>包裹。</example>
+                <example>我要托运包裹。</example>
+            </rule>
+            <rule>
+                <pattern>
+                    <ma
```
danielnaber: I see there are many rules without a `<suggestion></suggestion>` and thus with empty `correction` attributes. Is this on purpose? Is there no way to offer a suggestion for these rules?

languagetool-language-modules/zh/src/main/java/org/languagetool/tokenizers/zh/ChineseWordTokenizer.java
```diff
@@ -19,42 +19,25 @@
...
-  private ChineseJF chinesdJF;
+  private PerceptronLexicalAnalyzer seg;
 
   private void init() {
-    if (chinesdJF == null) {
-      chinesdJF = CJFBeanFactory.getChineseJF();
-    }
     if (seg == null) {
-      ResourceDataBroker dataBroker = JLanguageTool.getDataBroker();
-      try (InputStream coreDictIn = dataBroker.getFromResourceDirAsStream("/zh/coreDict.dct");
```
danielnaber: So these files are not needed anymore, is that correct? If so, please remove them from git (or did you? I didn't see their removal being part of the PR).

danielnaber: Mhh, maybe I just commented on an old PR? Not sure, if so, just ignore this...

languagetool-language-modules/zh/src/main/resources/org/languagetool/rules/zh/zh-CN/grammar.xml
hyousi: In the old version which maintained by Tao Lin, lots of rules are formed like you mentioned. So I keep it, but I think part of them should actually have a `<suggestion></suggestion>`. So should I fix it?

danielnaber: Yes, please add  a suggestion where it makes sense.

danielnaber:
Solved by https://github.com/languagetool-org/languagetool/pull/1981, I guess, so I'm closing this.

PR #1009, opened 2018-05-25
[en] fix outdated URLs in grammar.xml

junglast:
Hello.

I examined almost all URLs in the English grammar.xml and fixed some unavailable ones.


danielnaber:
Thanks!

PR #1047, opened 2018-07-09
Various fixes to Swedish grammar.xml

ordtrogen:
Spelling errors, corrected capitalisation, shortened some sentences to save space in a GUI, corrected some prepositions, changed some wordings

### DIFF
--- a/languagetool-language-modules/sv/src/main/resources/org/languagetool/rules/sv/grammar.xml
+++ b/languagetool-language-modules/sv/src/main/resources/org/languagetool/rules/sv/grammar.xml
@@ -24,35 +24,35 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
   <!-- ====================================================================== -->
   <!-- Särskrivningar -->
   <!-- ====================================================================== -->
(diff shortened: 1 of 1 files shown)


danielnaber:
Thanks! Could you change the `id` "KommaFöreAnfoeringsverb" so that it doesn't contain an "ö"? We try to avoid special characters in rule ids.

danielnaber:
The build currently fails with your changes, could you run `./testrules.sh` (Linux) or `./testrules.bat` (Windows) to see the issue?

ordtrogen:
Now it should be fixed, modulo any squashing and rebasing and whatnot that typically gets me into trouble.

TBH I have only used LanguageTool from within OmegaT where it's integrated as a plugin. There were a few typos there that had bothered me for a long time and finally I got round to going to the source and fixing them. And now I have the Maven environment set up so any future work should be easier.

I'll be looking at the Android app next, which I didn't know about.

(And let me quietly object to "ö" being referred to as a special character :-)  Probably a good idea to disallow it to avoid problems, but technically it's allowed in rule ids, according to the schema definition, right?)

danielnaber:
Thanks, I've merged your change. Further contributions would be very welcome.

PR #1132, opened 2018-08-02
Added some lines to README.md

NikaBan:


danielnaber:
Mhh, I don't think that wording would help users. The point of LT is that it goes beyond spell checking, and that's true for all languages it supports. How far it goes beyond depends on how well the language is supported.

PR #1133, opened 2018-08-02
Added punctuation and fixed typos.

NikaBan:


danielnaber:
Thanks, but `agid-readme.txt` is just copied from http://wordlist.aspell.net/agid-readme/ so it doesn't make much sense to edit it here in our repo.

PR #1164, opened 2018-08-29
fix for encoding errors in IPC with fasttext

fabrichter:


fabrichter:
Deployed & tested on HE3, works

PR #1208, opened 2018-10-04
(Grammar Rules) Gerund Improvement

jariojose:
Implementations of rules to improve detection for gerund

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
@@ -10741,6 +10741,26 @@ USA
           <example type="incorrect">I am <em>searched</em> through some fields.</example>
         </rule>
         -->
+        <rule id="I_A_NONBASE" name="'I am' + non-gerund form">
+            <pattern>
+                <token postag="PRP"/>
+                    <marker>
+                        <token inflected="yes">be</token>
+                        <or>
+                            <token postag_regexp="yes" postag="VB[DNPZ]?">
+                                <exception>delayed</exception>
+                            </token>
+                            <token>count
+                                <exception>delayed</exception>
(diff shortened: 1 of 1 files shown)


MikeUnwalla:
@jariojose , thank you for your contribution.

Unfortunately, the rule gives many incorrect warnings (false positives). Some examples:
I was accepted as a new member
He describes a state in which you are open to everything you see and hear...
... we participate in an illusion whereby we are deceived by elites into thinking we have a genuine...
... as an anarchist is quickly voided when it is shown that he only wants an end to the public ...
It was founded in 1936 by Lucía Sánchez Saornil,...
She was published in a variety of literary journals where work...
... though he is correct in recognising the difference between ...

To see only what we want to see is very easy. This is what I do to find false positives:

- Use the corpora on https://corpus.byu.edu/ to find counter-examples.
- Test the rule on a small corpus: http://wiki.languagetool.org/tips-and-tricks#toc11 (that's how I found the false warnings). If you want the test files that I use (a tiny part of the Wikipeda corpus), just ask.

Also, the exception on the token 'count' is not necessary. Travis gives this message:
`The English rule: I_A_NONBASE[1] has exception word [delayed] in token word [3] [count] which seems useless, or did you forget skip="..." or scope="previous"?`

@danielnaber , is it possible and sensible to somehow make these messages very clear? (When I use  'testrules' (http://wiki.languagetool.org/development-overview#toc7), the problem is easy to see, but to search the Travis checks to find this type of error is not easy. Although the code is correct in that it does the job, as the message says, the exception is useless,)

jariojose:
Thank you, Mike for your feedback! I will test it. Sorry for this mistake.

jariojose:
@MikeUnwalla Who reverts this modification?

danielnaber:
@jariojose The PR hasn't been merged yet, so no need to revert anything.

PR #1232, opened 2018-10-25
setup aho corasick search in static block of ProhibitedCompoundRule

fabrichter:


fabrichter:
Usage of parseText from multiple threads is safe: https://github.com/hankcs/AhoCorasickDoubleArrayTrie/issues/20#issuecomment-433497221

PR #1235, opened 2018-10-26
Fix no case bug.

aha-oretama:
I think that this is a bug.

`SpellingCheckRule#isIgnoredNoCase` means that case is ignored in spellchecking, in other words, case-no-sensitive.

To realize above, comparing word and compared word are both change to lower case.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/spelling/SpellingCheckRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/spelling/SpellingCheckRule.java
@@ -164,7 +164,7 @@ public abstract class SpellingCheckRule extends Rule {
 
   private boolean isIgnoredNoCase(String word) {
     return wordsToBeIgnored.contains(word) ||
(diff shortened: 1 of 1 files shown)


danielnaber:
Thanks, could you give an example what this change means from the user's point of view?

aha-oretama:
For example, dictionary includes URL, XML. And user's input words are url, xml.
After this pull request is merged, spell check treats these words as correct words, and doesn't point out these words.

danielnaber:
I wonder whether that new behavior is really what we want. Is e.g. "xml" correct even if it's not explicitly listed? Maybe @MikeUnwalla has an informed opinion on that...

aha-oretama:
> @aha-oretama seems to want to ignore typographic conventions. That's fine (I don't know the use case.) Let's add a way to do that.

Thank you for  considering a way to ignore typographic conventions.

I think that we keep `#isIgnoredNoCase` method and add a new feature of adding the lowercase word  to the dictionary if a specified value sets true.

PR #1236, opened 2018-10-26
Add missing variable declarations for JavaScript code

stweil:
This fixes several error messages like the following one from LGTM:

    Variable xhttp is used like a local variable,
    but is missing a declaration.

Signed-off-by: Stefan Weil <sw@weilnetz.de>


stweil:
See https://lgtm.com/projects/g/languagetool-org/languagetool/ for details.

PR #1238, opened 2018-10-29
Added limit on queue size for database logging

fabrichter:


danielnaber:
I think we should log if MAX_QUEUE_SIZE is reached (as a warning).

PR #1252, opened 2018-11-07
added new option and rule category for rules relying on neural networks

fabrichter:
Added option `neuralNetworkModel` allows providing a base directory of saved neural network models.
Rules listed by the new `Language.getRelevantNeuralNetworkModels` should expect any exported models they require to be found there.

### DIFF
--- a/languagetool-commandline/src/main/java/org/languagetool/commandline/CommandLineParser.java
+++ b/languagetool-commandline/src/main/java/org/languagetool/commandline/CommandLineParser.java
@@ -96,6 +96,9 @@ public class CommandLineParser {
       } else if (args[i].equals("--word2vecmodel")) {
         checkArguments("--word2vecmodel", i, args);
         options.setWord2VecModel(new File(args[++i]));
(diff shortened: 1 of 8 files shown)


languagetool-commandline/src/main/java/org/languagetool/commandline/CommandLineParser.java
```diff
@@ -222,6 +225,7 @@ public void printUsage(PrintStream stream) {
             + "                           ngram occurrence counts; activates the confusion rule if supported\n"
             + "  --word2vecmodel DIR      a directory with e.g. 'en' sub directory (i.e. a language code) that contains\n"
             + "                           final_embeddings.txt and dictionary.txt; activates neural network based rules\n"
+            + "  --neuralnetworkmodel DIR a base directory for various saved neural network models\n"
```
zuxfoucault, line 228: Thanks for the work!  I could find an introduction for [--word2vecmodel](https://github.com/gulp21/languagetool/tree/neuralnetworkrule), but how to use `--neuralnetworkmodel`? What kind of file type it is expected?

PR #1258, opened 2018-11-12
updated UserConfig .equals/hashCode

fabrichter:


languagetool-core/src/main/java/org/languagetool/UserConfig.java
```diff
@@ -105,38 +109,34 @@ public LinguServices getLinguServices() {
...
-    if (dictIdentifier != null && that.dictIdentifier != null) {
-      return dictIdentifier.equals(that.dictIdentifier);
-    } else {
-      return userSpecificSpellerWords.equals(that.userSpecificSpellerWords);
-    }
+    // -> we use user id & dictionary names
+    return new EqualsBuilder()
+      .append(configurableRuleValues, other.configurableRuleValues)
+      .append(userDictName, other.userDictName)
+      .append(userSpecificSpellerWords, other.userSpecificSpellerWords)
```
danielnaber: This is not in hashCode, is that on purpose? maxSpellingSuggestions is only in hashCode but not here?

languagetool-core/src/main/java/org/languagetool/UserConfig.java
```diff
@@ -105,38 +109,34 @@ public LinguServices getLinguServices() {
     return linguServices;
   }
 
+  /**
+   * @since 4.3
```
danielnaber: Next version will be 4.4

languagetool-core/src/main/java/org/languagetool/UserConfig.java
fabrichter: See comment in hashCode for userSpecificSpellerWords; I simply forgot maxSpellingSuggestions

PR #1333, opened 2019-01-03
Wrapped requests to hidden matches server in a Future

fabrichter:


danielnaber:
@fabrichter this is outdated, isn't it?

PR #1368, opened 2019-01-19
[de] add words

tiff:
@janschreiber what do you think?

### DIFF
--- a/languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/hunspell/names.txt
+++ b/languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/hunspell/names.txt
@@ -970,3 +970,9 @@ Zolleiche
 Zolpidem
 ZoneRadio
 Zytglogge
+Franzi/S
+Caren/S
+Karen/S
+Mats
+Michi/S
+Roxana/S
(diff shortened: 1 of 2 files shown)


janschreiber:
@danielnaber doesn't want names in the list.

janschreiber:
I have made the file
https://github.com/languagetool-org/languagetool/blob/90e72217d6e4d9cdd3c9d477b31dc9bb5be97d56/languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/hunspell/names.txt
but it isn't used (yet).

tiff:
@danielnaber how do we deal with names? I think that if we whitelist company names (Facebook, WhatsApp, Samsung, ...) across languages then person and city names should also be included.

danielnaber:
Additionally accepting the words from `names.txt` requires some code changes that are a bit tricky. We could add thew words to `spelling.txt` with `#name` as a tag. But there are some words which are very close to other words, so adding them can easily mask real errors, e.g.

Amper -> Ampere
Holl -> hol (Imperativ "holen")
Imke -> Imker

tiff:
I moved the names to names.txt. Feel free to merge. @danielnaber I like the idea. Shall we exclude words that could potentially mask real errors? If so who takes care?

danielnaber:
I'm okay with that. Thanks!

PR #1396, opened 2019-02-01
Make editorconfig user-friendly

rico-chet:
Editorconfig provides the concept of cascading rules, refer to [the manual](https://editorconfig.org/#file-location). For instance, I use a rule which limits line length in git commit messages which enables `vim` to re-flow the text by a simple command. Setting `root=true` in LT's own `.editorconfig` file prevents any such user-defined rules from being propagated.

The `.editorconfig` file declares that any files should have a terminating newline but fails to adhere to the rule itself, fix it.


danielnaber:
Thanks!

PR #1407, opened 2019-02-04
Fix for suggesting "analyse" as a replacement for "analyse"

DaveChild:


MikeUnwalla:
Dave, many thanks for your fix. 

The simplest option to solve the immediate problem was to remove 'analyse' and 'analyser' from the regexp, because they are correct BrE. The z spelling is AmE. I made that change in https://github.com/languagetool-org/languagetool/commit/99f43fe0d14da0c9dd3319b3426bd2ed9b834e4d

@TiagoSantos81 is dealing with this rule, so I will let him deal with your pull request.

TiagoSantos81:
@DaveChild @MikeUnwalla I believe a recent commit from Mike has solved this issue. Many thanks to both. I will close this now.

DaveChild:
Thanks. There's also `paralyse` in there with the same problem.

TiagoSantos81:
@DaveChild Thank you! I ended up looking into both solutions, and I also integrated yours since it adds functionality instead f removing it.

PR #1425, opened 2019-02-13
Update en_US.info

surekha1317:
I want to make spelling check case insensitive.

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/hunspell/en_US.info
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/hunspell/en_US.info
@@ -13,3 +13,4 @@ fsa.dict.speller.replacement-pairs=ninties 1990s, a ei, ei a, a ey, ey a, ai ie,
 fsa.dict.frequency-included=true
 fsa.dict.speller.ignore-all-uppercase=false
 fsa.dict.speller.ignore-camel-case=false
+fsa.dict.speller.convert-case=true


MikeUnwalla:
Doesn't case-insensitive spelling mean that LT will not find errors such as the following?
COnflict
automaticaLLY
unUSual

Do we want to ignore those errors?

surekha1317:
Yes. I have these failures.
Ex: BLue, THis, rosetta(actual word is Rosetta in dictionary)

I am implementing this in java. I tried this but did not work.

for (Rule rule : langTool.getAllActiveRules()) {			
    if (!(rule instanceof SpellingCheckRule)) {
		langTool.disableRule(rule.getId());
    }
   if (rule instanceof SpellingCheckRule)
  {
	((SpellingCheckRule)rule).setConvertsCase(true);
	System.out.println(((SpellingCheckRule)rule).isConvertsCase());  //returns true
   }
 }

surekha1317:
I posted question here
 https://forum.languagetool.org/t/from-where-dictionary-is-being-picked-in-spell-checker/3894/4  
Somebody replied to change this file. So, I just tried but I do not have idea how this will work.
I need this solution in Java.

MikeUnwalla:
@surekha1317 , I am sorry that I cannot help you with your Java question because I do not have the technical skill.

surekha1317:
Ok. No problem. Thank you.

yakovru:
I think you may custom build LT from surekha1317:patch-1 branch. And add this custom artifact to maven dependency.

yakovru:
Could it be better to send text to LT after converting it to lower case?

surekha1317:
Words in dictionary will not be in lower case right? We have match exactly with dictionary then only it will pass. 
Ex: India (word from dictionary) . I have this in lower case as 'india'. LT is failing on this word because it is in lower case.

surekha1317:
If custom build LT from surekha1317:patch-1 is the only option then please give me details about how to do it.  I do not have any idea about this. This is very critical for me. I need to close this as early as possible. Please help me. 

If possible please add case insensitive feature for Java in your next release. Ideally, all properties you have mentioned in http://wiki.languagetool.org/hunspell-support should be overridden by end-users based on their requirements in java using apis. If possible please do these improvements for Java as well.
Thank you.

surekha1317:
Please anyone help me on this. I have been waiting for this solution. I do not know how to create custom LT with surekha1317:patch-1 branch. Please guide me in this.

yakovru:
Another worked way is apply following patch for languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/MorfologikAmericanSpellerRule.java  to you surekha1317:patch-1 branch or you local copy of repository of LT and build LT from this branch. 




# This patch file was generated by NetBeans IDE
# It uses platform neutral UTF-8 encoding and \n newlines.
--- (BASE)
+++  (LOCAL)
@@ -27,6 +27,9 @@
 import org.languagetool.Language;
 import org.languagetool.UserConfig;
 
+import org.languagetool.rules.spelling.morfologik.MorfologikMultiSpeller;
+import org.languagetool.rules.spelling.morfologik.MorfologikSpellerRule;
+
 public final class MorfologikAmericanSpellerRule extends AbstractEnglishSpellerRule {
 
   public static final String RULE_ID = "MORFOLOGIK_RULE_EN_US";
@@ -56,6 +59,18 @@
   }
 
   @Override
+  protected boolean isMisspelled(MorfologikMultiSpeller speller, String word) {
+  
+  
+      if (!speller.isMisspelled(word.toLowerCase(conversionLocale))) {
+     return false;
+      }
+
+
+    return super.isMisspelled(speller, word);
+  }
+  
+  @Override
   public String getLanguageVariantSpellingFileName() {
     return LANGUAGE_SPECIFIC_PLAIN_TEXT_DICT;
   }

surekha1317:
Thank you for your reply. I did not work on github code before. Please guide me to create jar from this branch.
I tried this, 
1. Downloaded surekha1317:patch-1 branch 
2. Imported project to eclipse as maven project
There are lot of sub modules in this project. Which one should I use? 
Previously, I was using `language-en`  jar.

surekha1317:
I could create jar and using custom LT in my project. I have added `fsa.dict.speller.convert-case=true` but did not work. 
Your code is working but not for all cases.
For ex: `sunday`, this will fail because actual dictionary word is `Sunday`. As per your api we are converting target word(`sunday`) to lower case. So, in this case it will fail.
I think both target word and dictionary word should be converted and compared then only this would be easy.
`  protected boolean isMisspelled(MorfologikMultiSpeller speller, String word) {
+  
+  
+      if (!speller.isMisspelled(word.toLowerCase(conversionLocale))) {
+     return false;
+      }
+
+
+    return super.isMisspelled(speller, word);
+  }
+`

Please check what bold content part returns for all cases.
Ex: BLue, friday (one is mixed case and another is not)

` && (!dictionaryMetadata.isIgnoringPunctuation() || isAlphabetic)
        && (!dictionaryMetadata.isIgnoringNumbers() || containsNoDigit(wordToCheck))
        && !(dictionaryMetadata.isIgnoringCamelCase() && isCamelCase(wordToCheck))
        && !(dictionaryMetadata.isIgnoringAllUppercase() && isAlphabetic && isAllUppercase(wordToCheck))
        && !isInDictionary(wordToCheck)
        && (!dictionaryMetadata.isConvertingCase() || 
            **!(!isMixedCase(wordToCheck) && 
                (isInDictionary(wordToCheck.toLowerCase(dictionaryMetadata.getLocale())) 
                    || isAllUppercase(wordToCheck) && isInDictionary(initialUppercase(wordToCheck)))**));`

igorgurman:
GURme🇮🇱

yakovru:
en_US.dict with all lowercase words. /home/desk/LT-4.2/languagetool-standalone/target/LanguageTool-4.5-SNAPSHOT/LanguageTool-4.5-SNAPSHOT/org/languagetool/resource/en/hunspell/en_US.dict

surekha1317:
Thank you so much for your help. You saved my work.
How to know if there are any future updates in Languagetool..?

surekha1317:
@yakovru Did you remove contraction terms(we've, aren't) from dictionary? It was not failing with old dictionary. Now, they are failing. Could you check once?

yakovru:
[en_US.dict.zip](https://github.com/languagetool-org/languagetool/files/2916607/en_US.dict.zip)
Updated dictionary.

surekha1317:
@yakovru  Thank you.

yakovru:
Dictionary is in fsa format with frequency data. It can not decoded now, because it included frequency data (bug). Dictionary without frequency data is extracted to word list successfully.
For build dictionary we use build scripts, that remove unnecessary word from word list, but this words (can't, aren't, etc.) are correctly detected with speller.

I use small script to convert to lover case lower.awk
> BEGIN {FS="\t"}
> {print tolower($0)}

and buld script like https://github.com/languagetool-org/languagetool/blob/master/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/hunspell/make_en_gb_dict.sh

>#!/bin/sh
># Get dictionaries from https://github.com/marcoagpinto/aoo-mozilla-en-dict
># Get en_us_wordlist.xml from https://github.com/mozilla-b2g/gaia/raw/master/apps/keyboard/js/imes/latin/dictionaries/en_gb_wordlist.xml
>
>./unmunch en_US.dic en_US.aff  >  en_US1.txt
>cat en_US1.txt | gawk -f lower.awk  |java -cp languagetool.jar:languagetool-dev-4.5-SNAPSHOT.jar org.languagetool.dev.archive.WordTokenizer en | sort -u > en_US.txt
>java -cp languagetool.jar org.languagetool.tools.SpellDictionaryBuilder -i en_US.txt -info en_US.info -freq en_us_wordlist.xml  -o en_US.dict

danielnaber:
Closing this PR, as it doesn't seem to be meant for general LT. Correct me if I'm wrong.

PR #1476, opened 2019-03-18
Ko

Grisworld:
a


danielnaber:
I suggest you run test using `testrules.sh` or `testrules.bar` first before submitting pull requests. Looking at the XML changes, I don't think they will work the way they are now.

PR #1557, opened 2019-05-01
[it] add some words

tiff:
Add some common names to Italian dictionary. @danielnaber I think this is safe to merge.


paolob67:
Yes no prob

Paolo

PR #1558, opened 2019-05-01
[fr] add words

tiff:
I found many popular names and abbreviations that are not supported by our dictionary. @danielnaber


danielnaber:
The last line with `Ø` is there o purpose?

tiff:
yes, popular spelling mistake in the addon...
Bildschirmfoto 2019-05-01 um 21 19 11

tiff:
@danielnaber I removed the special character

languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/hunspell/spelling.txt
```diff
@@ -1,23 +1,95 @@
...
+Craigslist
+CRM
+Desjardins
+DHL
+Dussaud
+Dédé
+eBay
+EHPAD
+Elisabeth
+Emilie
```
dpelle: In my opinion, many of the entries in this file should be deleted:
* some add obscure words (e.g. stp?, svt?…) which I would rather be shown as typos, otherwise real typos will be missed.
* some are even wrong such as the above one (Emilie in French should be Émilie).
* some are too informal  e.g. mdr which I assume is "mort de rire" (more less like "lol" in English) which SMS style abbreviation which should rather be avoided, or "dédé" which can be an informal name for "André".
* most are just weird random choices e.g why put "Bouteflika" in this list? We're not going to put thousands of people's names.

tiff: Thanks for your feedback. Those words are common ignored words of the LT browser addon as well as words ignored by users of SpellBoy.com.

* "lol" is also in the English and German dictionary.
* "Bouteflika" is the name of the former president of Morocco (until April 2019). 
* stp = I assumed it's the short form of "S'il te plait"
* Why should `Dédé` be avoided? Referring to people by their short name is not necessarily an error. Many people even prefer being called by their short name (e.g. "Richard" → "Dick")

I would be happy to hear your full feedback for the entire list. Maybe you can even take care of maintaining this list in the future? We could provide you with access to the commonly ignored words by LT and SpellBoy users. Are you insterested?

dpelle: > "Bouteflika" is the name of the former president of Morocco (until April 2019).

I know, but my point was: why do we add this name in particular, as opposed to thousands of other names (which I'm not proposing). It's strange. I don't think we need to add such proper names. It's endless.

> stp = I assumed it's the short form of "S'il te plait". 
It's a bit uncommon.

> Why should Dédé be avoided? Referring to people by their short name is
> not necessarily an error. Many people even prefer being called by their short
> name (e.g. "Richard" → "Dick")

The English "Richard" → "Dick" is not the same as "André" → Dédé in my opinion. They are not used in the same contexts. The English Bob (for Robert), Steve (for Stephen), Dick (for Richard), Bill (for William) etc. are commonly used, even in formal ways. It's quite different from "Dédé" in French which is rather informal (more spoken than written). You can for example read "Bill Gates" in serious newspapers for "William H Gates", but I doubt someone would use "Dédé Citroën" for "André Citroën".

I just find it a tad dangerous to add all kinds of names which are not in the dictionary. This kinds of words are more suitable for a personal dictionary.

PR #1587, opened 2019-05-15
Fix issue 1586

serval2412:
See bt and comment attached to the bugtracker


serval2412:
Just to put it clearly, I don't use languagetool and don't know internals of it.
This patch hasn't been tested and may hide the real bug or may be completely wrong.
I just  wanted to report upstream the bug from LibreOffice bugtracker and proposed some fix considering the surrounding code.

PR #1594, opened 2019-05-27
[en] add words to english dict

tiff:


tiff:
@MikeUnwalla please have a look. Mostly names and abbreviations (IG = Instagram, POC = Proof of Concept, ...)

MikeUnwalla:
GmbH is not a proper noun. It can be part of a proper noun.

IG. Are you sure that this is a proper noun, not an informal abbreviation?

These are abbreviations, not proper nouns:
KPI (key performance indicator)
NDA (non-disclosure agreement)
POC (proof of concept)
SKU (stock-keeping unit)

Aside, if POC is NNP, then POCs should be NNPS, not NNS. Similarly for other NNP/NNS pairs

tiff:
@MikeUnwalla Thanks for your feedback. I'm still learning a lot about POS.

**GmbH**
What POS tag would "GmbH" have?

**IG**
Wikipedia refers to it as "also known as"
https://en.wikipedia.org/wiki/Instagram

**Abbreviations**
What POS tag would those abbreviations get?

MikeUnwalla:
GmbH should have a POS that indicates an abbreviation, because that is what it is: https://dictionary.cambridge.org/us/dictionary/english/gmbh. But, we cannot give it such a POS, because there isn't one (refer to tagset.txt). The closest we have is NN:U (singular non-count noun) or FW (foreign word).

The term 'LanguageTooler GmbH' is a proper noun. But, I am not sure that is it correct to give a POS to the word GmbH alone. Think of the phrase 'ne plus ultra' (https://en.oxforddictionaries.com/definition/ne_plus_ultra). The three words together are a noun. But, the word 'ne' alone is not a noun. Possibly, the best postag for GmbH is FW.

Wikipedia is not necessarily correct. I accept that some people use the word 'IG' as an informal term for 'Instagram'. The purpose of LanguageTool is to help people to write correctly. Thus, if we accept all words, terms, and phrases that sum peeps use, we'm can packs us bags an go 'ome now, init bruv?

For the abbreviations such as KPI, I think the best POS is NN, because the terms behave as singular count nouns.

What do the other maintainers think?

tiff:
@MikeUnwalla I removed GmbH and IG for now and fixed the POS tags for the abbreviations. I think we can merge this now, if you don't object.

PR #1595, opened 2019-05-27
[es] add words to spanish dict

tiff:
Added common words that are very often added to the user's personal dictionary.


tiff:
@danielnaber

danielnaber:
I've asked a translator about the non-trivial cases, waiting for this feedback.

danielnaber:
Pep confirmed these words are valid.

Issue #1633, opened 2019-06-17
Bad suggestion for "etant"


danielnaber:
Should be "étant" but isn't for some reason.

dpelle:
The ticket is rather short.  I assume we are talking about spelling suggestions and not a about grammar rule.  "etant" is not a valid word in French.  The spelling corrections proposed by LT are
not so good indeed.  Using the web demo, it proposes:
* tant
* état
* jetant
* étang
* Ethan

A spelling mistake with missing diacritics should not be penalized much. Certainly "étant" should be the top suggestion.

Actually, I don't know how to upgrade the French dictionary anymore.  It was using Hunspell earlier (which most certainly had batter suggestion but was slow), and now it uses a binary FSA dictionary.  It would be good to have a script that completely automates the creation of the French dictionary (just like any other binary dictionary) otherwise it's not well maintainable.  I see that the binary FSA spelling dictionary was introduced in this commit:
```
$ git log fr_FR.dictcommit 9ef84c6e38ac3c03d89ed1db53ed37213c4e1003
Author: Daniel Naber <naber@danielnaber.de>
Date:   Thu Dec 21 15:54:25 2017 +0100

    [fr] offer suggestions for spelling errors
```
There is a `fr_FR.README` text file but it's vague about how to re-create the dictionary.  A script would be more precise:

```
$ cat languagetool/languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/hunspell/fr_FR.README

This Morfologik dictionary (fr_FR.dict, fr_FR.info) is based on a word list exported by
https://github.com/Fanaen/Hunspell2WordList. The following script has then been used to generate the
binary Morfologik dictionary: https://gist.github.com/danielnaber/b194b1a53e5a76373763e206c0ccbf02

The license is that of the original data, namely MPL 2.0 (see ../README_lexique.txt) 

Also see https://github.com/languagetool-org/languagetool/issues/850
```

danielnaber:
Indeed, when I export the binary (Morfologik) dictionary with `java -cp languagetool.jar org.languagetool.tools.DictionaryExporter -i org/languagetool/resource/fr/hunspell/fr_FR.dict -info org/languagetool/resource/fr/hunspell/fr_FR.info -o /tmp/out.txt`,  "étant" is simple not in that list. That seems to be the problem. I'll spend some more time trying to debug/fix this.

danielnaber:
[...]

PR #1634, opened 2019-06-17
Updated french dictionnaries to dicollecte v6.4

OswaldLuc:
Upgrading french dictionnary from dicollecte lexique v6.1 to v6.4


danielnaber:
Thanks for the PR! Could you have a look at https://github.com/languagetool-org/languagetool/issues/1633 - unless I was testing it the wrong way, your PR has the same issue with "etant" not giving the best suggestion. I wonder whether using https://github.com/Fanaen/Hunspell2WordList isn't a simpler approach? Do the scripts in your PR solve something that Hunspell2WordList cannot solve?

danielnaber:
Closing as there was no further feedback.

PR #1642, opened 2019-06-19
made ResultCache expire time configurable for server

fabrichter:
Use `cacheTTLSeconds` in server configuration.

### DIFF
--- a/languagetool-server/src/main/java/org/languagetool/server/HTTPServerConfig.java
+++ b/languagetool-server/src/main/java/org/languagetool/server/HTTPServerConfig.java
@@ -76,6 +76,7 @@ public class HTTPServerConfig {
   protected int maxWorkQueueSize;
   protected File rulesConfigFile = null;
   protected int cacheSize = 0;
+  protected long cacheTTLSeconds = 300;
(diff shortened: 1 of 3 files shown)


languagetool-server/src/main/java/org/languagetool/server/HTTPServerConfig.java
```diff
@@ -237,6 +238,10 @@ private void parseConfigFile(File file, boolean loadLangModel, boolean loadWord2
         if (cacheSize < 0) {
           throw new IllegalArgumentException("Invalid value for cacheSize: " + cacheSize + ", use 0 to deactivate cache");
         }
+        if (props.containsKey("cacheTTLSeconds") && !props.containsKey("cacheSize")) {
+          System.err.println("Use of cacheTTLSeconds without also setting cacheSize has no effect.");
```
danielnaber: As warnings tend to get  ignored, I'd be fine to just throw an exception here.

fabrichter: done

PR #1666, opened 2019-06-27
[en] add words to dictionary

tiff:
Added names (companies, people, tools) and other terms that users often add to their personal dictionary.

@MikeUnwalla please have a look and let me know what you think

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/added.txt
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/added.txt
@@ -4504,8 +4504,6 @@ instrumentalized	instrumentalize	VBD
 instrumentalized	instrumentalize	VBN
 instrumentalizes	instrumentalize	VBZ
 instrumentalizing	instrumentalize	VBG
-integration	integration	NN
-integrations	integration	NNS
 intellectualisers	intellectualiser	NNS
 intelligise	intelligise	VB
 intelligise	intelligise	VBP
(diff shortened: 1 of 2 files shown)


MikeUnwalla:
Most are good, but I have these comments/suggestions:

auth is an informal abbreviation and slang:
https://www.merriam-webster.com/dictionary/auth
https://www.lexico.com/en/definition/auth
But, I'm not sure about the plural or the verbs. Do you have a reference?

med also in informal adjective: https://www.lexico.com/en/definition/med

FB. Is this a proper noun or is it an informal abbreviation of Facebook?

TCP should be NN:U https://www.lexico.com/en/definition/tcp

integration as a count noun is fine. integrations (NNS) is already in LT

Postags for these adjectives are already in LT
sellable
unfulfillable

If a proper noun (example, Stripe) is the same as a common noun, it's not necessary to add proper noun the spelling, because LT does not give a spelling warning.

tiff:
@MikeUnwalla 
Thanks for your feedback.

**auth**: https://en.wiktionary.org/wiki/auth 

**FB** is an informal but commonly used abbreviation for Facebook, so should it be NN?

I changed **TCP** to `NN:U`.

I removed **integration(s)**, **sellable** and **unfulfillable** from `added.txt` but left them in `spelling.txt` since they were marked as spelling errors (not in BrE but other variants).

I removed **Stripe** from both.

MikeUnwalla:
> FB is an informal but commonly used abbreviation for Facebook, so should it be NN?

I think that NN is better than NNP.

Issue #1645, opened 2019-06-20
[en] PLURAL_VERB_AFTER_THIS false alarm


tiff:
> I think this commit should be moved to open-source.

Maybe the word "commit" is not detected as a noun?

![Bildschirmfoto 2019-06-20 um 08 34 28](https://user-images.githubusercontent.com/37363/59826317-60057c00-9336-11e9-8298-1a62de9d743e.png)

MikeUnwalla:
Not a false alarm. In standard English, 'commit' is not a noun.

This problem of verbs being used as nouns occurs with other rules (example A_INFINITIVE).

In the pattern 'this + X + should', X is a noun. If we disambiguate it such that it IS a noun, then possibly, we will not find real errors.

A possible solution is to add a note to the message as I did yesterday in (https://github.com/languagetool-org/languagetool/commit/b4383e11e7aaa2ba10be0d02cd025da3840c49db): " Note: This error message can occur if you use a verb as a noun, and the word is not a noun in standard English."

Possibly, we should create grammar rules that find 'developer speak'.

Discostu36:
I don't think we should only allow standard English in the POS dictionary. Because this way, we mask other possible errors like "I've contributed two commit to the repository."

I think the best (and most complicated)  solution would be to add `commit/S` to added.txt and additionally write a rule that tells the user that "commit" is not used as a noun in standard English.

tiff:
@MikeUnwalla @Discostu36 Actually Wiktionary lists "commit" as a noun:
https://en.wiktionary.org/wiki/commit#Noun

MikeUnwalla:
@tiff, yes "In computer science and data management,"

In some specialist subjects, words are used in a non-standard way. For example, I have seen the verb 'compute' used as a noun or an adjective, as in, "These compute resources are..." (Parse as a noun in a noun cluster or as an adjective that modifies 'resources'.)

Before we make changes, we should answer these questions:

- How do we decide which terms (verbs used as nouns) to include and which to exclude?
- If we allow verbs used as nouns, then should we also allow common nouns and proper nouns that are used as verbs? (As in, "If you LanguageTool your document, you will probably find many grammar errors.") If no, why not? (Because we already allow "If you google, you will find the information.")
- What is the best way to deal with non-standard usage?
- How do we prevent false negatives for the users who use a term incorrectly (because they are not using it with its specialist meaning)?

@Discostu36 wrote:

[...]

PR #1740, opened 2019-07-23
[en] add names and technical terms to dictionary

tiff:
Many terms that users added to their personal dictionary. 

Contains a fix for https://github.com/languagetool-org/languagetool/issues/1645


danielnaber:
center: Should we have a rule then that explains centre/center? Same for programme? Just accepting the en-US variant for all other seems problematic.

tiff:
good point. I will remove both terms for now.

tiff:
In Canadian English we don't suggest "centre"
Bildschirmfoto 2019-07-23 um 11 30 07

danielnaber:
"configed" is colloquial for "configured" I guess? I think we should all `#coll` tags to colloquial words now, so that we can later more easily introduce profiles.

tiff:
yes, see https://en.wiktionary.org/wiki/config

tiff:
I have removed the verb form of config for now

PR #1767, opened 2019-07-30
[es] rule for am/pm abbreviation

tiff:
See https://forum.languagetool.org/t/es-usage-of-am-pm/4522

### DIFF
--- a/languagetool-language-modules/es/src/main/resources/org/languagetool/resource/es/hunspell/spelling.txt
+++ b/languagetool-language-modules/es/src/main/resources/org/languagetool/resource/es/hunspell/spelling.txt
@@ -1,6 +1,5 @@
 # Words that extend the spell checker. See ignore.txt for words that should be
 # completely ignored (i.e. not used to create suggestions).
-# AM
 # Center
 # City
 # derrepente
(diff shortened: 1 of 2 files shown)


fitojb:
I was wondering if this could be made so that the abbreviation is spaced with non-breaking spaces? Like so: _a.&nbsp;m., p.&nbsp;m._

tiff:
Hi @fitojb. Sure, I'm not familiar with Spanish grammar but if you could give some insights why there needs to be a space in between I would be happy to implement this change

fitojb:
We add a space because each abbreviated letter belongs to a different word (respectively, _ante meridiem_ or _post meridiem_). Further examples:
* _que en paz descanse → q. e. p. d._
* _visto bueno → v.º b.º_
* _nota del traductor → N. del T._
* _por ejemplo → p. ej._

The only abbreviations that do not use a space are those written with a slash (which are few): _sin número → s/n_ or _s. n._ We also do not use spaces in initialisms: _ONU_ (UN); _OTAN_ (NATO), etc.

tiff:
@fitojb done. See https://github.com/languagetool-org/languagetool/commit/3548c9a5c36dc425506eb0301b6cb0d639e2d661

Are you interested in contributing rules to the Spanish version of LT? We are in need of someone maintaining and growing the rule set.

PR #1816, opened 2019-08-12
[en] NNP for lowercased 'google'

tiff:
Hi Mike, I noticed that many users were using the noun "google" in lowercase:
> Have you checked the **google**  search?
> It's a **google** plugin, that checks your spelling.

This triggers a false alarm:
Bildschirmfoto 2019-08-12 um 08 43 40


MikeUnwalla:
@tiff, I disagree with the change. Lower-case 'google' as a proper noun is not correct. Or, if it is correct, then lower-case 'languagetool' is also correct.

tiff:
I understand. How can we suggest "Google" for "google" then to prevent the misleading false alarm?

MikeUnwalla:
Put an antipattern for 'google'  in rule A_INFINITIVE and make a POS-aware rule for lower-case 'google' when it is used as a noun or an adjective.

Issue #1694, opened 2019-07-09
[en] "several other"


tiff:
> He and several other.

In this case "other" should be "others".

Took this from the comments of this blog post about LanguageTool: 
https://www.ghacks.net/2019/07/08/languagetool-is-a-free-open-source-grammar-and-spell-checker/

@MikeUnwalla is this maybe a simple one for our open-source version?

tiff:
@MikeUnwalla I'm reopening this. There's no error found in:
> Tom and several other did a great job.

PR #1819, opened 2019-08-13
[en] improve SEVERAL_OTHER

tiff:
For https://github.com/languagetool-org/languagetool/issues/1694

@MikeUnwalla please let me know if this is good.


MikeUnwalla:
@tiff, You can optimize the 'or' and use `<token postag_regexp="yes" postag="SENT_END|VBD?"/>`

But, adding the VB/VBD postags causes many false positives for noun clusters. Use https://community.languagetool.org/ruleEditor/expert?devMode=true&factor=10 to see what I mean.

tiff:
@MikeUnwalla thanks for the feedback. See my recent changes. No more false positives are created.

MikeUnwalla:
@tiff , I don't understand this Git message:

```
Pull request successfully merged and closed
You’re all set—the fix-1694 branch can be safely deleted.
```
I will let you deal with it!

tiff:
It means you can safely delete a git branch after it got merged. I just did that. Thanks for looking into it.

Issue #1468, opened 2019-03-14
[en] false alarm "follow up"


tiff:
> Here's the follow up task.

![Bildschirmfoto 2019-03-14 um 12 23 30](https://user-images.githubusercontent.com/37363/54353330-08c77f80-4654-11e9-9972-db206892debe.png)

Similar to https://github.com/languagetool-org/languagetool/issues/1371

MikeUnwalla:
Why is this a false alarm? Suggestion 2 in the message explains the problem.

tiff:
ah sorry, Mike. I was expecting the same behavior as here:

![Bildschirmfoto 2019-03-14 um 13 49 56](https://user-images.githubusercontent.com/37363/54358221-171b9880-4660-11e9-8a00-1099dc0c4cbf.png)

If the behavior is intended, feel free to close this issue.

MikeUnwalla:
@tiff , it would be nice if the message were the same as the second example. (I think the message is good enough, but I'll leave the issue open because someone may disagree with me.)

Issue #1812, opened 2019-08-11
[en] Improve A_INFINITIVE


tiff:
One of the most deactivated rules from our English users is `A_INFINITIVE`. I think this happens because we don't provide a replacement suggestion.

Some example sentences that make users turn off this rule are:

> Please click the cancel button.

* [x] "Cancel" is not a noun. We should either suggest "Cancellation", suggest "cancel" in quotes or add an antipattern for "button"

> You need to click the unsubscribe link.

* [x] "Unsubscribe" is not a noun. We should either suggest "Cancellation", suggest "unsubscribe" in quotes or add an antipattern for "link"

> Press the delete key on your keyboard.

* [x] In this case we probably should either suggest "deletion" or also add an antipattern for "key".

> I'm waiting for the redeliver.

* [x] Not a noun. We should suggest "redelivery".

> The install failed. 

* [x] "Install" is not a noun. We should suggest "Installation" (same for "uninstall")

> Please give me a renew.

* [x] "renewal" should be suggested.

> I need a recharge.

* [x] "recharge" is not a noun. Maybe this should be one. Seems quite common. I don't know which noun to suggest here.

> I got a write up at work.

* [x] I think we should suggest "write-up".

> I will sent him a thank you letter.

* [x] I think we should suggest "thank-you".

> Thanks for the follow.

* [x] Quite common. "follow" might actually be a noun. (see https://www.merriam-webster.com/dictionary/follow)

What do you think @danielnaber @MikeUnwalla ?

PR #1821, opened 2019-08-13
[en] partially address #1812

tiff:
What's in this PR:
1. Add NN as POS for "follow" (see https://en.wiktionary.org/wiki/follow and https://www.merriam-webster.com/dictionary/follow) "a follow up" is now caught by `CA_FOLLOW_UP` which also offers a suggestion. (this would close https://github.com/languagetool-org/languagetool/issues/1468)
2. Add NN as POS for "recharge" (see https://en.wiktionary.org/wiki/recharge and https://en.wiktionary.org/wiki/recharge)
3. Improve A_INFINITIVE by adding antipatterns
4. In those antipatterns are now "thank" and "google" for which I created special rules
5. Added "merge" as an antipattern to `A_INFINITIVE` to prevent some false alarms in common tech speak: "Can you have a look at the merge request?"

@MikeUnwalla let me know what you think. Thanks!

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/added.txt
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/added.txt
@@ -3558,6 +3558,8 @@ focalizes	focalize	VBZ
 focalizing	focalize	VBG
 FOCs	FOC	NN
 fo'c'sle	fo'c'sle	NN
+follow	follow	NN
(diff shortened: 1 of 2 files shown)


MikeUnwalla:
@tiff , @danielnaber , the term 'merge request' is a 2-word noun. We don't have an easy way to deal with such terms. It would be nice if we could deal with them in the disambiguation (I'm not sure how).

PR #1823, opened 2019-08-14
[en] add POS for cancel

tiff:
Hey @MikeUnwalla,

I saw many false positives caused by "a/the cancel". 

Webster and Oxford lists it as a noun:
https://www.lexico.com/en/definition/cancel
https://www.merriam-webster.com/dictionary/cancel

Apparently there's no plural form of "cancel". Do I need to add the uncountable POS here?
What is the difference between `NN:U` and `NN:UN`?

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/added.txt
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/added.txt
@@ -952,6 +952,8 @@ Canadianized	Canadianize	VBD
 Canadianized	Canadianize	VBN
 Canadianizes	Canadianize	VBZ
 Canadianizing	Canadianize	VBG
+cancel	cancel	NN
+cancels	cancel	NNS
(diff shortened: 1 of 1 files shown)


tiff:
Update: Wiktionary lists "cancels" as plural form. https://en.wiktionary.org/wiki/cancel#Noun

MikeUnwalla:
I cannot think of a reason for the plural noun 'cancels' not to exist. The terms are used in a countable way. Refer to the lexico page. If we can have one cancel, we can have "Two stamps franked and with adhesive cancels."

If you want to research how the terms are used, use https://www.english-corpora.org/.

NN:U means that the word is used only as a non-count noun. So, the singular noun 'cancel' must have the postag NN, because it is a countable noun.
NN:UN means that the word can be both countable and non-countable.

tiff:
Thanks @MikeUnwalla. I have added NNS for cancels.

PR #1909, opened 2019-09-04
+ added: CORS preflight OPTIONS request handling for /v2/check

### LINKED ISSUE #1908: CORS preflight OPTIONS request answered 400
It seems like the automatic preflight request that's issued by the browser when firing a CORS request is answered with a status code 400 instead of 200.

Also the server complains:
> 2019-09-04 14:40:10 +0000 An error has occurred: 'Missing 'text' or 'data' parameter', sending HTTP code 400.

Glancing at the code it seems like there's actually no handling for preflight requests yet. So when I issue a request against `/v2/check` the browser sends an OPTIONS request to `/v2/check` but languagetools behaves as if it was a POST request and for there are no parameters (`text`) it answers the preflight request with `400`.

The check handler should check if the http method is OPTIONS and answer accordingly. The same is probably true for any other endpoints. So it might be useful to handle OPTIONS requests further up the chain.

black-snow:
Maybe we should refactor `ServerTools.setAllowOrigin()` into `ServerTools.setAccessControlAllowOrigin()` for consistency.
Also, I only added the OPTIONS check to `/check` - nothing else.

Fixes #1908.

### DIFF
--- a/languagetool-server/src/main/java/org/languagetool/server/ApiV2.java
+++ b/languagetool-server/src/main/java/org/languagetool/server/ApiV2.java
@@ -62,6 +62,11 @@ class ApiV2 {
     if (path.equals("languages")) {
       handleLanguagesRequest(httpExchange);
     } else if (path.equals("check")) {
+      // check for CORS preflight options request
+      if (httpExchange.getRequestMethod().equalsIgnoreCase("options")) {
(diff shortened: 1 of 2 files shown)


black-snow:
The new imports in `ServerTools` are unused and could / should be removed.

PR #1943, opened 2019-09-16
[en] add antipatterns to CHILDISH_LANGUAGE

tiff:
@TiagoSantos81 @danielnaber 
I added some basic antipatterns to CHILDISH_LANGUAGE. Unfortunately they don't work because antipatterns don't work with regexps. What's the alternative to solve this?


danielnaber:
Only short-term and simple alternative is to change the rule to use `<token regexp="yes">` instead of `<regexp>`. If that doesn't work, we could use `<filter>` where Java code removed the false alarms.

tiff:
Closed via https://github.com/languagetool-org/languagetool/commit/7aaf508d9b56e090a0839c767b3a131f77ed7cc6

PR #1981, opened 2019-09-30
[zh] replace ictclas4j with HanLP which supports more Chinese characters. Add more rules.

passerbya:
Ictclas4j  using GB2312 character set which only supports 6763 Chinese characters.


danielnaber:
Thanks! Could you check the test results at https://travis-ci.org/languagetool-org/languagetool/builds/591552328?utm_source=github_status&utm_medium=notification? Tests fail after your change it seems.

passerbya:
OK, I'll do it.

danielnaber:
Thanks, I have merged your PR. Support for Chinese isn't maintained in LT, so more contributions would be very welcome.

PR #1987, opened 2019-10-01
[en] rule to detect confusion of 'heave' and 'heavy'

tiff:
Just made this mistake myself. Seems like a common confusion.


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -9731,6 +9731,35 @@ USA
...
         </rulegroup>
+        <rulegroup id="HEAVE_USE_OF" name="make heave (heavy) use of" default="off">
+            <rule>
+                <pattern>
+                    <token inflected="yes">make</token>
+                    <token min="0" />
+                    <marker>
+                        <token>heave</token>
+                    </marker>
+                    <token>use</token>
```
danielnaber: Maybe also `usage`?

PR #1988, opened 2019-10-01
[en] add some POS and words (common polish names)

tiff:


languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/added.txt
```diff
@@ -10889,6 +10903,12 @@ underpraised	underpraise	VBD
 underpraised	underpraise	VBN
 underpraises	underpraise	VBZ
 underpraising	underpraise	VBG
+underreport	underreport	VB
```
tiff: https://www.merriam-webster.com/dictionary/underreport

languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/hunspell/spelling.txt
```diff
@@ -3371,4 +3371,32 @@ GDPR
...
-GmbH
\ No newline at end of file
+GmbH
+underreport
+underreported
+underreports
+underreporting
+Salzburg
+Marcin
+Michal
```
danielnaber: Very close to "Michael" and probably much less common, I'd says leave it out because of that.

tiff: I just noticed that it's "Michał" (notice the special 'ł'). I changed it to that.

PR #2128, opened 2019-11-06
Log via slf4j instead of stdout / stderr

black-snow:
In the long run we should replace the calls (from outside ServerTools) with calls to the classes' own loggers (in order to get the correct class name in the output). This is rather a quick fix when you don't want LT to *"pollute"* your stdout/stderr. There might also be other places where `System.out` is used.

Also I had to drop `languagetool-server/src/main/resources/logback.xml` and rebuild in order to provide my own config via `-Dlogback.configurationFile`. But this might be a change with side-effects for some users so I won't propose this for now.

### DIFF
--- a/languagetool-server/src/main/java/org/languagetool/server/ServerTools.java
+++ b/languagetool-server/src/main/java/org/languagetool/server/ServerTools.java
@@ -22,11 +22,12 @@ import com.sun.net.httpserver.HttpExchange;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.languagetool.JLanguageTool;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


danielnaber:
I just tried this - it seems now the default is that you don't see any output by default? So you have to provide your own logback config? I think by default, we should always print visible logging.

black-snow:
Weird, I though the `logback.xml` from `languagetool-server` would be used. I guess we should then copy it into the generated bundle so that it'll be used when you start the server from there. What do you think?

danielnaber:
> copy it into the generated bundle

I think it should be there even for developers in their IDE, so that the IDE setup doesn't differ from the end user setup?

black-snow:
Not quite sure if we'd then need a `logback.xml` in every module or just one top-level. Can you give this a quick shot - I don't have this set up *atm*.

PR #2147, opened 2019-11-13
[zh] updated hanlp, added more rules

passerbya:


danielnaber:
Thanks for your contribution!

PR #2158, opened 2019-11-16
PatternTokenBuilder: minor fix. 

bebras123:
It seems that `.negate()` works only for POS tokens (e.g. `new PatternTokenBuilder().posRegex("NN.*?").negate().build()`). This fix should enable negation for "text" tokens (e.g. `new PatternTokenBuilder().tokenRegex("an?|the").negate().build()`).


danielnaber:
Thanks! Could you maybe also add a test case?

bebras123:
Added test cases.

Issue #2001, opened 2019-10-04
quote within spaces


baarsrj:
It is not correct when a single or double quote has spaces on both sides. The Java rules check a lot about these special characters, but not this.
It is easy enough to make in XML, but since it is probably true for all languages, should it not be added to the existing Java routines for checking " ' and ( ) ?

s0nata:
just made a pull request (#2159) with a possible fix, it addresses the issues of whitespace around quotes, but the () brackets were already being treated well.

s0nata:
just updated the pull request with testing code

danielnaber:
The pull request has been merged, thanks! Can this issue be closed now?

s0nata:
@baarsrj @danielnaber while working on this pull request I have noticed the following things:
* many classes in  `languagetool-language-models` use the deprecated constructor `CommaWhitespaceRule(messages)`;
* it seems that for some languages (I only noticed it for German, maybe there are more) whitespace around quotes is being treated not exclusively with the java code, but maybe with XML rules
* I am in favour of moving the treatment of whitespace around quotes to a separate rule, current one is called `CommaWhitespaceRule`, it's ID is `COMMA_PARENTHESIS_WHITESPACE` and it also treats quotes, while this is not indicated anywhere.

I would like to work on this. Should I open an issue, or just submit a new pull request?

danielnaber:
Feel free to open an issue.

PR #2159, opened 2019-11-17
fix for the issue #2001: quote within spaces

s0nata:
a simple fix suggestion that checks that no quote is surrounded by whitespaces on both sides. Very minor class-local refactoring. All tests pass on build.


danielnaber:
Thanks! Could you add tests to CommaWhitespaceRuleTest that check the new behavior?

s0nata:
@danielnaber done!

PR #2287, opened 2020-01-04
[ga] specific case rule

jimregan:


languagetool-language-modules/ga/src/main/java/org/languagetool/rules/ga/SpecificCaseRule.java
```diff
@@ -0,0 +1,118 @@
...
+    return "GA_SPECIFIC_CASE";
+  }
+
+  @Override
+  public String getDescription() {
+    return "Checks upper/lower case spelling of some proper nouns";
+  }
+
+  @Override
+  public RuleMatch[] match(AnalyzedSentence sentence) {
```
danielnaber: Is this a 1:1 copy? In that case, we should introduce a class in core that all languages can derive from, to avoid duplication.

jimregan: Not exactly, because the case checking for Irish is weird. For languages with normal casing, I think you could go with a class in core.

Issue #2341, opened 2020-01-17
Command-line tool: problem with match column number on first line


matze-dd:
The command-line version languagetool-commandline.jar sometimes produces too small column numbers, if the problem arises on the first text line. This only happens at the first line, and it is not restricted to language de-DE.

Matthias

EDIT: JSON output is correct, version is 4.8.

In this example, column number should be 1:
```
$ cat t.txt 
das ist ein Test.
$
$ java -jar languagetool-commandline.jar --language de-DE t.txt 
Expected text language: German (Germany)
Working on t.txt...
1.) Line 1, column 0, Rule ID: UPPERCASE_SENTENCE_START
Message: Dieser Satz fängt nicht mit einem großgeschriebenen Wort an
Suggestion: Das
das ist ein Test. 
^^^               
Time: 792ms for 1 sentences (1.3 sentences/sec)
```

matze-dd:
See pull request #2342

PR #2342, opened 2020-01-17
PR for Issue #2341

matze-dd:
Related to issue #2341:
- fix for `JLanguageTool.java`:
  match column number now is 1-based for first input text line, too
- fix for `GenericUnpairedBracketsRuleTest.java`:
  `mvn clean test` does run successfully

Matthias

EDIT: revised fix for `JLanguageTool.java`

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
+++ b/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
@@ -1274,12 +1274,13 @@ public class JLanguageTool {
             newMatch.setOffsetPosition(newFromPos, newToPos, newMatch);
             newMatch.setLine(range.from.line);
             newMatch.setEndLine(range.to.line);
-            if (match.getLine() == 0) {
+            if (newMatch.getLine() == 0) {
               newMatch.setColumn(range.from.column + 1);
+              newMatch.setEndColumn(range.to.column + 1);
             } else {
               newMatch.setColumn(range.from.column);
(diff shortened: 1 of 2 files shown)


danielnaber:
The doc says "Get the column number in which the match occurs (zero-based).", so this is a semantic change, or am I missing something?

matze-dd:
Languagetool-commandline.jar displays both line and column of a match as numbers counting from one. Only if the match refers to the first input text line, then for some rules the column number is zero-based. For a human reader this will be no problem. But it may disturb if one tries to process the output for some reason.

Consider [CommandLineTools.java](https://github.com/languagetool-org/languagetool/blob/master/languagetool-commandline/src/main/java/org/languagetool/commandline/CommandLineTools.java), method `printMatches()`. Here one sees on line 174 that `match.getColumn()` is supposed to return a one-based number, compare `match.getLine()` on the previous line.

Matthias

matze-dd:
Here some more explanation.

As already commented in the post to [LanguageTool forum](https://forum.languagetool.org/t/sometimes-wrong-column-number-from-command-line-tool/5286), a different treatment of the first input text line is foreseen in [JLanguageTool.java](https://github.com/languagetool-org/languagetool/blob/master/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java), line 1277.
```
            if (match.getLine() == 0) {
```
But the test fails, since the used object `match` does not have an initialised line number; it is always -1. The proposed fix uses `newMatch` instead. Its line number has been set just before.

Matthias

Issue #1783, opened 2019-08-01
[pt] "email" requires three corrections until correct


tiff:
> Meu email

![Aug-01-2019 15-33-55](https://user-images.githubusercontent.com/37363/62297608-da670700-b471-11e9-817b-553270b0833a.gif)

TiagoSantos81:
@tiff 
1st is the preferred spelling for email.
2nd is a barbarism rule, which recommends the translated version (some people may ignore this style recommendation). This database was built in European Portuguese and extended to all branches due to its usefulness.
3rd is the Brazilian preferred spelling for 'eletrónico' rule. This step could be omitted if the replacement table is language specific and a new replacement rule is created for Brazilian Portuguese, where entries like this would be translated, or, to avoid duplication, the replacement rule is improved to permit specifying variant specific replacements. 
For now, the several trigger workaround works fine, particularly in the offline tools.

tiff:
Thanks for clarification

Issue #2596, opened 2020-03-23
[pt] user feedback


tiff:
> Correção errada. Em português brasileiro é eletrônico e não eletrónico.
> Em substituição ao termo e-mail for sugerido "correio eletrónico" ao invés de "correio eletrônico".

Not sure if that's correct.

tiff:
Wikipedia lists "correio eletrônico" as valid for pt-BR

marcoagpinto:
@tiff 

I will have a look at it after the LT 4.9 release.

Right now I don't want to make regressions in the eve of the release.

marcoagpinto:
Hello @tiff 

It works correctly:
![correio_eletronico_20200325](https://user-images.githubusercontent.com/5192600/77523067-175a9080-6e7d-11ea-8ba6-feaf5cc3832e.png)

Maybe the user forgot to change the speller to pt_BR

tiff:
@marcoagpinto no, I think you misunderstood the issue. See the suggestion for "e-mail":
![Bildschirmfoto 2020-03-25 um 10 46 57](https://user-images.githubusercontent.com/37363/77523431-06fae380-6e86-11ea-976c-03d6621baa2b.png)

tiff:
The suggested replacement is not Brazilian Portuguese.

marcoagpinto:
Ahhhhh... I see... I will have to create a rule in grammar.xml PT and grammar.xml BR, and remove the word from barbarisms.txt.

I will do it today or tomorrow.

marcoagpinto:
@tiff 

Quick fix:
https://github.com/languagetool-org/languagetool/commit/5ddc2d8d6b50defbd9160c70edca635497a7d4fa

Can it be closed now?

tiff:
@marcoagpinto I would prefer a cleaner solution. Also, I think the term "(Brasil)" will be inserted into the text when applying the fix, which will be very annoying

marcoagpinto:
@tiff 
Yes, I will fix it later then.

:-)

tiff:
Ok, cool. Just to give you a feeling: We have 5 times more users from Brazil than we have from Portugal ;)

marcoagpinto:
@tiff 

I have decided that the easiest way is just to suggest writing in italic.

So, I have added to grammar.xml:

`<!ENTITY barbarismos3 "e-mails?|mails?|tablets?">`

But, it doesn't recognize the hyphen and "e" before.

[...]

Issue #2040, opened 2019-10-17
[pt] words missing in dictionary?


tiff:
@marcoagpinto some more commonly ignored words:
* iFood
* Outlook
* Nathalia
* Chromecast

Also:
* Whatsapp
* whatsapp (all lowercase)
* instagram (all lowercase)
* facebook (all lowercase)

I think we need rules for them because their correct spelling isn't suggested at least not in pt-br.

marcoagpinto:
Well, the proper way to write the last four is in uppercase.

I will add in uppercase.

tiff:
I think they are already in the dict in uppercase. The problem is rather that the correct word isn't suggested:
<img width="365" alt="Bildschirmfoto 2019-10-17 um 12 50 06" src="https://user-images.githubusercontent.com/37363/67002656-ab5d4a80-f0dc-11e9-9c71-1966a7b492fd.png">

I would suggest creating simple XML rules for those 4 terms. (also Whatsapp should be WhatsApp)

PR #2617, opened 2020-03-25
PT improvements

tiff:
@marcoagpinto I created the rules discussed in 
https://github.com/languagetool-org/languagetool/issues/2040
https://github.com/languagetool-org/languagetool/issues/2596

Can you have a look and merge this, if ok? Thanks!


tiff:
It also partially fixes: https://github.com/languagetool-org/languagetool/issues/1783

PR #2690, opened 2020-04-03
Grazie migration: Add ClassBroker and fix resource requests not from ResourceDataBroker

Fatalll:
ClassBroker Is responsible for loading the necessary classes for LanguageTool library. It is necessary to provide the ability to load classes from custom classloaders.


danielnaber:
@Fatalll I had to undo this: `URL res = getDataBroker().getAsURL(JLanguageTool.class.getSimpleName() + ".class");` in `org.languagetool.JLanguageTool#getBuildDate`, as it didn't work, i.e. the build date was always `null`. I haven't spent time to see if there's a better fix than just rolling it back.

Fatalll:
Sorry, my fault, pushed fixes: [PR](https://github.com/languagetool-org/languagetool/pull/2756)

PR #2691, opened 2020-04-03
Grazie migration: Lowercase words if necessary after adding determiner for English language

Fatalll:
If it at the start of the sentence you need to lowercase word after determiner.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/tools/StringTools.java
+++ b/languagetool-core/src/main/java/org/languagetool/tools/StringTools.java
@@ -18,20 +18,15 @@
  */
 package org.languagetool.tools;
 
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.Reader;
-import java.nio.charset.StandardCharsets;
-import java.util.*;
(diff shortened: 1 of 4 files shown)


languagetool-core/src/main/java/org/languagetool/tools/StringTools.java
```diff
@@ -232,6 +227,15 @@ public static String lowercaseFirstChar(String str) {
     return changeFirstCharCase(str, false);
   }
 
+  /**
+   * Return <code>str</code> if str is not abbreviation, otherwise return
+   * modified <code>str</code> so that its first character is now a lowercase character.
+   */
+  public static String lowercaseFirstCharIfNotAbbreviation(String str) {
```
danielnaber, line 234: I'm not sure yet. Could you make the Javadoc describe how "abbreviation" is defined? It looks to me as if this is actually a test for an acronym, i.e. an all-uppercase abbreviation like NATO, NASA etc?

Fatalll: Thanks for the notice. We've reworked idea and now will lowercase word only if it is capitalized.

PR #2692, opened 2020-04-03
Grazie migration: Add pattern position to RuleMatch

Fatalll:
Add possibility to know rule pattern range.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
+++ b/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
@@ -986,6 +986,11 @@ public class JLanguageTool {
     }
     RuleMatch thisMatch = new RuleMatch(match);
     thisMatch.setOffsetPosition(fromPos, toPos);
+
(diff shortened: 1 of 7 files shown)


danielnaber:
Thanks, I understand this is the pattern position, i.e. unlike the usual position, which is just the position of the `<marker>`, isn't it? Could you describe the use case for it?

Fatalll:
Well, we have 2 use cases:
1) We use rule pattern text as user exception mechanism because marker is too narrow for this.
ezgif-4-3f3551a22608
2) We use it to detect borders of unrelated parts of text. For exampe string literal with template:
`var result = "status is ${response.status}, data key is ${response.data.key}"`
Template will be deleted and text concatenated, and we can get false positives due to it.

danielnaber:
I've merged the PR, thanks. For the second use case, have you considered using https://languagetool.org/development/api/org/languagetool/markup/AnnotatedText.html?

Fatalll:
Oh, thanks for the tip! Not sure that it will solve all our problems in second case, but I will take a look :)

PR #2693, opened 2020-04-03
Grazie migration: Add category and messages to DashRule

Fatalll:
Currently all DashRules don't have a correct category.


danielnaber:
Thanks, it has been merged.

PR #2696, opened 2020-04-03
Grazie migration: Add intern tags for taggers and EnglishTagger

Fatalll:
Add possibility to intern string tags for Taggers. Intern tags for English because it have a megabytes of duplicated strings.


danielnaber:
Thanks. Are there side effects of this? Or in other words, should this be an option at all, can't it be always done?

Fatalll:
It depends. The cache is maintained by JVM in permanent memory pool which is usually limited in size compared to heap. So if you have like <1000 tags in language its ok to intern it, but if it's thousands and thousands, it is probably not a good idea.

PR #2698, opened 2020-04-03
Grazie migration: Add exceptions for EnglishWordRepeatBeginningRule

Fatalll:
If several sentences start with determiner - it is a correct case.


danielnaber:
This is a style rule, so it's a bit a matter of taste, but I tend to agree with your change.

PR #2710, opened 2020-04-06
Add IT terms dictionary

Fatalll:
Top by popularity IT-related terms from StackOverflow


danielnaber:
Thanks for the PR. I have added a few of these terms to our dictionary, but I think most of them are too specific to IT.

PR #2777, opened 2020-04-20
Added new rules and antipatterns

vkyfox:


danielnaber:
Please add `default="temp_off"` to be sure there are no false alarms

PR #2812, opened 2020-04-27
Grazie migration: Added added_custom.txt and removed_custom.txt for taggers

Fatalll:
Added the ability to expand the tagger dictionary in forks without having to resolve merge conflicts (similar to spellchecker). Also moved default names and duplicated code to the BaseTagger.


languagetool-core/src/main/java/org/languagetool/tagging/BaseTagger.java
```diff
@@ -114,24 +116,51 @@ protected WordTagger getWordTagger() {
...
-      if (manualRemovalFileName != null) {
-        try (InputStream stream = JLanguageTool.getDataBroker().getFromResourceDirAsStream(manualRemovalFileName)) {
+      InputStream stream = null;
+      try {
+        for (String file : getManualRemovalsFileNames()) {
+          for (URL url : JLanguageTool.getDataBroker().getFromResourceDirAsUrls(file)) {
+            if (stream == null) {
+              stream = url.openStream();
+            } else {
+              stream = new SequenceInputStream(stream, url.openStream());
```
danielnaber, line 127: Does this really work for more than two items in the `for` loop?

Fatalll: Why this shouldn't work? [Here](https://stackoverflow.com/questions/14295099/how-to-chain-multiple-different-inputstreams-into-one-inputstream) is a good discription for such solution.

danielnaber: Ah, okay!

PR #2835, opened 2020-04-29
Fix for example of `Readability: Three nouns in a row` rule in English

TanVD:
Minor fix of spelling mistake in `Readability: Three nouns in a row` rule


danielnaber:
Thanks!

PR #2850, opened 2020-05-02
Add a rule to the group QUESTION_MARK for common MD+PRP

dataindataout:
Add a rule for sentences beginning with common English modal verbs followed by common personal pronouns for the rule group QUESTION_MARK.

I avoided using the generic modal (MD) part of speech (postag), because there are several modal verbs that cause false matches. For example, "May we all do well." is not a question.

Likewise, I avoided using the generic PRP postag, because there are several pronouns that cause false matches in the pattern. For example, "Do them a favor." is not a question.

The original piece of text I wanted to be marked as incorrect in an email conversation was, "Can you help me."

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
@@ -42303,6 +42303,25 @@ The accident victim died from her injuries.
                 <example>Who does not love wine, women and song; remains a fool his whole life long.</example>
                 <example>"What did the professor talk about?" the student asked.</example>
             </rule>
+            <rule>
+                <pattern>
+                    <token postag="SENT_START"/>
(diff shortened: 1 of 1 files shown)


danielnaber:
Thanks! `Don't you worry.` might also be a false alarm? You can get a more extensive check when using this link: https://community.languagetool.org/ruleEditor/expert?devMode=true

Also, `type="correct"` is optional.

dataindataout:
Thank you! I will use that devMode link for the next one, very helpful output.

danielnaber:
Our nightly test has found some more false alarms: https://internal1.languagetool.org/regression-tests//20200502/result_en_20200502_table.html. I have added `default="temp_off"` to the `<rule>` so it's active in the nightly tests, but not in the production system (we auto-deploy changes once a day to the production system at languagetool.org). Could you have a look at the false alarms?

dataindataout:
Yes, taking a look!

dataindataout:
I have the antipattern for an inversion of the conditional clause working locally. 

I need to work next on an antipattern for the imperative case.

PR #2873, opened 2020-05-07
Improve newest QUESTION_MARK rule to cover more test cases

dataindataout:
This covers the majority of the test cases that didn't pass the first time (especially the large chunk of sentences with inverted conditional clauses), and there were only a handful of unavoidable false negatives in the language tool checker.
Do you in general prefer false negatives or false positives, in cases in which one or the other is necessary? (e.g., "Don't you run." and "Don't you run?" can both be correct.)


danielnaber:
Thanks, I have manually merged your changes. It's always a good idea to keep the type=correct examples, to make sure there's no false alarm.

PR #2892, opened 2020-05-10
Remove redundant usage

asashour:
This PR removes:
- Redundant types
- Redundant cast
- Redundant string construction
- Unnecessary escaped character
- No need to check for existence of an object before adding into a `Set`

Which where mostly detected by IntelliJ cleanup.

Also, `System.lineSeparator()` should better be used.


danielnaber:
Thanks. Could you remove the change for `BertLmProto.java`, as it's auto-generated?

asashour:
Done, thanks.

PR #2893, opened 2020-05-10
Use collection.isEmpty()

asashour:
This is for better readability and slightly better performance.


danielnaber:
Thanks. I'm not convinced that this always improves readability, though. `size() > 0` seems more logical to me than the negated `isEmpty()`

asashour:
I also see your point.

Now only `.size() == 0` was changed.

PR #2904, opened 2020-05-11
Use Files.newInputStream instead of FileInputStream

asashour:
As hinted in a [PMD suggestion](https://pmd.github.io/latest/pmd_rules_java_performance.html#avoidfilestream):

>The FileInputStream and FileOutputStream classes contains a finalizer method which will cause garbage collection pauses. See [JDK-8080225](https://bugs.openjdk.java.net/browse/JDK-8080225) for details.

This PR resolves that, also it adds `try-with-resource` in some missing places.

Also, `File` could be moved to `Path`, but I don't know the deprecation policy in place.

Some places may lack the stream `.close()`, e.g. `Word2VecModel`, but I would check that in a separate PR.


asashour:
Closing, as it seems not of much interest.

Issue #2900, opened 2020-05-11
[fr] "et ca" suggests "etc a"


tiff:
> Fortes pluies : déjà 110 interventions dans les Landes et ca continue.

I think correctly it would be "et ça" but the spell checker doesn't get that.
<img width="758" alt="Bildschirmfoto 2020-05-11 um 11 03 32" src="https://user-images.githubusercontent.com/37363/81544011-2aff7f00-9377-11ea-91bf-c2aa998b3ea6.png">

Can we improve this somehow (maybe with a simple rule?) @jaumeortola @dpelle @vkyfox

tiff:
Probably a similar issue:
> La journé internationale de lutte contre le glaucome a été célébrée ce 12 mars.

<img width="654" alt="Bildschirmfoto 2020-05-11 um 11 07 36" src="https://user-images.githubusercontent.com/37363/81544317-a82af400-9377-11ea-9bc7-c9d20267c6a6.png">

vkyfox:
I can definitely add case-by-case rules for the correct suggestions; but I don't know how many such occurrences there are. We might be missing a lot of those by correcting them individually.

jaumeortola:
I will take a look at the spelling rule for wrong split words.

vkyfox:
I added those two rules in the meanwhile.

jaumeortola:
This kind of problems has been fixed for languages not using the Hunspell spelling rule. But French uses Hunspell. 

I would like to understand what exactly is achieved with Hunspell that cannot be achieved by other means. I guess it is related to giving good suggestions for words with hyphens and apostrophes. Is this the reason? For example: `*Etais-tu > Étais-tu, *la-bas > là-bas`. Or is there anything more? Most combinations are in the spelling dictionary (là-bas), only a few are missing (étais-tu). @danielnaber

danielnaber:
I don't remember about French, but we use Hunspell for German because it supports compound words. Morfologik doesn't. We'll maybe have to dig into the git history about why hunspell is also used for French.

dpelle:
I'm guessing that to solve this, typos related to diacritics (journee→ journée,  ca → ça) should be considered minor and thus thus when making suggestions, a difference with diacritics only should be at the top of suggestion. I assume Hunspell does that.  I've actually never been so happy with using Morfologik for spelling as opposed to Hunspell. Hunspell does a better job in general, but unfortunately it's slow.  Hunspell is also easier to integrate since it's just 2 text files.

[...]

PR #2909, opened 2020-05-11
Vkyfox

vkyfox:
[fr] rule improvements, added new rules #2900 and anglicisme

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -8009,7 +8009,7 @@ though, and has since been slightly modified)-->
       </antipattern>
       <pattern>
         <token regexp="yes">[dljmnst]e|la|jusque|que</token>
-        <token regexp="yes">h?[àâaéêèeiîou]\D.*|à|hy\D.*|y[pts].*
+        <token regexp="yes">h?[àâaéêèeiîouœ]\D.*|à|hy\D.*|y[pts].*
           <exception inflected="yes" regexp="yes">&exceptions_ELISION1;</exception>
           <exception regexp="yes">&exceptions_ELISION2;</exception><!-- https://fr.wikipedia.org/wiki/La_Haye -->
           <exception postag="UNKNOWN|Z.*" postag_regexp="yes" regexp="yes">h.+</exception><!-- no error for unknown words or proper names -->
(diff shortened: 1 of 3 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -17896,35 +17950,77 @@ though, and has since been slightly modified)-->
...
+              <example correction="captures d'écran">J'en ai pris un <marker>screen shots</marker> si tu veux.</example>
+          </rule>
+      </rulegroup>
+      <rulegroup id="SCREENCAST" name="Anglicisme : screencast" default="temp_off">
+          <rule>
+              <pattern>
+                  <token regexp="yes">screencasts?</token>
+              </pattern>
+              <message>« \1 » est un anglicisme. L'équivalent français est « capture(s) vidéo ».</message>
+              <suggestion><match no="1" regexp_match="(screencast)(s)?" regexp_replace="capture$2"/> vidéo</suggestion>
```
tiff, line 17980: Will this also work when "Screencast" is at the beginning of a sentence and therefore capitalized? @vkyfox

tiff: Same applies to the other rules.

jaumeortola: @vkyfox You need to add `regexp_match="(?iu)(screencast)(s)?"` to make it case-insensitive.

vkyfox: Such does not seem to be the case in <pattern> rules as the capitalized examples I added are detected in the test runs. I suppose this has to do with the case_sensitive attribute for the token taking priority on the regexp itself?
I had to adjust the regexp in the suggestion for it to work properly, however, so thank you for bringing this to my attention!

jaumeortola: The pattern rules are case-insensitive by default. But the `regexp_match` (for generating suggestions) is case-sensitive. You can do `regexp_match="([Ss]creenshot)(s)?"` (for capitalized words) or regexp_match="(?iu)(screencast)(s)?" (fully case-insensitive).

vkyfox: I did.

PR #2914, opened 2020-05-12
Use BufferedReader instead of Scanner

asashour:


asashour:
Closing, as it seems not of much interest.

PR #3120, opened 2020-06-18
[fr] disable and rework some anglicism rules 

tiff:
* disabled some rules that have a very low acceptance  of less than 5% (show suggestions → apply suggestion). We can enable them again once we have a "picky" mode
* added words to the dictionary so that they are available to the spell checker
* improved the suggestions of some anglicism rules by offering a plural form
* added antipatterns based on common proper nouns or idioms

cc @jaumeortola @vkyfox


jaumeortola:
I see nothing wrong in the changes.

vkyfox:
I couldn't see any mistakes or further antipatterns to add.

Issue #1500, opened 2019-04-03
[en] false alarm DT_PRP


tiff:
> It's an I/O device

![Bildschirmfoto 2019-04-03 um 09 05 58](https://user-images.githubusercontent.com/37363/55459378-b7c20000-55ef-11e9-84fe-4701d8192607.png)

I/O = input/output

PR #3186, opened 2020-06-30
[fr] #1500

vkyfox:


vkyfox:
Adjusted ANGLICISM for suggestions to account for plural.

PR #3368, opened 2020-07-29
[fr] add rule être + de + verbe à l'infinitif

thelittlefireman:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -32208,6 +32208,19 @@ though, and has since been slightly modified)-->
       <example><marker>savoir gré</marker></example>
       <example>Le contrat est négocié de gré à gré.</example>
     </rule>
+    <rule id="ETRE_DE_VERBE" name="être de + verbe inf">
```
jaumeortola, line 32211: Please, add here `default="temp_off"` in new rules, so they can be checked with a corpus before being enabled.

thelittlefireman: Ok i will remember.
How can i check with a french corpus in development ?

jaumeortola: You can check rules here (before committing): https://community.languagetool.org/ruleEditor/expert
You can add "?devMode=true" to get even more results. 

After committing, there are daily tests that check for differences. If you want to see these differences, you can be added to a developer mailing list by @danielnaber.

PR #3374, opened 2020-07-30
[fr] improve ce/se rule

thelittlefireman:


jaumeortola:
Thanks. 
Perhaps we could just add a specific rule for `ce placer (never or rarely correct?) > se placer` (except "ce placer d'or).
I didn't find false alarms for the rule. Just sentences where the error in fact is not "ce" but other word: `Comment réparer ce *ficher (fichier) ?`

thelittlefireman:
Thanks for your comment.
I don't understand `ce placer d'or` what does it mean in french ? Do you have an example ?
Maybe we could add `fichier` as an exception ?

There is some other word like placer that are recognized as a noun but i have'nt looking for all os them. Maybe we could add more words in the futur ?

jaumeortola:
> I don't understand `ce placer d'or` what does it mean in french ? Do you have an example ?

"placer" as a noun, according to the dictionaries, is "Gisement d'or, de pierres précieuses, de minerais lourds" or "Gisement secondaire de roches sédimentaires, le plus souvent d'origine alluvionnaire...". I guess it is infrequent and very technical. 

I would do this:
- A specific rule for "ce/se placer".
- A specific rule for "ce/se faire + infinitive". 

We could search in our dictionaries for more cases of noun/infinitive like "placer". 

If the confusion fichier/ficher is common, it would be useful to try to detect it.

jaumeortola:
I merge the rule, but I will make some adjustments.

PR #3379, opened 2020-08-02
[fr] Update byte rule for lower case Mb

LlianeFR:
Hello,
The rule currently doesn't work for Gb, as it suggests replacing it by Gb (works for GB).

image


jaumeortola:
Thank you.

PR #3385, opened 2020-08-03
Vkyfox

vkyfox:


languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/added.txt
```diff
@@ -711,4 +711,6 @@ Bouray-sur-Juine	Bouray-sur-Juine	Z e s
 °C	°C	N m sp
 O'Connor	O'Connor	Z e sp
 Barnier	Barnier	Z e sp
-CEPS	CEPS	Z m s
\ No newline at end of file
+CEPS	CEPS	Z m s
+Tik	Tik	Z m s
+Tok	Tok	Z m s
```
jaumeortola, line 716: But the name of the social network is one word: TikTok. Perhaps we need a French rule for Tik Tok > TikTok.

PR #3398, opened 2020-08-06
[fr] adjusted rules to account for elision

vkyfox:


udomai:
After further thought: I think we should create two entities, one for verbs starting with _h aspiré_ and one for nouns with _h aspiré_. The verb entity should contain this simple way of accounting for all the verb forms: `hauss.*`  etc.
For the nouns, the entity can be without `.*` (but plural forms need to be accounted for!). This avoids false recognitions (_pas **de hauss**e_ is correct, but so is _le Paris **d'Hauss**mann_). If we had only one entity for both verbs and nouns, the Haussmann example would be falsely identified as having an _h aspiré_.
The entity for verbs with _h aspiré_ must then be applied for tokens where the _h aspiré_ is relevant for a verb (e. g. in POUR_PAS_QUE).

jaumeortola:
To include all the inflected forms of a word, use: `<token regexp="yes" inflected="yes">&H_ASPIRE;</token>` (it only works if it is word properly tagged in the dictionary). 
We could divide `&H_ASPIRE;` in different entities, or we can select the words we need (nouns, verbs...). To select only verbs, for example: `<token postag="V.*" postag_regexp="yes" regexp="yes" inflected="yes">&H_ASPIRE;</token>`

udomai:
That is an excellent idea, Jaume! And it saves so much time :) thank you!!!

Théotime, could you please make sure that all the rules where `&H_ASPIRE;` applies to a verb will only select verbs, including their inflected forms, the way Jaume proposed it?
And could you check that everywhere `&H_ASPIRE;` applies to a noun, it accounts for the plural form (also by using `inflected="yes"`)?

PR #3460, opened 2020-08-20
Added words: jitter, jittering, jitteriness

SethiPandi:
Added the following words: 
* [jitter](https://www.wordnik.com/words/jitter)
* [jittering](https://www.wordnik.com/words/jittering)
* [jitteriness](https://www.wordnik.com/words/jitteriness)

Not 100% sure if I added them in the right place, so please just let me know if there's a problem!


danielnaber:
Yes, it's the right place, thanks!

PR #3543, opened 2020-09-08
Vkyfox

vkyfox:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/replace.txt
```diff
@@ -5,13 +5,13 @@
 # Both 'phrase' and 'replacement' can contain more than one value when separated by '|'.
 # Example:
 # incorrect=suggestion one|suggestion two
-ya=y a
+jai=j'ai
+ya=y'a|y a|il y a
```
jaumeortola, line 9: I don't know if "ya" is a common misspelling. If it is, I suggest writing different XML rules with the appropriate replacement. For example, for `il n'ya` the only suggestion should be `il n'y a`, etc.

jaumeortola (approved):

PR #3549, opened 2020-09-09
Vkyfox

vkyfox:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -9007,6 +9005,31 @@ though, and has since been slightly modified)-->
...
+                <pattern>
+                        <marker>
+                            <token>ya</token>
+                        </marker>
+                </pattern>
+                <message>Cette formulation est considérée familière. Vouliez-vous dire <suggestion>y a</suggestion> ?</message>
+                <example correction="Y a"><marker>Ya</marker> douze personnes.</example>
+            </rule>
+        </rulegroup>
       <rule id="COMME_ME_LE_DIRE" name="comme me le dire accord">
```
jaumeortola, line 9033: I don't see the point of the two rules for "ya". The suggestion is the same in both rules. Is "ya" sometimes colloquial (as it is implied in the second rule) or is it always wrong? 

And what about "y'a"? We need something similar to avoid this duplicate suggestion: 
imatge

udomai: _ya_ and _y'a_ are both written colloquialisms. They can occur with and without _il_ before them. @vkyfox, I think we need to differentiate: when there's an _il_ before _ya_ or _y'a_, we can treat it like a typo, no need to acknowledge the colloquial form. When there's no _il_ before _ya_ or _y'a_, your message is perfect, of course.

vkyfox: I updated the pull request, it should be good now.

PR #3662, opened 2020-10-08
[en] fix FPs for WHOSE_DT

udomai:


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -6722,14 +6722,26 @@ USA
             </rule>
             <!-- whose .*/DT::word=who's \1::pivots=whose,who's -->
             <rule id="WHOSE_DT" name="whose DT (possessive)">
+                <antipattern><!-- I wonder whose these beautiful flowers on the counter are. -->
+                    <token>whose</token>
+                    <token skip="-1" regexp="yes">this|that|these|those</token>
+                    <token regexp="yes">be|is|are|was|were|been</token>
```
udomai, line 6728: This looks silly, but it didn't work like this (example in line 6743 triggered alarm):

`<token postag="VB[PZDN]" postag_regexp="yes" inflected="yes">be</token>`

MikeUnwalla: 'be' has the postag VB. Try:  `postag="VB[PZDN]?"` or  `postag="VB.*"`

udomai: `VB.*` wasn't an option because I didn't want "being" in there.
I've tried your other suggestion, but the result was the same as for `postag="VB[PZDN]"`: If I make the third token of the antipattern `<token postag="VB[PZDN]?" postag_regexp="yes" inflected="yes">be</token>`, this example triggers the alarm although it shouldn't:

_I wonder whose these beautiful flowers on the counter are._

That's why I went with the above solution — it doesn't look very clever, but it does the job. I will check with Christopher later today why that might be!

udomai: I've found the culprit: 't was the disambiguator. "are" is being disambiguated into a `NN`, for reasons unknown.
image

PR #3724, opened 2020-10-19
speed up XmlRuleDisambiguator

donnerpeter:
don't even run rules that can't possibly match, judging by their token strings/regexps

to determine that, parse simple yet most frequent regexps that have just |?[] in them, extract all possible values from them
and use that set for faster PatternToken.isStringTokenMatched and to build index of rules by word form in XmlRuleDisambiguator

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/patterns/PatternRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/patterns/PatternRule.java
@@ -18,7 +18,9 @@
  */
 package org.languagetool.rules.patterns;
 
-import org.languagetool.*;
+import org.languagetool.AnalyzedSentence;
+import org.languagetool.JLanguageTool;
+import org.languagetool.Language;
 import org.languagetool.rules.RuleMatch;
 import org.languagetool.tagging.disambiguation.rules.DisambiguationPatternRule;
 import org.languagetool.tools.StringTools;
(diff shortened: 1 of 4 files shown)


donnerpeter:
For me, this speeds up the `testGrammarRulesFromXML` part of `EnglishPatternRuleTest.testRules` by about 40%.

danielnaber:
Peter, I can confirm the speed-up, this is great! I'd suggest moving calcFormHints etc. to its own class. Could you also write some tests for it and add a comment about how the actual optimization in `disambiguate` works?

donnerpeter:
`calcFormHints` is pretty coupled with `PatternToken` internals. What would you imagine the responsibility of that new class to be? Or did you mean just `getPossibleRegexpValues` which is easy to extract to some utility and write tests about?

As a comment, would this be enough?
     /**
   * @return rules that have a chance to be triggered by the given sentence, e.g. the ones that match against tokens
   * present there, or with too general regular expressions.
   */

danielnaber:
> Or did you mean just getPossibleRegexpValues which is easy to extract to some utility and write tests about?

Yes, I care mostly about the tests.

jaumeortola:
In languages with a large disambiguation file the speed-up is very noticeable. Thank you! 

Could this approach be used with the grammar files?

donnerpeter:
@danielnaber Would a couple of tests in `PatternTokenTest` without class extraction be OK as well?

@jaumeortola You're welcome! I believe `simpleRuleTokens`/`inflectedRuleTokens` already serve a similar purpose with grammar rules, but enhancements from regex parsing might also benefit the performance. Any idea of a test that runs a lot of rules on a lot of sentences, where the effects could be noticeable?

danielnaber:
@donnerpeter Yes, just a test would be fine.

danielnaber:
I've created an issue at https://github.com/languagetool-org/languagetool/issues/3744 with an idea on how we can maybe improve performance for regex also for the pattern rules.

PR #3763, opened 2020-10-27
Using try-with-resources block to close the Scanner which reads the aff file.

robertgherlan:
Using try-with-resources block to close the Scanner which reads the aff file.


danielnaber:
Thanks!

Issue #3775, opened 2020-10-29
[en] rule for is seeming -> seems


plumaj:
As discussed with @udomai 

> It is seeming that this is not a good idea.

Should be **seems**

EDIT: Can't assign myself yet, but I am working on this!

danielnaber:
See https://github.com/languagetool-org/languagetool/pull/3781

PR #3776, opened 2020-10-29
[en] add SEEMING_SEEMS for #3775

plumaj:
Rule added for #3775 on request of @udomai


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -14776,6 +14776,21 @@ USA
     <!-- Grammar -->
     <!-- ====================================================================== -->
     <category id="GRAMMAR" name="Grammar" type="grammar">
+        <rule id="SEEMING_SEEMS" name="confusion of is seeming/seems">
```
danielnaber, line 14779: Even though the pattern looks very specific, I still think we should use `default="temp_off"`, just to be sure it's tested first in the nightly test runs.

PR #3777, opened 2020-10-29
[fr] added rule SONT_SOMMES #1984

vkyfox:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -9026,6 +9026,23 @@ though, and has since been slightly modified)-->
         <example correction="trois fois par semaine">Je joue au Softball <marker>trois fois au semaine</marker>.</example>
       </rule>
       <!-- End of additions -->
+      <rule id="SONT_SOMMES" name="sont au lieu de sommes">
```
danielnaber, line 9029: Shouldn't this have `default="temp_off"`?

vkyfox: Done, sorry about that.

PR #3781, opened 2020-10-30
[en] add SEEMING_SEEMS for #3775 (temp_off)

plumaj:
Add SEEMING_SEEMS for #3775, this time as `temp_off`


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -14776,6 +14776,21 @@ USA
...
+                <token>am</token>
+                <token>seeming</token>
+                <token>like</token>
+            </antipattern>
+            <pattern>
+                <marker>
+                    <token inflected="yes">be</token>
+                    <token>seeming</token>
+                </marker>
+            </pattern>
```
udomai, line 14790: Could there be FPs for "seeming" used as an adjective?

_There are seeming differences between these instances, but on closer inspection, they turn out to be negligible."

PR #3801, opened 2020-11-06
call sentence.getText only when it's needed to speed up the analysis

donnerpeter:
For me, this results in ~10% English text checking speedup.


jaumeortola:
For me, same percentage of speedup (~10%) checking Spanish and Catalan.

PR #3802, opened 2020-11-06
PatternRule: check for antipatterns only after a rule actually produced problems

donnerpeter:
because immunization is pretty expensive


donnerpeter:
For me, this results in ~40% speedup for English text analysis

jaumeortola:
I get around 20% speedup analyzing Spanish and Catalan texts.

danielnaber:
@donnerpeter Thanks, I can confirm it speeds up English a lot, will merge now.

Issue #3744, opened 2020-10-22
improve performance for simple regex?


danielnaber:
Similar to https://github.com/languagetool-org/languagetool/pull/3724: maybe we can change https://github.com/languagetool-org/languagetool/blob/98f75c2e69bfbd4d1d11edfa4db8217f855c5a6a/languagetool-core/src/main/java/org/languagetool/rules/patterns/PatternRule.java#L247 to also allow simple regex in the form of `a|b|c` (and no other regex-specific chars) and extract the words from them?

PR #3804, opened 2020-11-06
PatternRule.canBeIgnoredFor: check token form hints as well #3744

donnerpeter:
to avoid expensive matching for rules whose simple regexps are definitely inapplicable to the current sentence


donnerpeter:
This shouldn't be merged before #3808 is fixed

danielnaber:
Do you have any numbers of how much this helps performance?

donnerpeter:
AFAIR it was about 7-10% for English after applying all my other recent optimizations

Issue #3808, opened 2020-11-06
calcFormHints and OR


danielnaber:
@donnerpeter I wonder, shouldn't [this test](https://github.com/languagetool-org/languagetool/blob/master/languagetool-core/src/test/java/org/languagetool/rules/patterns/PatternTokenTest.java#L206-L209) case work?

 ```
    token = new PatternTokenBuilder().tokenRegex("an|the").build();
    token.setOrGroupElement(new PatternTokenBuilder().tokenRegex("foo|bar").build());
    assertEquals(Sets.newHashSet("an", "the", "foo", "bar"), token.calcFormHints());
 ```

It fails with:
```
Expected :[the, bar, an, foo]
Actual   :[the, an]
```

donnerpeter:
Yes, I believe you're right and it's a bug. I'll fix it, thanks!

PR #3815, opened 2020-11-09
PatternRuleTest: check rule's canBeIgnoredFor sanity

donnerpeter:
To catch issues like #3808 earlier


donnerpeter:
Or, maybe better, what do you think of invoking `canBeIgnoredFor` right from `org.languagetool.rules.patterns.PatternRule#match`?

danielnaber:
> Or, maybe better, what do you think of invoking canBeIgnoredFor right from org.languagetool.rules.patterns.PatternRule#match?

That sounds more logical to me than adding a special case to the test.

donnerpeter:
OK, I'll see how this plays out with other optimizations I'm planning

PR #3828, opened 2020-11-10
PatternRule: check all form hints, sort them by length to fail faster

donnerpeter:


donnerpeter:
I get some 5-10% improvement in English with this

PR #3829, opened 2020-11-10
speedup command line error reporting on large texts

donnerpeter:
don't replace/mark in entire text only to extract a tiny substring afterwards

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/tools/ContextTools.java
+++ b/languagetool-core/src/main/java/org/languagetool/tools/ContextTools.java
@@ -18,6 +18,8 @@
  */
 package org.languagetool.tools;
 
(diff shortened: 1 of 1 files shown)


danielnaber:
Thanks! Have you measured how much of a difference this makes?

donnerpeter:
Ah sorry. It's about 5 seconds on 10K English sentences with about 1.5K errors

danielnaber:
I assume you measured this more than once, didn't you? I'm asking because I often see a lot of variation between runs when I try to test performance improvements. Also, that's 5 seconds compared to what, i.e. how much is it in percent?

donnerpeter:
Yes, I usually measure several (about 3) runs, more if the difference isn't very visible.
It's 5 seconds as in 35->30 :)

PR #3833, opened 2020-11-10
[fr] added rule VOUS_LE_PPA_1ER_GROUPE #1989, minor other rule adjust…

vkyfox:
…ments

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -9026,6 +9026,17 @@ though, and has since been slightly modified)-->
         <example correction="trois fois par semaine">Je joue au Softball <marker>trois fois au semaine</marker>.</example>
       </rule>
       <!-- End of additions -->
+      <rule id="VOUS_LE_PPA_1ER_GROUPE" name="s manquant à vous le ppa" default="temp_off">
+          <pattern>
+              <token>vous</token>
+              <token>le</token>
+              <marker>
+                  <token postag="V ppa.*s" regexp="yes" postag_regexp="yes">\w+e</token>
+              </marker>
+          </pattern>
+          <message>Il semblerait manquer la marque du pluriel. Vouliez-vous dire « <suggestion>\3s</suggestion> ».</message>
+          <example correction="faites">Comme vous le <marker>faite</marker>.</example>
+      </rule>
       <rule id="CONFUSION_PAR_PART" name="Confusion part/par expression de par" default="temp_off">
           <pattern>
               <token>de</token>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -9175,6 +9186,17 @@ though, and has since been slightly modified)-->
       </rule>
       <rulegroup id="D_N_QUE_R_AVOIR_PPA" name="Relative avec erreur d'accord" default="off">
           <rule name="f s">
```
jaumeortola, line 9188: default="off" or default="temp_off" to see the results?

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -9026,6 +9026,17 @@ though, and has since been slightly modified)-->
         <example correction="trois fois par semaine">Je joue au Softball <marker>trois fois au semaine</marker>.</example>
       </rule>
       <!-- End of additions -->
+      <rule id="VOUS_LE_PPA_1ER_GROUPE" name="s manquant à vous le ppa" default="temp_off">
+          <pattern>
+              <token>vous</token>
+              <token>le</token>
```
udomai, line 9032: Would this also work with more freedom in the second token?

_C'est comme vous le dite._
_Pourquoi vous faite cela ?_
_Vous ne me dite jamais rien._

By the way, we will have to prioritize this rule over `CONFUSION_RULE_PREMIUM` since that rule would suggest the infinitive here. Otherwise it won't even show up in the diff...

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -9175,6 +9186,17 @@ though, and has since been slightly modified)-->
       </rule>
       <rulegroup id="D_N_QUE_R_AVOIR_PPA" name="Relative avec erreur d'accord" default="off">
           <rule name="f s">
+              <antipattern>
+                  <token postag="SENT_START"/>
+                  <token postag="D f s"/>
+                  <token min="0" max="1" postag="A"/>
+                  <token postag="N f s"/>
+                  <token regexp="yes">qu[e']</token>
```
udomai, line 9194: We might want to include the entire palette of possible apostrophe candidates: `[’´'‛′‘]`

jaumeortola: As it is now, `<token regexp="yes">qu[e']</token>` covers (internally) typewriter apostrophe (') and typographical apostrophe (’). The other characters are not tokenized this way. They are separate tokens. So, it is pointless to add them here. They are considered typographical errors by other rules.

udomai: Okay, cool! Thank you for the info :)

PR #3834, opened 2020-11-10
decrease the number of line/column calculations

donnerpeter:
precompute them just once for all sentences, not for each rule afresh
use this data to speedup text-level match anchoring

This gives me some 5-10% speedup for English

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
+++ b/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
@@ -1830,7 +1830,7 @@ public class JLanguageTool {
       int low = 0;
       int high = sentences.size() - 1;
       while (low <= high) {
-        int mid = (low + high) >>> 1;
+        int mid = (low + high) / 2;
         SentenceData sentence = sentences.get(mid);
         if (sentence.startOffset < offset) low = mid + 1;
         else if (sentence.startOffset > offset) high = mid - 1;
--- a/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
+++ b/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
@@ -1151,9 +1151,10 @@ public class JLanguageTool {
   /**
    * @since 3.7
    */
-  protected List<RuleMatch> performCheck(List<AnalyzedSentence> analyzedSentences, List<String> sentences,
+  protected List<RuleMatch> performCheck(List<AnalyzedSentence> analyzedSentences, List<String> sentenceTexts,
                                          List<Rule> allRules, ParagraphHandling paraMode, AnnotatedText annotatedText, RuleMatchListener listener, Mode mode, Level level, boolean checkRemoteRules) throws IOException {
-    Callable<List<RuleMatch>> matcher = new TextCheckCallable(allRules, sentences, analyzedSentences, paraMode, annotatedText, 0, 0, 1, listener, mode, level, checkRemoteRules);
+    List<SentenceData> sentences = computeSentenceData(analyzedSentences, sentenceTexts);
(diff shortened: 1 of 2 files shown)


languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
```diff
@@ -1772,56 +1806,37 @@ public static void removeTemporaryFiles() {
...
+      return new LineColumnPosition(
+        sentence.startLine + countLineBreaks(prefix),
+        processColumnChange(sentence.startColumn, prefix));
+    }
+
+    private SentenceData findSentenceContaining(int offset) {
+      int low = 0;
+      int high = sentences.size() - 1;
+      while (low <= high) {
+        int mid = (low + high) >>> 1;
```
danielnaber, line 1833: This operator is not very common, could you add a comment?

danielnaber (changes requested):

languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
```diff
@@ -1703,14 +1743,10 @@ public static void removeTemporaryFiles() {
...
-            if (match.getLine() == 0) {
-              newMatch.setColumn(range.from.column + 1);
-            } else {
-              newMatch.setColumn(range.from.column);
-            }
-            newMatch.setEndColumn(range.to.column);
+            newMatch.setLine(from.line);
+            newMatch.setEndLine(to.line);
+            newMatch.setColumn(from.column - (from.line == 0 ? 1 : 0));
+            newMatch.setEndColumn(to.column - (to.line == 0 ? 1 : 0));
```
donnerpeter, line 1749: Here's an inconsistency. Some clients expect column to be 0-based (as in javadoc) and some — 1-based. From the history I've seen that the similar condition in the previous version of code was a hack to make some tests happy (since they're using deprecated API anyway), so I've decided to create an equivalent.

languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
donnerpeter: It's inspired by JDK code, but indeed, this could be simplified. Will do!

languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
danielnaber: Line/column information is deprecated in LT, we don't show it anymore in the JSON response (our main use case), we just keep it running for the command-line LT.

donnerpeter: Any plans to remove it?

danielnaber: I'd prefer to have it removed, but I see the line number is useful when running LT on the command-line, so it would need to be calculated somewhere else (maybe not in JLanguageTool). But I have no plans to work on this.

donnerpeter: Do you mean only `org.languagetool.commandline.Main` or anything else? If the former, I think this calculation could be moved there relatively easily, and `JLanguageTool` could become free of this logic, and maybe even faster.

danielnaber: I haven't checked closely, but yes, `org.languagetool.commandline.Main` sounds like the logical place for this code.

donnerpeter: I can also see some column usages in XML serialization, Wikipedia and UI HTML. I don't know of their importance nowadays. But maybe I'll take a look at this some day :)

PR #3836, opened 2020-11-11
PatternToken: reuse known possible forms

donnerpeter:
don't parse the same regexp several times


donnerpeter:
Currently this doesn't much improve performance, but I plan to enhance regexp possible value extraction, and that'd make it a bit more expensive. And it turns out that `calcFormHints` is called more than I expected, e.g. when a new fake rule is created at `DisambiguationPatternRuleReplacer#keepByDisambig`

donnerpeter:
I've figured out a better way out

PR #3845, opened 2020-11-12
store possible regexp values in an array instead of TreeMap

donnerpeter:
to reduce memory usage


danielnaber:
Thanks! I assume you have checked that this doesn't slow down LT? If we can use more memory to gain performance, we usually like to do so (for the server use case, which is the most important for us).

donnerpeter:
Yes, I've seen no difference in time to the previous code. If anything, this version might be a tiny bit faster due to memory locality.

For our IDE use case, keeping memory footprint low is important. But don't worry, I'll keep an eye on the execution time as well :)

PR #3866, opened 2020-11-16
slightly reduce PatternToken size by using bytes instead of ints in f…

donnerpeter:
…ields

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/patterns/PatternToken.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/patterns/PatternToken.java
@@ -71,9 +71,9 @@ public class PatternToken implements Cloneable {
   /** List of exceptions that are valid for a previous token. */
   private List<PatternToken> previousExceptionList;
 
-  private int skip;
-  private int minOccurrence = 1;
-  private int maxOccurrence = 1;
+  private byte skip;
+  private boolean mayBeOmitted;
+  private byte maxOccurrence = 1;
 
   private Predicate<String> stringMatcher = s -> true;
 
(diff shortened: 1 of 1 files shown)


languagetool-core/src/main/java/org/languagetool/rules/patterns/PatternToken.java
```diff
@@ -434,7 +434,7 @@ public int getSkipNext() {
    * The minimum number of times the element needs to occur.
    */
   public int getMinOccurrence() {
-    return minOccurrence;
+    return mayBeOmitted ? 0 : 1;
```
danielnaber, line 437: This may be rare, but it should be possible to have a min occurrence of e.g. 2.

donnerpeter: `setMinOccurrence` forbids it now

danielnaber: Yes, but should we do so? There's `min="x"` in the XML and I think this should allow values > 1, even if this is rare.

donnerpeter: I don't know, I only took existing contracts and optimized based on them :)

I think, in the unlikely case where `min="2"` is really-really needed, it can be simulated by creating two similar `<token>` tags.

danielnaber: Ah sorry, I missed the fact that the restriction to 0 or 1 existed forever...

PR #3885, opened 2020-11-19
Update thresholds logic for the ngram language identifier and add test class

nopepper:
This is a fix for issue #2002 in lt-premium
Thresholds are now token-based and cover a range [min, max) for each language. I also added a class to test some examples.


languagetool-core/src/test/java/org/languagetool/language/NGramLangIdTest.java
```diff
@@ -0,0 +1,171 @@
...
+    public final List<Double> pred;
+
+    public ProcessedSentence(String part1, String part2, String part3) {
+     sentence = part1;
+     encoding = Arrays.stream(part2.split(" ")).map(Integer::parseInt).collect(Collectors.toList());
+     pred = Arrays.stream(part3.split(" ")).map(Double::parseDouble).collect(Collectors.toList());
+    }
+  }
+
+  private final File TESTER_MODEL = new File("C:\\Users\\Robert\\Desktop\\ngram-language-identification\\testing_model\\model\\model.zip");
```
danielnaber, line 34: How large is this file? Should it be in git? If not, the test will need to be `@Ignore`.

nopepper: Yes, the file contains a trimmed model (~1.5MB) that is extracted from the main one - I'll add it to git once I got the final file. I noticed that the current model with a vocabulary of 30k is too big/slow, so I'm reducing it.

PR #3887, opened 2020-11-19
MorfologikSpellerRule: calculate suggestions only when they're needed

donnerpeter:
add new RuleMatch.setLazySuggestedReplacements API for that

For single-threaded analysis of AmericanEnglish without requesting suggestions, this gives ~40-45% analysis speedup.


danielnaber:
Thanks. It would be great if you could put the cleanup-related changes (removing `this` etc) into a separate PR, to make it easier to read.

danielnaber:
> Would you like me to do that here or only in future PRs?

It's enough for future PRs.

PR #3892, opened 2020-11-19
ManualSynthesizer/ManualTagger: reduce memory overhead

donnerpeter:
don't waste space on arrays of length 1

For English, this saves a bit more than 1MB


languagetool-core/src/main/java/org/languagetool/tools/MostlySingularMultiMap.java
```diff
@@ -0,0 +1,35 @@
+package org.languagetool.tools;
```
danielnaber, line 1: Could you please add the LGPL license header here? (see other .java files)

danielnaber (changes requested):

PR #3901, opened 2020-11-20
HunspellRule: calculate suggestions lazily

donnerpeter:
This speeds up single-threaded German checking (without suggestion retrieval) by ~50%

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/HunspellRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/HunspellRule.java
@@ -29,6 +29,8 @@ import org.languagetool.rules.Categories;
 import org.languagetool.rules.RuleMatch;
 import org.languagetool.rules.SuggestedReplacement;
 import org.languagetool.rules.spelling.SpellingCheckRule;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.*;
 import java.net.URISyntaxException;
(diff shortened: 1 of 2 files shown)


languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/HunspellRule.java
```diff
@@ -181,59 +181,16 @@ protected boolean isQuotedCompound (AnalyzedSentence analyzedSentence, int idx,
...
-                  suggestions.add(Math.min(pos, suggestions.size()), new SuggestedReplacement(suggestion.substring(0, suggestion.length()-1)));
-                  pos += 2;  // we mix the lists, as we don't know which one is the better one
+            ruleMatch.setLazySuggestedReplacements(() -> {
+              try {
+                synchronized (this) { // some getSuggestions overrides are thread-unsafe
+                  return calcSuggestions(word, cleanWord);
                 }
+              } catch (IOException e) {
+                logger.error("Error while calculating speller suggestions", e);
+                return Collections.singletonList(new SuggestedReplacement("Internal error: " + e.getMessage()));
```
danielnaber, line 191: Do you know when this can happen? I wonder whether it makes sense to give out the error. in general, we tend to throw an exception instead.

donnerpeter: The exception can be thrown from `init()`, but it seems `init` should be already done by the time `getSuggestions` is called (at least in the current codebase, I can imagine someone calling `getSuggestions` explicitly without preceding `match`). So if you believe it's safe, I can remove `needsInit` checks from `getSuggestions`.

My idea with message was to give users indication that something went wrong. I can rethrow a `RuntimeException` if you think it's better. `MorfologikSpellerRule#appendLazySuggestions` would need to be changed as well.

danielnaber: > I can rethrow a RuntimeException if you think it's better.

Yes, I think it's better to go with the "fail early" approach that serves us well so far.

PR #3906, opened 2020-11-23
JLanguageTool: cache active rules

donnerpeter:
to avoid calling `ignoreRule` for each rule in each `check` invocation, often for each sentence
and to avoid filtering out picky rules by tag on each `check` call

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
+++ b/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
@@ -1155,6 +1155,10 @@ public class JLanguageTool {
     }
   }
 
+  /**
+   * @deprecated use {@link #performCheck(List, List, List, ParagraphHandling, AnnotatedText, RuleMatchListener, Mode, Level, boolean)}
+   */
+  @Deprecated
   protected List<RuleMatch> performCheck(List<AnalyzedSentence> analyzedSentences, List<String> sentences,
                                          List<Rule> allRules, ParagraphHandling paraMode, AnnotatedText annotatedText, Mode mode, Level level) throws IOException {
     List<Rule> nonIgnored = allRules.stream().filter(r -> !ignoreRule(r)).collect(Collectors.toList());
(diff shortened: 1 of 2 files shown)


languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
```diff
@@ -1145,7 +1157,8 @@ protected void printSentenceInfo(AnalyzedSentence analyzedSentence) {
 
   protected List<RuleMatch> performCheck(List<AnalyzedSentence> analyzedSentences, List<String> sentences,
                                          List<Rule> allRules, ParagraphHandling paraMode, AnnotatedText annotatedText, Mode mode, Level level) throws IOException {
-    return performCheck(analyzedSentences, sentences, allRules, paraMode, annotatedText, null, mode, level, true);
+    List<Rule> nonIgnored = allRules.stream().filter(r -> !ignoreRule(r)).collect(Collectors.toList());
```
donnerpeter, line 1160: This method is unused in LT, so I can remove it instead of changing. Not sure if there are external callers though.

languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
```diff
@@ -1201,7 +1214,8 @@ private int processColumnChange(int columnCount, String sentence) {
    */
   public List<RuleMatch> checkAnalyzedSentence(ParagraphHandling paraMode,
                                                List<Rule> rules, AnalyzedSentence analyzedSentence) throws IOException {
-    return checkAnalyzedSentence(paraMode, rules, analyzedSentence, false);
+    List<Rule> nonIgnored = rules.stream().filter(r -> !ignoreRule(r)).collect(Collectors.toList());
```
donnerpeter, line 1217: Another unused method, public

donnerpeter:
This mostly benefits performance in scenarios where `check` is called multiple times, e.g. `EnglishTest#testLanguage` goes down 28->24 seconds for me.

languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
danielnaber: As LT can be used from Java directly via its API, we should expect public methods to be used. We can clean them up, but for that, we should first deprecate them and only remove them 2 releases later or so.

donnerpeter: This one is deprecated.
So, should I deprecate these methods?

languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
donnerpeter: Also documented as "This is an internal method that's public only for technical reasons, please use one of the {@link #check(String)} methods instead"

languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
danielnaber: > So, should I deprecate these methods?

Yes

PR #3921, opened 2020-11-24
synchronize thread-unsafe MorfologikSpeller queries

donnerpeter:


languagetool-core/src/main/java/org/languagetool/rules/spelling/morfologik/MorfologikSpeller.java
```diff
@@ -87,14 +87,16 @@ public MorfologikSpeller(String fileInClassPath) throws IOException {
...
-            && speller.isMisspelled(word);
+    if (word.isEmpty() || SpellingCheckRule.LANGUAGETOOL.equals(word) || SpellingCheckRule.LANGUAGETOOLER.equals(word)) {
+      return false;
+    }
+    synchronized (this) {
+      return speller.isMisspelled(word);
+    }
   }
 
-  public Speller getSpeller() {
```
danielnaber, line 96: As this is public, it might be in use somewhere and probably shouldn't be removed?

donnerpeter: Theoretically it can, although I don't think it's likely. This method was introduced relatively recently, and exposes internal stuff that needs synchronization, that's why I've removed it. I also see other possibilities, if you think it's so important:
1. Leave it but deprecate
2. Leave it but return a thread-safe new speller, like in `getSuggestions` below.
WDYT?

danielnaber: Okay, I suggest deprecating it

Issue #3948, opened 2020-11-30
[en] archaic VBD "staid" in synthesizer


udomai:
This one was spotted by Ophir: Rules that propose a VBD in the suggestion will display "stayed" and "staid" for the verb _stay_. Example: [HE_VERB_AGR[5]](https://internal1.languagetool.org/regression-tests/via-http/2020-11-30/en/result_grammar_HE_VERB_AGR[5].html)

We should take away the tag "VBD" from "staid" because it's really archaic and only have it as an adjective. [More info](https://www.lexico.com/definition/staid)

PR #3949, opened 2020-11-30
[en] removed.txt staid as VBD and VBN

udomai:
@tiff Will adding "staid" to removed.txt stop it from being suggested by the synthesizer? See [#3948](https://github.com/languagetool-org/languagetool/issues/3948)


tiff:
yes

PR #3957, opened 2020-12-02
StringMatcher.getPossibleRegexpValues: support more regexp constructs

donnerpeter:
This speeds up my `Main -l en` by 5-10% and decreases memory usage a bit


languagetool-language-modules/es/src/main/resources/org/languagetool/rules/es/grammar.xml
```diff
@@ -1522,7 +1522,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
             </antipattern>
             <antipattern>
                 <token regexp="yes">practica|practicas</token>
-                <token regexp="yes">tú|?</token>
+                <token regexp="yes">tú|\?</token>
```
donnerpeter, line 1525: An incorrect regexp accepted by the previous algorithm, but not by the current one. I'll probably add a bulletproof wellformedness check in a separate PR

PR #3958, opened 2020-12-02
AbstractCompoundRule: avoid locking overhead in a single-threaded queue

donnerpeter:
Minor, unlikely to result in an obvious performance improvement (except maybe if just this rule is run on huge texts), but it just felt wrong to see this in CPU snapshots.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/AbstractCompoundRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/AbstractCompoundRule.java
@@ -18,21 +18,15 @@
  */
 package org.languagetool.rules;
 
-import java.io.IOException;
-import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


danielnaber:
Thanks. Just a general remark: Let me know if I forget about one of your PRs. I often don't merge more than one optimization PR per day on purpose, so that in the unlikely case an issue isn't detected by the unit tests, it's easier to debug (we auto-deploy every day, so we'll see issues soon).

donnerpeter:
Thanks for this information, it's good to know! But how can I determine whether your forgot about a PR ot are just thinking deeply of it?

danielnaber:
> But how can I determine whether your forgot about a PR ot are just thinking deeply of it?

If there's no reaction after a week, chances are good the PR has just been forgotten...

PR #3964, opened 2020-12-03
[en] new rule to remove some false positives for Barca → Barça

hftf:
Removes `Barça` from `en/diacritics.txt` due to false positives like "Hannibal Barca" in `EN_DIACRITICS_REPLACE` and adds a new rule `BARCA_DIACRITIC`.


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -35196,6 +35196,17 @@ USA
             <message>Did you mean <suggestion><match no="1" regexp_match="riel" regexp_replace="rial" /></suggestion> (= substance) or <suggestion><match no="1" regexp_match="eriel" regexp_replace="ériel" /></suggestion> (= military equipment)?</message>
             <example correction="materials|matériels">I needed more electronic <marker>materiels</marker> to build a fully-functional machine.</example>
         </rule>
+        <rule id="BARCA_DIACRITIC" name="Barca or Barça">
```
udomai, line 35199: This rule still gives two FPs in the Rule Editor:

- If the latter, then Hannibal and his brothers also bore the name "Barca".
- Barca (, ) was the Semitic surname of his aristocratic family, meaning "shining" or "lightning".

The former can be fixed by adding skip="-1" to the first token. The latter is given away by the word "surname" afterwards. I think this might be a case for [Ruud's trick](https://dev.languagetool.org/tips-and-tricks#warn-for-misused-word-except-in-correct-context).

hftf: Thanks for the pointer! Does the new antipattern suffice, or is it redundant to except the same keywords in two places?

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -35450,6 +35450,11 @@ USA
             <example correction="materials|matériels">I needed more electronic <marker>materiels</marker> to build a fully-functional machine.</example>
         </rule>
         <rule id="BARCA_DIACRITIC" name="Barca or Barça">
+            <antipattern>
+                <token skip="-1" postag="SENT_START"/>
+                <token skip="-1" regexp="yes">surname|cognomen|Hannibal|Hamilcar|Hasdrubal</token>
+                <token postag="SENT_END"/>
+            </antipattern>
             <pattern>
                 <token>
```
udomai, line 35459: For some reason, there are three lines that are not being shown here. The rule editor shows no more FPs if this part becomes:

```
<token>barca
                    <exception scope="previous" regexp="yes">da|la|cognomen|Hannibal|Hamilcar|Hasdrubal|Latin</exception>
                </token>
            </pattern>
```
Notice the `scope="previous"` and the added `Latin`.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -35450,6 +35450,11 @@ USA
             <example correction="materials|matériels">I needed more electronic <marker>materiels</marker> to build a fully-functional machine.</example>
         </rule>
         <rule id="BARCA_DIACRITIC" name="Barca or Barça">
+            <antipattern>
+                <token skip="-1" postag="SENT_START"/>
+                <token skip="-1" regexp="yes">surname|cognomen|Hannibal|Hamilcar|Hasdrubal</token>
+                <token postag="SENT_END"/>
+            </antipattern>
```
udomai, line 35457: I think this is an interesting variant on Ruud's method. I think it makes sense that some elements are forbidden in the entire sentence (`surname` and `cognomen`). For the first names and `Latin`, it should be sufficient to exclude them directly before `Barca`. The word "Latin" is especially likely to appear in a sentence where one's actually talking about Barça.

PR #3965, opened 2020-12-03
[en] improves some false positives for !Kung

hftf:
Several improvements involving false positives where the `SENTENCE_WHITESPACE` rule suggested to "Add a space between sentences" for some names (demonyms/ethnonyms) containing terminal punctuation (exclamation points), such as `!Kung`.

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/multiwords.txt
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/multiwords.txt
@@ -1517,3 +1517,9 @@ North Dakotan	JJ
 South Devon	NNP
 Iset River	NNP
 Nizhny Novgorod	NNP
+!Kung	NNP
+!Xun	NNP
+!Xoo	NNP
+!Xóõ	NNP
+!Xoon	NNP
+!Khong	NNP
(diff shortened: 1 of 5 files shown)


MikeUnwalla:
If _!Kung_ is now a single token, then shouldn't the POS be specified in` added.txt` (as is done for `fo'c'sle`), rather than` multiwords.txt`?

Also, if a term is in `multiwords.txt`, it is not necessary to add spelling (refer to the comments at the top of `multiwords.txt`).

languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/multiwords.txt
```diff
@@ -1501,4 +1501,10 @@ Janet Yellen	NNP
...
 Keira Knightley	NNP
-Project Veritas	NNP
\ No newline at end of file
+Project Veritas	NNP
+!Kung	NNP
+!Xun	NNP
+!Xoo	NNP
+!Xóõ	NNP
+!Xoon	NNP
+!Khong	NNP
```
udomai, line 1510: See Mike's comment about using added.txt rather than multiwords

hftf:
Thanks for the advice! I should have noticed the documentation at the tops of those files more carefully (I got to this solution just from going down a few rabbit holes in the Java code).

PR #3975, opened 2020-12-07
[en] fix 1 false positive for rule OCCASION_TRANSITIVE_VERB_VERY_FORMAL

hftf:
This is intended to fix the following false positive I noticed in the nightly diff:
> On a number of occasions the Beano has allowed its readers to vote for which new strips they want to appear in the comic.


MikeUnwalla:
@hftf, if the disambiguator gives wrong tagging, please make an open-source issue for the wrong tagging.

The tagger is not always correct, but it can be useful. A possible option for the AP to to include chunk information.
`<token chunk_re="E-NP-(singular|plural) postag="N.*" postag_regexp="yes" inflected="yes">occasion</token>`

PR #3982, opened 2020-12-07
[en] improve ADVERB_WORD_ORDER[9] and [10]

### LINKED ISSUE #3969: [en] false alarm ADVERB_WORD_ORDER?
A user says that this sentence is perfectly fine and our suggestion doesn't really improve the sentence.

> It has been removed from the server and should never be run there.

Bildschirmfoto 2020-12-04 um 08 37 07

hftf:
Some improvements and test cases. Fixes #3969.

### DIFF
--- a/languagetool-language-modules/en/src/main/java/org/languagetool/language/English.java
+++ b/languagetool-language-modules/en/src/main/java/org/languagetool/language/English.java
@@ -408,7 +408,8 @@ public class English extends Language implements AutoCloseable {
       case "WE_BE":                     return -1;
       case "A_LOT_OF_NN":               return -1;
       case "IT_VBZ":                    return -1;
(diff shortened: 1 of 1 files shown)


hftf:
This pull request should be ready. But there were already many false positives for ADVERB_WORD_ORDER[9], such as:
* It doesn't **happen often**, but...
* It **happens often** enough to be a problem (Perhaps "enough" is a useful antipattern trigger)
* Oil is cheap, so **check often** and add when low.
* What you **remember usually** would involve some low amounts of keys (Here, "remember" is part of the embedding "What you remember"; the correction should be to put "usually" _after_ "would")

udomai:
Note: Needs testing either using the SentenceSourceChecker or by using a temp_off (quasi-duplicate) rule (with higher priority) with the changes in it. As discussed on 4 Jan.

PR #3984, opened 2020-12-07
[en] strip diacritics to prevent clobbering important initial letter in "a vs. an" determiner rule

hftf:
Split off from the original pull request #3967.

Improved the "a vs. an" rule by stripping diacritics and keeping the initial letter, rather than lopping it off; added test cases.

Note: The localized test case `AvsAnRuleTest.java` succeeded for me inside IntelliJ and I also manually tested the standalone build on the command line, but the full test suite still fails regardless despite new progress on troubleshooting that.


danielnaber:
Closing old PRs that don't compile or that are still in draft

PR #3988, opened 2020-12-08
StringMatcher: extract substrings from regexp that any matching text must contain

donnerpeter:
use them to speed up token text matching and regexp rules

~20% speedup in my `Main -l de`

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/patterns/Substrings.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/patterns/Substrings.java
@@ -1,0 +1,187 @@
+/* LanguageTool, a natural language style checker
+ * Copyright (C) 2020 Peter Gromov
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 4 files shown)


languagetool-core/src/main/java/org/languagetool/rules/patterns/Substrings.java
```diff
@@ -0,0 +1,187 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Arrays;
+
+/**
+ * A list of strings that are expected to occur in a larger string in the given order,
+ * with optional requirements for them to start/end the text.
+ * This class is used to avoid more expensive regular expression matching where possible.
+ */
+class Substrings {
```
danielnaber, line 30: This class would be a good candidate for a unit test, wouldn't it?

donnerpeter: It's more of an implementation detail of `StringMatcher`, extracted to keep it smaller. So most of it is already tested (indirectly) in `StringMatcherTest`.

PR #3993, opened 2020-12-09
don't invoke pattern rules with hints inapplicable to a sentence

donnerpeter:
extract RuleSet out of XmlRuleDisambiguator, support lemma hints there, use it in JLanguageTool.check as well

~5% speedup on `Main -l en` and `Main -l de`; ~10% speedup in IntelliJ-specific English setup (1 thread, no spellchecker and `NoopChunker`)

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/patterns/RuleSet.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/patterns/RuleSet.java
@@ -1,0 +1,149 @@
+/* LanguageTool, a natural language style checker
+ * Copyright (C) 2020 Daniel Naber (http://www.danielnaber.de)
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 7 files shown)


languagetool-core/src/main/java/org/languagetool/rules/patterns/RuleSet.java
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * An object holding a set of rules with an optional possibility to fetch only the ones applicable for a given sentence
+ * (via {@link #rulesForSentence}), to avoid even invoking the definitely inapplicable ones.
+ * The filtering is based on cues provided by the rules, e.g. possible token strings or lemmas in pattern rules.
+ *
+ * @since 5.2
+ */
+@ApiStatus.Internal
+public abstract class RuleSet {
```
danielnaber, line 35: Could you maybe add a unit test for this class? I think it's a good approach to have a test for any new class.

PR #4006, opened 2020-12-11
[en,de,ru] speed up language and JLanguageTool initialization

donnerpeter:
make some expensive fields lazy, so that they'll be loaded later, hopefully even in parallel

this improves startup time a bit,
and the IDE responsiveness when LT needs to be initialized in the background to highlight a file


donnerpeter:
e.g. `new JLanguageTool(new GermanyGerman())` goes from 9 to 4 seconds for me

PR #4013, opened 2020-12-14
[en] improvements to messages and coverage for I_LOWERCASE

hftf:
The first commit standardizes the messages, and removes "Did you mean" from them.
The second commit generalizes two antipatterns into one antipattern for `where i is` (I’ve seen no false negatives in iWeb).
The third commit combines two rules that are basically duplicates.
In the last commit (ff66997), the antipatterns are quite zealous and over-applicable (so there may be some false negatives).


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -2909,8 +2909,28 @@ USA
                 <token spacebefore="no">i</token>
             </antipattern>
             <rule>
+                <!-- antipatterns intended for mathematical or scientific usages of the symbol "i" -->
                 <antipattern>
-                    <token regexp="yes">row|index</token>
+                    <token regexp="yes">(row|index|variable|quantit(y|ie)|number|term|symbol|constant|point|element|unit|value|size|equal)s?|sub|negative|minus|lowercase</token>
+                    <token>i</token>
```
udomai, line 2915: To make this AP a bit less zealous, maybe you could add

`<exception scope="next" postag="VB[PD]" postag_regexp="yes"/>`

That way, we'll still find mistakes like

_This is not the quantity i ordered._
_He didn't even try to understand the point i wanted to make._
_I'm going to show you how negative i can be._

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -2909,8 +2909,28 @@ USA
...
+                </antipattern>
+                <antipattern>
+                    <token>square</token>
+                    <token>of</token>
+                    <token>i</token>
```
udomai: I think the AP would be even better without "square". When is the PRP "I" ever preceeded by "of"?
The number i has of before it all the time:

_The distance of i to the origin is 1._

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -2909,8 +2909,28 @@ USA
...
+                </antipattern>
+                <antipattern>
+                    <token>equal</token>
+                    <token>to</token>
+                    <token>i</token>
```
udomai, line 2925: Same comment as above: does "to I" ever occur regularly? In a mathematical sense, it would appear more frequently, I guess:

_The distance from the origin to i is 1._
_In addition to i, quaternions also have j and k._

PR #4033, opened 2020-12-16
PatternToken rules: match only from likely tokens if possible

donnerpeter:
track token indices by lemma/token text in AnalyzedSentence

if a pattern has a token with token/lemma hints and fixed offset from the start,
use these indices to narrow down the search space

7-10% speedup in `Main -l en` for me

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/AnalyzedSentence.java
+++ b/languagetool-core/src/main/java/org/languagetool/AnalyzedSentence.java
@@ -19,7 +19,9 @@
 package org.languagetool;
 
 import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.*;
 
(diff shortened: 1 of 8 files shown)


donnerpeter:
@danielnaber ping? :)

danielnaber:
@donnerpeter I'll merge it after release (https://dev.languagetool.org/roadmap), we're in feature-freeze now so that only bug fixes are supposed to be added.

languagetool-core/src/main/java/org/languagetool/AnalyzedSentence.java
```diff
@@ -309,7 +322,29 @@ public String getAnnotations() {
...
    */
   public Set<String> getLemmaSet() {
-    return lemmaSet;
+    return lemmaOffsets.keySet();
+  }
+
+  /**
+   * @return all offsets in {@link #getTokensWithoutWhitespace()} where tokens with the given text occur (case-insensitive),
+   * or {@code null} if there are no such occurrences
+   * @since 5.2
```
donnerpeter, line 331: This (and below) probably should be changed to 5.3

danielnaber: Thanks, fixed

PR #4051, opened 2020-12-21
StringMatcher: honor backslashes when detecting finishing boundaries

donnerpeter:


donnerpeter:
Note that while I've run `mvn test`, it seems that Github haven't, although previously it used to

danielnaber:
> Note that while I've run mvn test, it seems that Github haven't, although previously it used to

Indeed, but the tests after the merge are okay, and they include the CircleCI run: https://github.com/languagetool-org/languagetool/commits/master

donnerpeter:
Sure, I'm just worried that this might be unintended and tests might be accidentally skipped in any other PR

Issue #3052, opened 2020-06-10
[en] Min/max does not work correctly with chunks


MikeUnwalla:
Here is my rule (based on a problem I found during the correction for  #3048.):

```
<rule id="MIN_MAX_TEST" name="min/max test">
    <antipattern>
        <token>are</token>
        <token chunk="B-NP-singular" postag="NNP"/>
        <token min="1" max="4" chunk="I-NP-singular"/>
        <token chunk="I-NP-singular">&amp;</token>
        <token chunk="I-NP-singular"/>
    </antipattern>
    <pattern>
        <marker>
            <token>are</token>
        </marker>
        <token postag="NNP"/>
    </pattern>
    <message>Min/max. Use <suggestion><match no="1" postag="VBZ"/></suggestion> with <match no="2"/>.</message>
    <short>Min/max test</short>
    <example correction="is">When <marker>are</marker> Mary going to Paris?</example>
    <example>When <marker>is</marker> Mary going to Paris?</example>
    <example type="triggers_error">How are Chris John Jones &amp; Jennifer Aniston important to this test?</example>
    <example type="triggers_error">How are DNA test sequencing &amp; DNA material working?</example>
</rule>
```

The antipattern should prevent the rule from firing, but it does not.

The Tagger Result shows that the chunks in the sentences that trigger an error match the chunks in the antipattern :
![image](https://user-images.githubusercontent.com/5816160/84257366-23193300-ab0d-11ea-8580-1b7b8bcf5f7f.png)

(The chunker is not correct, but that is a different problem.)

bebras123:
I had a similar issue https://github.com/languagetool-org/languagetool/issues/402  with `min, max` usage. It seems that `<token min="1" max="4" chunk="I-NP-singular"/>` is greedy and covers tokens `John Jones &amp; Jennifer`, so the next token `<token chunk="I-NP-singular">&amp;</token>` in the antipattern doesn't match. Using `<exception>&amp;</exception>` solves the problem: 
```xml
<antipattern>
    <token>are</token>
    <token chunk="B-NP-singular" postag="NNP"/>
    <token min="1" max="4" chunk="I-NP-singular"><exception>&amp;</exception></token>
    <token chunk="I-NP-singular">&amp;</token>
    <token chunk="I-NP-singular"/>
</antipattern>
```

Issue #402, opened 2016-05-03
Bug on token with max and min values


bebras123:
Hello, I recently found that my rule:

```
<rule id="IS_A_AN_PLURAL" name="IS_A_AN_PLURAL">
 <pattern>
  <token postag='VBZ'></token>
  <token regexp='yes' postag='DT' chunk='B-NP-plural'>a|an</token>
  <token min='0' max='3'></token>
  <marker>
  <token chunk='E-NP-plural'><exception postag='NN'></exception></token>
  </marker>
 </pattern>
 <message>Please check verb-subject agreement. Verb: "<match no="1"/>", subject: "<match no="4"/>"</message>
 <example correction=''>L is an extreme <marker>points</marker>.</example>
 <example>L is an extreme point.</example>
</rule>
```

doesn't match because of `<token min='0' max='3'></token>`. When i replace this with `<token></token>` it works. 

http://forum.languagetool.org/t/help-on-rule-with-optional-token/841

oriain:
Just cross-posting the link to the LanguageTool [forum post](http://forum.languagetool.org/t/help-on-rule-with-optional-token/841/5) on this issue.

f-knorr:
Could you try `skip="3"`instead?

```
<rule id="IS_A_AN_PLURAL" name="IS_A_AN_PLURAL">
 <pattern>
  <token postag='VBZ'></token>
  <token regexp='yes' postag='DT' chunk='B-NP-plural'>a|an</token>
  <token skip="3"></token>
  <marker>
  <token chunk='E-NP-plural'><exception postag='NN'></exception></token>
  </marker>
 </pattern>
 <message>Please check verb-subject agreement. Verb: "<match no="1"/>", subject: "<match no="4"/>"</message>
 <example correction=''>L is an extreme <marker>points</marker>.</example>
 <example>L is an extreme point.</example>
</rule>
```

arysin:
I think `<token min='0' max='3'></token>` is greedy so it also "grabs" the "points".
Try adding exception clause to it.

f-knorr:
I have closed this bug because there two suggestions to solve the problem, but there has been now feedback for more than 9 months.

PR #4071, opened 2020-12-28
[en] add rule MAKE_PRP_VBZ

hftf:
There are 5 commented-out example sentences that I'm not sure how to best deal with in the antipatterns. (It seems that the last antipattern should be excluding the first of those sentences, but it doesn't do so for some reason.)

I'm not sure how to make an optional token (`<token min="0">`) greedily match: for example, I want the VBZ `does` in `I can see it does alter`/`it does alters` to be skipped (matched by `<token min="0">does</token>`) so that the VBZ pattern in the marker can apply to `alter`/`alters`, but I still want to match `does` in other sentences (`he made me does it`).

The two comments in the `<marker>` are looser alternatives to the adjacent lines.


MikeUnwalla:
Known problems with min/max:
https://github.com/languagetool-org/languagetool/issues/402
https://github.com/languagetool-org/languagetool/issues/3052

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -15787,6 +15787,127 @@ USA
...
+        <rule id="MAKE_PRP_VBZ" name="makes me goes -> makes me go" default="temp_off">
+            <antipattern>
+                <token postag_regexp="yes" postag="SENT_START|PCT|CC|WP|PRP" />
+                <token min="0">as</token>
+                <token postag="PRP" />
+                <token min="0" max="2" postag_regexp="yes" postag="MD|RB" />
+                <token min="0" inflected="yes">do</token>
+                <token>see</token>
+                <token min="0" postag="PCT" />
+                <token>it</token><!-- subject or expletive -->
```
udomai, line 15799: In my experience, antipatterns don't handle multiple `min="0"` very well. They tend to only work when all optional tokens are present or absent. Maybe try `skip="1"` and an `exception scope="next" negate_pos="yes" postag="PRP"` for example.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -15787,6 +15787,127 @@ USA
...
+                <marker>
+                    <token postag="VBZ" case_sensitive="yes" regexp="yes">[a-z]+
+                    <!--<token postag="VBZ" regexp="yes">-->
+                        <exception postag="VB" />
+                        <!--<exception inflected="yes">be</exception>-->
+                        <exception regexp="yes">is|isn</exception>
+                    </token>
+                </marker>
+            </pattern>
+            <message>The bare infinitive <suggestion><match no="3" postag="VB" /></suggestion> should follow the verb '\1' and the object '\2'.</message>
```
udomai, line 15856: Not to be nit-picky, but doesn't this inverted syntax focus "follow"? Maybe a passive construction would be clearer...
"The verb ... and the object ... should be followed by the bare infinitive ... ."

PR #4112, opened 2021-01-08
[fr]expression à tout à l'heure

Rebecca-auque:
erreur d'àrthographe "à toute à l'heure"


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -20156,6 +20156,20 @@ though, and has since been slightly modified)-->
       <example correction="De toute façon"><marker>De toutes façons</marker></example>
       <example><marker>De toute façon</marker></example>
     </rule>
+      <rule id="A_TOUTE_A_L_HEURE" name="à tout à l'heure">
```
jaumeortola, line 20159: Always add `default="temp_off"` here, to test the rule before enabling it in production, even if you think it is completely safe because sometimes it is not.

udomai: 1) Jaume is right, I forgot to tell you about this.

2) It might be worth it making this a `rulegroup`. This rule would be the first `<rule>`, and then you could make a second `<rule>` for "tout(e) à l'heure" (without à).

_Je vais le lui dire tout**e** à l'heure._

(There might be false positives, if "toute" just happens to appear before phrases like "à l'heure qu'il est" or "à l'heure actuelle")

3) I've seen that you've put this at the end of the `category`. Could you put new rules on top, instead? Also, I've seen that you've put this in the category `CAT_REGIONALISMES`, I think it's rather `CAT_GRAMMAIRE`.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -20163,6 +20163,20 @@ though, and has since been slightly modified)-->
...
+          <pattern>
+              <token>à</token>
+              <marker>
+                  <token>toute</token>
+              </marker>
+              <token>à</token>
+              <token>l'</token>
+              <token>heure</token>
+          </pattern>
+          <message>Voulez-vous écrire <suggestion>tout</suggestion> ?</message>
```
udomai, line 20176: Generally, we tend to avoid simple "did you mean" messages, unless the error is really obvious. I don't think the error is super-ovious to everyone in this case.
"Dans l'expression 'tout à l'heure', 'tout' est écrit sans -e final.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -20163,6 +20163,20 @@ though, and has since been slightly modified)-->
       <example correction="De toute façon"><marker>De toutes façons</marker></example>
       <example><marker>De toute façon</marker></example>
     </rule>
+      <rule id="A_TOUTE_A_L_HEURE" name="à tout à l'heure">
```
udomai, line 20166: `default="temp_off"`

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -20251,6 +20280,20 @@ though, and has since been slightly modified)-->
...
+                  <token>toute</token>
+              </marker>
+              <token>à</token>
+              <token>l'</token>
+              <token>heure</token>
+          </pattern>
+          <message>Voulez-vous écrire <suggestion>tout</suggestion> ?</message>
+          <url>https://www.lalanguefrancaise.com/orthographe/a-tout-a-l-heure-ou-a-toute-a-l-heure/</url>
+          <example correction="tout">À <marker>toute</marker> à l'heure.</example>
+      </rule>
```
udomai, line 20296: With the last commit, this old version of the rule somehow found its way back into the file.

Before I merge this, can you please delete this part? It is identical with TOUTE_A_L_HEURE[1].

Also, I recommend not `temp_off`ing the single `<rule>`s, but the entire `<rulegroup>`.

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -8789,20 +8789,6 @@ though, and has since been slightly modified)-->
...
-                      <token>toute</token>
-                  </marker>
-                  <token>à</token>
-                  <token>l'</token>
-                  <token>heure</token>
-              </pattern>
-              <message>Voulez-vous écrire <suggestion>tout</suggestion> ?</message>
-              <url>https://www.lalanguefrancaise.com/orthographe/a-tout-a-l-heure-ou-a-toute-a-l-heure/</url>
-              <example correction="tout">À <marker>toute</marker> à l'heure.</example>
-          </rule>
```
udomai: I see that you removed this part (TOUTE_A_L_HEURE[1]) — why?

But the old rule (A_TOUTE_A_L_HEURE, lines 20269–20282) is still there. Was there a mix-up?

Rebecca-auque: Yes there was a mix-up on my part. I made the correct changes.

udomai (approved):

PR #4114, opened 2021-01-08
[fr]accord la Covid ou le Covid

Rebecca-auque:
La Covid est la détermination officielle.

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -728,7 +728,19 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_HOMONYMES_PARONYMES" name="Confusion d’homonymes et paronymes">
-    <rule id="GENTE_GENT" name="gente → gent">
+      <rule id="LE_COVID" name="la Covid et le Covid" default="temp_off">
+          <pattern>
+              <marker>
+                  <token>le</token>
+              </marker>
+              <token regexp="yes">Covid-19|Covid|covid|covid-19</token>
+          </pattern>
+          <message>Confusion probable entre le virus et <suggestion>la</suggestion> maladie.</message>
+          <url>http://www.academie-francaise.fr/le-covid-19-ou-la-covid-19</url>
+          <example correction="la">Il faut faire attention car <marker>le</marker> Covid-19 est récent.</example>
+          <example correction="La"><marker>Le</marker> Covid-19 est une maladie grave.</example>
+      </rule>
+      <rule id="GENTE_GENT" name="gente → gent">
       <pattern>
         <token>la</token>
         <marker><token>gente</token></marker>


dpelle:
Je pense plutôt que le mot "Covid" doit être épicène.
De nombreux journaux l'utilisent au masculin.

"Covid" est d’ailleurs épicène dans le dictionnaire dicollecte (ce qui est correct à mon avis) :
https://grammalecte.net/entry.php?prj=fr&id=235226

This PR can thus be rejected and shoul

udomai:
I don't think the matter is that simple. In France, the académie has considerable influence. Even if the linguistic reality is different, a user might appreciate the tip that some language perfectionists consider "le covid" to be wrong.

I would give that rule `type="style"` and `tags="picky"` and formulate the message accordingly, something along the lines of "Voulez-vous utiliser l'article recommandé ('la COVID') au lieu du masculin, qui est également correct ?" and maybe link to this site (une fois n'est pas coutume...): http://www.academie-francaise.fr/le-covid-19-ou-la-covid-19

udomai (commented):
I've left two comments in the respective places :)

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -728,7 +728,19 @@ though, and has since been slightly modified)-->
...
   <category id="CAT_HOMONYMES_PARONYMES" name="Confusion d’homonymes et paronymes">
-    <rule id="GENTE_GENT" name="gente → gent">
+      <rule id="LE_COVID" name="la Covid et le Covid" default="temp_off">
+          <pattern>
+              <marker>
+                  <token>le</token>
+              </marker>
+              <token regexp="yes">Covid-19|Covid|covid|covid-19</token>
+          </pattern>
+          <message>Confusion probable entre le virus et <suggestion>la</suggestion> maladie.</message>
```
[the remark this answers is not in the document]
udomai: Could you put `le` in single quotes ?

... `au détriment du masculin 'le' couramment utilisé ?`

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -8733,6 +8745,35 @@ though, and has since been slightly modified)-->
...
+                  </marker>
+                  <token>à</token>
+                  <token>l'</token>
+                  <token>heure</token>
+              </pattern>
+              <message>Voulez-vous écrire <suggestion>tout</suggestion> ?</message>
+              <example correction="tout">Je lui dirai <marker>toute</marker> à l'heure.</example>
+              <example correction="Tout"><marker>Toute</marker> à l'heure, j'irai au marché.</example>
+          </rule>
+      </rulegroup>
```
udomai, line 8776: This part is not supposed to be in this pull request, right?

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
udomai: Note so self: will solve this myself

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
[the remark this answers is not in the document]
udomai: Could you add two more subrules to this? We should also find "au Covid" and correct it to "à la Covid" as well as "du Covid" and correct it to "de la Covid".

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -728,7 +728,19 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_HOMONYMES_PARONYMES" name="Confusion d’homonymes et paronymes">
-    <rule id="GENTE_GENT" name="gente → gent">
+      <rule id="LE_COVID" name="la Covid et le Covid" type="style" tags="picky" default="temp_off">
+          <pattern>
+              <marker>
+                  <token>le</token>
+              </marker>
+              <token regexp="yes">Covid-19|Covid|covid|covid-19</token>
```
udomai, line 736: Unless you tell it to be case-sensitive, the string will be case-insensitive. This means you can simply write
`<token regexp="yes">covid(-19)?</token>`, thus covering all of the matches you wrote above. Plus the all-caps version COVID.

udomai (changes requested):

PR #4117, opened 2021-01-08
[en] new rule IS_IS

hftf:
Doesn't catch the following example at the moment: <code>What you can do **is is** just put them</code>


tiff:
Hi @hftf please check my rule IS_VBZ, this might find the same/similar mistakes. Maybe they can be merged.

udomai:
Please check whether this rule can be added to IS_VBZ. In the meantime, I need to find out why the test keeps failing.

PR #4126, opened 2021-01-12
[fr] rule superlatif

Rebecca-auque:
superlatif plus+mieux et plus+bien


udomai:
Idea for further differentiating this rule:

_les sites les plus mieux de Mallorca_ → les meilleurs sites de Mallorca

This would require an AP for the first rule (`les [N s?p .] les plus mieux`) and a subrule that transforms this structure (→ `les meilleurs [ N s?p .]`).

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -207,6 +207,42 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
 (the rule on "ajouter en plus"
 was already there before the category was created,
 though, and has since been slightly modified)-->
+          <rulegroup id="PLUS_MIEUX" name="plus mieux" default="temp_off">
+          <rule>
+              <pattern>
+                  <token>plus</token>
+                  <token regexp="yes">mieux?<exception postag="J.*" postag_regexp="yes" scope="next"/></token>
+              </pattern>
+              <message>« Plus mieux » est une collocation inadéquate. Voulez-vous écrire <suggestion>bien mieux</suggestion> ou   <suggestion>meilleur</suggestion> ? </message>
```
udomai, line 216: There are three spaces after "ou". Also, the messages of the three subrules are not the same, two say "voulez-vous" and one says "souhaitez-vous". Could you unify those?

PR #4128, opened 2021-01-12
[fr] plus mauvais

Rebecca-auque:
faute de collocation


udomai:
Question: "Le diesel sent **plus mauvais** que l'essence, mais il pollue moins" → "sent pire" doesn't seem right

"Il a obtenu des plus mauvais résultats" → maybe "moins bons" or "inférieurs" would be also good suggestions?

The message needs to be nicer and more helpful.

PR #4129, opened 2021-01-12
[fr] la ou les plus

Rebecca-auque:
Faute de grammaire

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -8796,7 +8796,7 @@ though, and has since been slightly modified)-->
               </marker>
               <token regexp="yes">plus|moins</token>
               <token min="0" max="2" postag="A.*" postag_regexp="yes"></token>
-              <token postag="SENT_END" postag_regexp="yes">.</token>
+              <token postag="SENT_END" regexp="yes">[\.\?!]</token>
           </pattern>
           <message> Il semble qu’il y ait une faute d'article. Nous vous suggérons d'employer l'article <suggestion>le</suggestion>.</message>
           <example correction="le">C'est Julie qui court <marker>la</marker> plus vite.</example>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -8788,6 +8788,22 @@ though, and has since been slightly modified)-->
...
   <category id="CAT_GRAMMAIRE" name="Grammaire">
+      <rule id="LA_OU_LES_PLUS" name="la ou les plus" default="temp_off">
+          <pattern>
+              <token postag="V.*" postag_regexp="yes"></token>
+              <marker>
+                  <token regexp="yes">la|les</token>
+              </marker>
+              <token regexp="yes">plus|moins</token>
+              <token min="0" max="2" postag="A.*" postag_regexp="yes"></token>
+              <token>.</token>
```
udomai, line 8799: `<token postag="SENT_END" />`

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -8788,6 +8788,22 @@ though, and has since been slightly modified)-->
...
+          </pattern>
+          <message>« la|les» est une expression fautive.</message>
```
udomai, line 8801: `<message>Il semble qu'il ait une faute dans l'article '\2'... </message>`

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
udomai: If your postag is just `SENT_END`, there's no need for `postag_regexp="yes"`. You can include other punctuation at the end of the sentence: `<token postag="SENT_END" regexp="yes">[\.\?\!]</token>`.

Note that the `.` and the `?` need to be "escaped" with a `\` so the system knows you mean an actual period and an actual question mark (and not the regex elements `.` and `?`)

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
udomai: Please do not put quotation marks around the `<suggestion>` — quotation marks are added automatically to that element.

udomai (changes requested):

udomai:
Merged manually ab86c33026598d21e5a3ab7f90ac323c1bd1fa87

udomai:
Manually merged 9fa32638dfdd520aac1a2cdb657141b0915037b2

PR #4150, opened 2021-01-14
[en] improve GONNA and DONTCHA

hftf:


udomai:
As discussed, this is a pretty busy rule. It is best to make a duplicate to test the changes temp_off. Alternatively, you can run the main test on the nightly corpus on your local machine.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -86541,14 +86541,17 @@ The accident victim died from her injuries.
             <rule>
                 <pattern>
                     <marker>
-                        <token>gonna</token>
+                        <token regexp="yes">g[ou]n?na|finna</token>
```
tiff, line 86544: "Gunna" is also a rapper:
https://en.wikipedia.org/wiki/Gunna_(rapper)

udomai: How about an AP like this (to exclude case sensitive "Gunna" except at SENT_START):

```xml
<antipattern>
    <token negate_pos="yes" postag="SENT_START"/>
    <token case_sensitive="yes">Gunna</token>
</antipattern>
```

udomai: _These allegations concerning Gunna have no basis in reality._

PR #4161, opened 2021-01-15
ra-11

Rebecca-auque:
Collocation fautive

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -9800,7 +9800,7 @@ though, and has since been slightly modified)-->
               <token regexp="yes">de|d'</token>
               <token regexp="yes">perdre|échouer</token>
           </pattern>
-          <message>Faute de style possible car cette expression peut paraître péjorative.
+          <message>Une faute de style est possible car cette expression peut paraître péjorative.
               Nous vous suggérons  d’employer <suggestion>risquer</suggestion>.</message>
           <example correction="risquer"><marker>avoir des chances</marker> de perdre</example>
           <example correction="risquer"><marker>avoir des chances</marker> d'échouer</example>
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -9779,6 +9779,33 @@ though, and has since been slightly modified)-->
     </rulegroup>
   </category>
   <category id="CAT_GRAMMAIRE" name="Grammaire">
+    <rulegroup id="RISQUE_ET_CHANCE" name="risque et chance">
+      <rule>
+          <pattern>
+              <marker><token inflected="yes">risquer</token></marker>
+              <token regexp="yes">de</token>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -9779,6 +9779,33 @@ though, and has since been slightly modified)-->
...
   <category id="CAT_GRAMMAIRE" name="Grammaire">
+    <rulegroup id="RISQUE_ET_CHANCE" name="risque et chance">
+      <rule>
+          <pattern>
+              <marker><token inflected="yes">risquer</token></marker>
+              <token regexp="yes">de</token>
+              <token regexp="yes">gagner</token>
+          </pattern>
+          <message>Faute de style possible car cette expression peut paraître péjorative.
+              Nous vous suggérons <suggestion>avoir des chances</suggestion>.</message>
```
udomai, line 9790: No line breaks allowed in `<message>`s :)

Also, it would be best if we proposed the verb "avoir" in the same form the verb "risquer" is in the original. Otherwise, if the user writes "Il risque de gagner", our correction will result in "Il avoir des chances de ganger" (unacceptable).

Unfortunately, this is a little complicated due to "avoir" having "avoir" in the postag, contrary to "normal" verbs. This is how it can be done:

```xml
      <rule>
          <pattern>
              <marker>
                <token postag="V (ind|sub).*" postag_regexp="yes" inflected="yes">risquer</token>
              </marker>
                <token regexp="yes">de</token>
                <token regexp="yes">gagner</token>
          </pattern>
          <message>Faute de style possible car cette expression peut paraître péjorative.</message>
          <suggestion><match no="1" postag="(V )(ind|sub)(.*)" postag_regexp="yes" postag_replace="(V avoir ind$3)|(V avoir sub$3)">avoir</match> des chances</suggestion>
          <example correction="a des chances|ait des chances">Il <marker>risque</marker> de gagner.</example>
      </rule>
```

About formatting: please notice that we normally don't put `<marker><token/></marker>` in one line, but in three separate lines.

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -9779,6 +9779,35 @@ though, and has since been slightly modified)-->
...
+                  <token regexp="yes">des</token>
+                  <token regexp="yes">chances</token>
+              </marker>
+              <token regexp="yes">de|d'</token>
+              <token regexp="yes">perdre|échouer</token>
+          </pattern>
+          <message>Une faute de style est possible car cette expression peut paraître péjorative.
+              Nous vous suggérons  d’employer <suggestion>risquer</suggestion>.</message>
+          <example correction="risquer"><marker>avoir des chances</marker> de perdre</example>
+          <example correction="risquer"><marker>avoir des chances</marker> d'échouer</example>
```
udomai, line 9808: - Pas d'alinéa dans le <message>
- Transfer the verb form from avoir to risquer

udomai:
Manually merged here: 26c95b3d8348d985016bc55787fb1a26a7ab9e1f

I put it into the CAT_TOURS_CRITIQUES. Nice solution for the verb form of _risquer_!

PR #4162, opened 2021-01-15
ra-10

Rebecca-auque:
Erreur d'orthographe

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -35067,7 +35067,7 @@ though, and has since been slightly modified)-->
           <pattern>
               <token inflected="yes">rénumérer</token>
           </pattern>
-          <message>Une erreur d’orthographe semble être détectée. Souhaitez-vous écrire « <suggestion>rémunérer</suggestion> » ?</message>
+          <message>Une erreur d’orthographe semble être détectée. Voulez-vous écrire « <suggestion>rémunérer</suggestion> » ?</message>
           <suggestion>rémunérer</suggestion>
           <example correction="rémunérer">Il faut <marker>rénumérer</marker> les ouvriers pour leur travail.</example>
       </rule>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -35063,7 +35063,15 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_TOURS_CRITIQUES" name="Tours critiqués divers (barbarismes, impropriétés, solécismes, etc.)">
-    <rulegroup id="POUR_PAS_QUE" name="familier : pour (ne) pas que...">
+      <rule id= "RENUMERER_OU_REMUNERER" name="rénumérer ou rémunérer">
+          <pattern>
+              <token inflected="yes">rénumérer</token>
```
udomai, line 35068: The thing is: our dictionary doesn't know about the very rare verb "rénumérer". So it cannot "inflect" it. You can either put the expected verb forms in there (with plenty of regex) or — which may be easier — we can improve the suggestions given for any typo starting with `rénumér.*`. Remind me about this on Monday :)

udomai:
Good point, I'll look into the technical details :)

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
udomai: Here is a solution (powered by Jaume): make the token `<token regexp="yes">rénum.*</token>`. Le message devient `Une erreur d’orthographe semble être détectée.`. La suggestion sera `<suggestion suppress_misspelled="yes"><match no="1" regexp_match="(?iu)num" regexp_replace="mun" case_conversion="preserve"/></suggestion>`.

Please check whether the rule, in this form, works okay and doesn't produce false positives.

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
udomai: We will discuss this on Monday :)

udomai:
Thanks to your idea, Jaume has created a more general rule. See 6bf1fa5ae85ef4bc7d11b5289419ba8cfd1d58e8

PR #4198, opened 2021-01-20
[fr] se demander qu'est-ce que

Rebecca-auque:
se demander ce que

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -35472,16 +35472,14 @@ though, and has since been slightly modified)-->
               <token regexp="yes">qu(&apostrophe;)</token>
               <token>est</token>
               <token>-ce</token>
-              <token regexp="yes" skip="-1">qu.*</token>
+              <token regexp="yes" skip="-1">qu[ie]?|qu(&apostrophe;)</token>
               <token postag="M fin inte"/>
           </antipattern>
           <pattern>
-              <marker>
                   <token regexp="yes">qu(&apostrophe;)</token>
                   <token>est</token>
                   <token>-ce</token>
                   <token regexp="yes">qu(&apostrophe;)|que</token>
-              </marker>
           </pattern>
           <message>Dans le langage courant, cette question indirecte devrait être introduite par <suggestion>ce \4</suggestion>.</message>
           <example correction="ce qu'"> Je me demande <marker>qu'est-ce qu'</marker>il pense.</example>
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -35467,6 +35467,21 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_TOURS_CRITIQUES" name="Tours critiqués divers (barbarismes, impropriétés, solécismes, etc.)">
+      <rule id="SE_DEMANDER_QU_EST_CE_QUE" name="se demander qu’est-ce que" type="style" default="temp_off">
+          <pattern>
+              <token postag="R pers obj.*" postag_regexp="yes"/>
+              <token inflected="yes">demander</token>
+              <marker>
+                  <token>qu'</token>
+                  <token>est</token>
+                  <token>-ce</token>
+                  <token regexp="yes">qu.*</token>
+              </marker>
+          </pattern>
+          <message>Cette expression est familière. Souhaitez-vous écrire les termes <suggestion>ce que</suggestion> employés dans le language courant.</message>
+          <example correction="ce que"> Je me demande <marker>qu'est-ce qu'</marker>il pense.</example>
+          <example correction="ce que"> Il faut se demander <marker>qu'est-ce que</marker> tu feras par la suite.</example>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -35467,6 +35467,21 @@ though, and has since been slightly modified)-->
...
+              <token postag="R pers obj.*" postag_regexp="yes"/>
+              <token inflected="yes">demander</token>
+              <marker>
+                  <token>qu'</token>
+                  <token>est</token>
+                  <token>-ce</token>
+                  <token regexp="yes">qu.*</token>
+              </marker>
+          </pattern>
+          <message>Cette expression est familière. Souhaitez-vous écrire les termes <suggestion>ce que</suggestion> employés dans le language courant.</message>
```
udomai, line 35481: Nice rule! We should think about how to broaden it, I think it could catch more cases.

In the message, we should explain: "Dans le langage standard, cette question indirecte devrait être introduite par 'ce que'." ...or something like that.

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -35467,6 +35467,19 @@ though, and has since been slightly modified)-->
...
   <category id="CAT_TOURS_CRITIQUES" name="Tours critiqués divers (barbarismes, impropriétés, solécismes, etc.)">
+      <rule id="QU_EST_CE_QUE" name="qu est-ce que" default="temp_off">
+          <pattern>
+              <marker>
+                  <token case_sensitive="yes">qu'</token>
+                  <token>est</token>
+                  <token>-ce</token>
+                  <token regexp="yes">qu.*</token>
+              </marker>
+          </pattern>
```
udomai, line 35478: This pattern now matches any occurrence of _qu'est-ce que/qu'_.

This will cause FPs for:

- Sinon, qu'est-ce qu'on va manger aujourd'hui ?
- Paul, qu'est-ce que tu as encore fait ?

This rule is now too general. You'll have to avoid going just for any "qu'est-ce que" and define some kind of context.

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -35467,6 +35467,26 @@ though, and has since been slightly modified)-->
...
+          </antipattern>
+          <pattern>
+              <marker>
+                  <token case_sensitive="yes">qu'</token>
+                  <token>est</token>
+                  <token>-ce</token>
+                  <token regexp="yes">qu.*</token>
+              </marker>
+          </pattern>
+          <message>Dans le langage courant, cette question indirecte devrait être introduite par <suggestion>ce que</suggestion>.</message>
```
udomai, line 35486: We're going to be in trouble with the spaces here. I only see the solution to put another `<token/>` within the `<marker>` (i.e. a fifth token that can be anything) and splitting the rule.
One subrule for the case that \4 is `qu'`, in that case the suggestion is `ce \4\5`.
The second subrule for the case that \4 is `que`, in that case the suggestion is `ce \4 \5`.
Dans l'état actuel, ton premier exemple se terminerait en "... ce queil pense".

Also: if the `<marker>`s conver the entire `<pattern>`, you don't need to write `<marker>`. If you don't put any markers, the entire pattern is marked by default.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -35467,6 +35467,26 @@ though, and has since been slightly modified)-->
...
+          <example correction="ce que">Je me demande <marker>qu'est-ce qu'</marker>il pense.</example>
+          <example correction="ce que">Dis-moi <marker>qu'est-ce que</marker> tu en pense.</example>
```
udomai, line 35488: typo: penseS

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -35467,6 +35467,26 @@ though, and has since been slightly modified)-->
...
+              <token regexp="yes" skip="-1">qu.*</token>
+              <token postag="M fin inte"/>
+          </antipattern>
+          <pattern>
+              <marker>
+                  <token regexp="yes">qu(&apostrophe;)</token>
+                  <token>est</token>
+                  <token>-ce</token>
+                  <token regexp="yes">qu(&apostrophe;)|que</token>
+              </marker>
```
udomai, line 35484: The spacing problem seems to be resolved (or there never was one). But the unnecessary markers are still there.

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: I made the necessary changes so the rule is done.

udomai (approved):

udomai:
manually merged: 5e238bc4e35276175ba3d6c22395098f789ba8e2

PR #4267, opened 2021-02-01
[fr] part ou par rule

Rebecca-auque:
part ou par

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -1049,7 +1049,52 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_HOMONYMES_PARONYMES" name="Confusion d’homonymes et paronymes">
-    <rule id="GENTE_GENT" name="gente → gent">
+      <rulegroup id="PART_OU_PAR" name="part ou par" default="temp_off">
+          <rule>
+              <antipattern>
+                  <token>part</token>
+                  <token postag="P.*" postag_regexp="yes"/>
+              </antipattern>
+              <antipattern>
+                  <token>de</token>
+                  <token>part</token>
+                  <token>et</token>
+                  <token min="0">d'</token>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -1049,7 +1049,52 @@ though, and has since been slightly modified)-->
...
+                  <marker>
+                      <token>part</token>
+                  </marker>
+                  <token postag="D.*" postag_regexp="yes"/>
+              </pattern>
+              <message>Une erreur d’orthographe semble être détectée.</message>
+              <suggestion>par</suggestion>
+              <url>https://www.lalanguefrancaise.com/orthographe/de-par-ou-de-part-orthographe/</url>
+              <example correction="par">Il a réussi de <marker>part</marker> sa pugnacité.</example>
+          </rule>
```
udomai, line 1082: The example of the first rule is already found by another rule.
Check out whether your rule is better or worse than that rule:
1) see whether your rule finds all the `<examples>` of that rule
2) `grep -i -w -E "(de part des|[mts](on|a|es)|l(a|es?|'))"` (the `D.*` part of the pattern you'll have to express using regex, this is just an example off the top of my head) in the French corpus and see which rule finds more errors
3) If you see that your rule finds errors the other rule doesn't find, try and "merge" them (e.g. extend the pattern of the old rule or create a new subrule)

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: I checked the rule that already existed. I have created two new rules for this case therefore the rule previously done becomes void. Could it be taken out ?

udomai (approved):

udomai:
Bon boulot, après le diff, on peut supprimer l'ancienne. Manually merged: cc755da5d0d773d27345d22ef2f981b6b6d8929f

PR #4269, opened 2021-02-01
[fr] feus ou feux rule

Rebecca-auque:
feus ou feux


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -1049,6 +1049,22 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_HOMONYMES_PARONYMES" name="Confusion d’homonymes et paronymes">
+      <rule id="FEUS_OU_FEUX" name="feus ou feux">
+          <antipattern>
+              <token regexp="yes">feu|feus</token>
```
udomai, line 1054: It's not wrong, it just doesn't look good ;)

→ `feus?`

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -1049,6 +1049,22 @@ though, and has since been slightly modified)-->
...
+              <token postag="[NZ].*" postag_regexp="yes"/>
```
udomai, line 1055: I'd also put the `UNKNOWN` tag and tokens that are all-caps (those cases are rare and FPs are probable), see rule editor results

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: I have put the UNKNOWN tag but I didn't put the all-caps as I didn't see FPs in the rule editor. It would also exclude sentences like "Il a des feux AR." and "Il a des feux AV." which are often used to refer to "avant" and "arrière". I think the rule is ready.

udomai:
manually merged: 9ec3b6678e23c45006c9a0fdc955998c3a1c4d1e

PR #4281, opened 2021-02-03
[en] add typo rule FAIR_SURE

hftf:


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -1082,6 +1082,26 @@ USA
...
+            </antipattern>
+            <pattern>
+                <marker>
+                    <token>fair
+                        <exception case_sensitive="yes">Fair</exception>
+                    </token>
+                    <token>sure</token>
+                </marker>
+            </pattern>
+            <message>Did you mean <suggestion>fairly sure</suggestion> or <suggestion>fair share</suggestion> or <suggestion>for sure</suggestion>?</message>
```
udomai, line 1099: I will merge it now, but before it is activated, could you make the following change?
The message contains three suggestions which will be repeated directly below it (as options the user can click on). This is considered somewhat redundant. It's better to just say "The words '\1' and '\2' don't seem to fit together in this context." and let the user choose from the suggestions.

Also, you might want to add a fourth suggestion for cases like this:

_This isn't **fair sure** but that's just how it is._ → fair, sure,

udomai:
I left a comment in there, just mentioning it once more here in case you're not notified after I merged this.

PR #4298, opened 2021-02-04
[fr] adj f et adj f rule

Rebecca-auque:
Accord adj f et adj f.

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -10084,6 +10084,68 @@ though, and has since been slightly modified)-->
     </rulegroup>
   </category>
   <category id="CAT_GRAMMAIRE" name="Grammaire">
+      <rule id="ADJ_F_ET_ADJ_F" name="adj f et adj f">
+          <antipattern>
+              <token postag="N f s"/>
+              <token postag="J f s"/>
+              <token>et</token>
+              <marker>
+                  <token postag="V ppa m s"/>
+              </marker>
+              <token postag="V.*" postag_regexp="yes"/>
+          </antipattern>
+          <antipattern>
+              <token postag="N f s"/>
+              <token postag="J f s"/>
+              <token>et</token>
+              <marker>
+                  <token postag="J m s"/>
+              </marker>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -10084,6 +10084,68 @@ though, and has since been slightly modified)-->
...
+          <pattern>
+              <token postag="N f s"/>
+              <token postag="J f s"/>
+              <token>et</token>
+              <marker>
+                  <token postag="J m s"/>
+              </marker>
+              <token postag="V.*" postag_regexp="yes"/>
+          </pattern>
+          <message>Faux</message>
```
udomai, line 10142: Faux ! (Tu connais la chaîne de Norman ?)

This message wasn't ready yet, apparently.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -10084,6 +10084,68 @@ though, and has since been slightly modified)-->
     </rulegroup>
   </category>
   <category id="CAT_GRAMMAIRE" name="Grammaire">
+      <rule id="ADJ_F_ET_ADJ_F" name="adj f et adj f">
```
udomai, line 10087: This is a nice rule because it covers a basic error. Using unification and the AdvancedSynthesizerFilter, you can cover all combinations of masculine/feminine as well as singular/plural at once.

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: Indeed, unification and ASF are very useful ! Besides the adjectives interfere with the verb after them as in "Des personnes grandes et intelligentes arrive vers nous." I think it could be look into.

udomai:
This rule isn't ready yet (at least one message still isn't fixed). Could you then run it by the SentenceSourceChecker (on the nightly diff corpus) and confirm when the rule is ready?

udomai:
Replaced by https://github.com/languagetool-org/languagetool/pull/4449

PR #4404, opened 2021-02-18
[fr] adj avant ou arrière rule

Rebecca-auque:
accord adj avant et arrière

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -10087,14 +10087,14 @@ though, and has since been slightly modified)-->
   <category id="CAT_GRAMMAIRE" name="Grammaire">
       <rule id="ADJ_AVANT_OU_ARRIERE" name="adj avant ou arrière" default="temp_off">
       <pattern>
-             <token postag="N.*" postag_regexp="yes"></token>
+             <token postag="N.*" postag_regexp="yes"/>
     <marker>
         <token regexp="yes">avants|arrières</token>
     </marker>
 </pattern>
 <message>Une erreur d’orthographe semble être détectée.</message>
 <suggestion suppress_misspelled="yes"><match no="2" regexp_match="(?iu)s" regexp_replace=""/></suggestion>
-<example correction="avant">Les roues <marker>avants</marker> de la voiture.</example>
+<example correction="avant">Il regarde les roues <marker>avants</marker> de la voiture.</example>
         </rule>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -10085,6 +10085,17 @@ though, and has since been slightly modified)-->
...
   </category>
   <category id="CAT_GRAMMAIRE" name="Grammaire">
+      <rule id="ADJ_AVANT_OU_ARRIERE" name="adj avant ou arrière" default="temp_off">
+      <pattern>
+             <token postag="N.*" postag_regexp="yes"/>
+    <marker>
+        <token regexp="yes">avants|arrières</token>
+    </marker>
+</pattern>
+<message>Une erreur d’orthographe semble être détectée.</message>
```
udomai, line 10095: Could you make the message a little more explanatory? Something like "Quand 'arrière' est utilisé en tant qu'adjectif, il est invariable.".

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -10085,6 +10085,17 @@ though, and has since been slightly modified)-->
     </rulegroup>
   </category>
   <category id="CAT_GRAMMAIRE" name="Grammaire">
+      <rule id="ADJ_AVANT_OU_ARRIERE" name="adj avant ou arrière" default="temp_off">
+      <pattern>
```
udomai, line 10089: Can you confirm that none of the examples found by the rule editor are false positives?
Especially "lignes arrières" seems to be very popular.
Let's put a `<url>` in there, too. How about https://www.larousse.fr/dictionnaires/francais/arri%c3%a8re/5400 ? It says there "L'adjectif est invariable".

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: That's done.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: It isn't a FP as it means "les lignes qui sont à l'arrière (du terrain...)". I put another url taking into account both "avant" and "arrière" and explaining the rule.

udomai:
merged here: 9799f2771bde3d9d6990dc0679cc9b82240ea8cf

PR #4449, opened 2021-02-25
[fr] adj f et adj f rule improved

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -10164,6 +10164,68 @@ though, and has since been slightly modified)-->
     </rulegroup>
   </category>
   <category id="CAT_GRAMMAIRE" name="Grammaire">
+      <rule id="ADJ_F_ET_F" name="adj f et f" default="temp_off">
```
udomai, line 10167: FP trouvés dans le rule editor :

> Marius Vazeilles, Le Pays d'Ussel : préhistoire et **histoire ancienne et Manuel** pratique d'archéologie régionale, Tulle, Impr.

> La **munition explosive et antipersonnel** pèse , avec une charge explosive (TNT) de .

Have you considered making the second token `<token postag="J [ef] sp?" postag_regexp="yes"/>` ? That will require a lot more APs and exceptions, but it might find a lot more errors.

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: Indeed, the rule finds more errors. I improved it and added APs so I think it is ready.

udomai: I found a tricky FP:

> Elle est membre de la commission finances, administration générale et **personnel** et de la commission citoyenneté et démocratie locale.

This is about "la commission (affectée au) personnel". Maybe write `(anti)?personnel` in the exception in line 10256?

Also, this rule doesn't find any errors in the nightly diff corpus. I was wondering whether some AP or exception was too broad, since these errors should occur more often. Could you double-check the APs?

I've already found this: your last AP makes the `sp?` in your third token obsolete:

```xml

          <antipattern>
              <token postag="J f s"/>
              <token>et</token>
              <token postag="V ppa [me] sp?|Z.*|J m sp" postag_regexp="yes"></token>
          </antipattern>

udomai (changes requested):

PR #4453, opened 2021-02-26
[fr] a ou en moyen de transport rule

Rebecca-auque:
Confusion entre les prépositions à et en.

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -37493,6 +37493,57 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_TOURS_CRITIQUES" name="Tours critiqués divers (barbarismes, impropriétés, solécismes, etc.)">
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -37493,6 +37493,57 @@ though, and has since been slightly modified)-->
...
+              <token>en</token>
+              <token regexp="yes">vélo|byciclette|moto|cheval|skis</token>
+              <token regexp="yes">tout|à</token>
+              <token regexp="yes">chemin|terrain|cailloux</token>
+          </antipattern>
+          <pattern>
+              <marker>
+                  <token>en</token>
+              </marker>
+              <token regexp="yes">vélo|byciclette|moto|cheval|skis</token>
```
udomai, line 37537: The rule editor results gave me some ideas: Does this rule work also for `snowboard|skateboard|VTT|raquettes|crampons` or some of those?

Do the many rule editor results all look okay to you?

Rebecca-auque: It would have been great if it worked for these cases too as well as "tandem, tricycle, surf..." but for some reason, it doesn't work. 
The last time I checked, all the results were okay.

PR #4503, opened 2021-03-02
[fr] confusion est et

Rebecca-auque:
Confusion entre le verbe est et la conjonction et.

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -37502,6 +37502,73 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_TOURS_CRITIQUES" name="Tours critiqués divers (barbarismes, impropriétés, solécismes, etc.)">
+      <rule id="CONFUSION_EST_ET" name="confusion est et" default="temp_off">
+          <antipattern>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -37502,6 +37502,73 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_TOURS_CRITIQUES" name="Tours critiqués divers (barbarismes, impropriétés, solécismes, etc.)">
+      <rule id="CONFUSION_EST_ET" name="confusion est et" default="temp_off">
```
udomai, line 37505: Nice! I can't believe we don't find this error yet.

There is an FP in the rule editor, though:

> Pourquoi elle **et** pas son homonyme géorgienne ?

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: I added the exception and checked the rule editor for more FPs.

udomai (approved):

udomai:
manually merged: 2b3941a773941496aa2e37dd3f3620fcb746b7ea

PR #4545, opened 2021-03-04
confusion quel

Rebecca-auque:
quel que

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -10236,7 +10236,7 @@ though, and has since been slightly modified)-->
                   <token min="0">en</token>
                   <token>soient</token>
                   <token postag="D [me] p" postag_regexp="yes"/>
-                  <token postag="N m p"/>
+                  <token postag="N m p"><exception postag="N f p"/></token>
               </pattern>
               <message>Une erreur semble être détectée dans cette phrase. Souhaitez-vous écrire <suggestion>quels</suggestion> ?</message>
               <example correction="quels">Ils les supprimeront <marker>qu'elle</marker> que soient les noms.</example>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -10181,6 +10181,68 @@ though, and has since been slightly modified)-->
...
+                  <token min="0">en</token>
+                  <token>soit</token>
+                  <token postag="D [me] s" postag_regexp="yes"/>
+                  <token postag="N m s"/>
+              </pattern>
+              <message>Une erreur semble être détectée dans cette phrase. Souhaitez-vous écrire <suggestion>quel</suggestion> ?</message>
+              <example correction="quel">Il le fera <marker>qu'elle</marker> que soit son statut.</example>
+              <example correction="quel">Il le fera <marker>quelle</marker> qu'en soit son problème.</example>
+          </rule>
+          <rule>
```
udomai, line 10229: FP in the rule editor:

> Quelles que soient les teneurs en uranium des milieux, …

(When it means "level (of content)", it's "la teneur", n'est-ce pas ? The word is correctly labelled as both `N m s` and `N f s`, so a simple exception for that in your last token will do. Logically, it should make sense in the first subrule as well.

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: I added the exception to this subrule. I think that in the other subrule, the m/f is clearly stated so an exception is not needed.

udomai:
This commit came with an older version of the grammar.xml. Merged only new rule here: c6db0f1a411bd020b1c0751b74f3cdbc5f17aaec

PR #4549, opened 2021-03-05
[fr] être allé à draft rule

Rebecca-auque:


udomai:
Do you still consider this a draft?
I've looked at it myself and I found this (FP because of wrong apostrophe?):

> Cette famille a été à l´origine de deux régiments d´Harambure.

Are you sure about this rule?

"J'ai été à Paris pendant 3 jours" will be converted to "Je suis allée à Paris pendant 3 jours"

Also, when the verb form is "je suis", we don't want `allée?s`. How about you take the singular/plural info from \1?

```xml
          <suggestion><match no="1" postag="(V avoir) (ind) (.*)" postag_regexp="yes" postag_replace="(V etre) $2 $3">être</match> <match no="1" postag="(V avoir) (.* [123]) ([sp])" postag_regexp="yes" postag_replace="V ppa . $3">aller</match></suggestion>
          <example correction="suis allé|suis allée">J'<marker>ai été</marker> à la boulangerie.</example>
```

Rebecca-auque:
Yes, I'm sure about this rule, it is a common error in French. "J'ai été à Paris pendant 3 jours" is exactly the type of example it should modified. It is a draft because the suggestion isn't specific enough. For example "Elles ont été à la boulangerie" becomes "Elles sont allées/allés à la boulangerie".  It is then catched by another rule to become "allées". This isn't fine, is it?

udomai:
Good question… in principle, this can be accepted if a precise solution isn't practical.

In this case, I think it would be very complex to adapt the rule to the different contexts.

_Mon mari que j'ai connu grâce à elle a été à la boulangerie._ → this is one of the harder cases where the rule would have to correctly guess _est all**é**_ with no extra _e_.

We can talk about it on Thursday, we might accept this as long as the correct choice is among the offered suggestions.

udomai:
Manually merged: 1609bb84cefb3492f46116b5db80f607fd96fd29

PR #4551, opened 2021-03-05
[fr] barbarismes question rule

Rebecca-auque:
Double question et registre familier.

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -37657,8 +37657,7 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_TOURS_CRITIQUES" name="Tours critiqués divers (barbarismes, impropriétés, solécismes, etc.)">
-      <rulegroup id="BARBARISMES_QUESTION" name="barbarismes question" default="temp_off">
-        <rule>
+      <rule id="BARBARISME_QUESTION" name="barbarisme question" default="temp_off">
           <pattern>
               <token>est</token>
               <token>-ce</token>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -37657,6 +37657,39 @@ though, and has since been slightly modified)-->
...
+      <rule>
+          <pattern>
+              <marker>
+                  <token>qu'</token>
+                  <token>est</token>
+                  <token>-ce</token>
+                  <token regexp="yes">qu'|que</token>
+                  <token postag="R pers suj.*" postag_regexp="yes"/>
+                  <token min="0" max="1" postag="R pers obj.*" postag_regexp="yes"/>
+                  <token postag="V.*" postag_regexp="yes"/>
```
udomai, line 37685: The rule would probably work with skip="-1" in this penultimate token, right? That way, you'll also catch "Qu'est-ce que tu fais dans ma maison ?"

Making `inte` optional in the last token is dangerous. This risks an overlap with your "Qu'est-ce qu'il est beau"-rule (if you add skip="-1").

If you write "Qu'est-ce que je pense ?", your rule will produce "Que pense-je", which will then be corrected to "Que pensé-je" by another rule.
If you write "Qu'est-ce qu'il ajoute ?", your rule will produce "Que ajoute-il ?", which will be corrected to "Que ajoute-t-il ?", then to "Qu'ajoute-t-il ?".
We could accept this. But I think two extra corrections is a little much. We could split up this rule and, if the verb starts with vowel or h (except H_ASPIRE), put `qu'` in the suggestion.
Since this rule is pretty picky (because "Qu'est-ce que" questions are pretty common even in common speech), one could even think about splitting those two subrules into verb forms that require T_EUPHONIQUE and those who don't – but that might be a lot of work, so we could also leave it for now.

You could also consider adding an optional "en". "Qu'est-ce que vous en dites ?" → "Qu'en dites-vous ?"

For this rule, the message might be more like "Dans des contextes très formels, une inversion pourrait être préférable.". Qu'en penses-tu ?

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: I added the skip="-1" and the absolute M fin inte as well as subrules for the vowel case, 'en' case and 'R pers obj' case. The message is modified accordingly and the tag style picky is added as it concerns only a specific language registrer. Finally, I created a new pull request "Question registre soutenu" as this one contains another rule for another issue.

udomai:
other PR merged

PR #4562, opened 2021-03-08
[fr] pléonasmes divers

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -424,6 +424,136 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+          <example correction="dans laquelle elle">Je n'ai pas regardé la série <marker>dans laquelle elle y</marker> joue.</example>
+      </rule>
+      <rule>
+          <pattern>
+              <token inflected="yes">pouvoir</token>
+              <token>peut-être</token>
+          </pattern>
+          <message>Cette expression peut être considérée comme un pléonasme.</message>
+          <suggestion>\1</suggestion>
+          <example correction="peux">Je <marker>peux peut-être</marker> l'aider.</example>
```
udomai, line 554: So, my dissertation was about modality. That's why I am very keen on getting modal rules right. I don't agree with this one, it is not tautological.

peut-être: expresses epistemic possibility: the speaker considers something possible.
pouvoir faire qqch: expresses mostly ability, permission, and only sometimes epistemic possibility.

Bottom line: The two modals "peux" and "peut-être" express two different things. One is about the speaker's stance, the other one about the subject's abilities. It could be rephrased this way:

"Je suis possiblement capable de l'aider".

If you really want to keep this rule, you would have to make it about the phrasing. "Voulez-vous éviter les deux formes rapprochées de 'pouvoir' en substituant l'une d'entre elles ?" peux → sais; peut-être → possiblement

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -424,6 +424,136 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+          <message>Cette expression peut être considérée comme un pléonasme.</message>
+          <suggestion>\1</suggestion>
+          <example correction="krach">En 1929, il y a eu un terrible <marker>krach boursier</marker>.</example>
+      </rule>
+      <rule>
+          <pattern>
+              <token>légère</token>
+              <token>collation</token>
+          </pattern>
+          <message>Cette expression est un pléonasme.</message>
```
udomai, line 462: Careful with any pleonasm rules. If two words occur frequently next to each other (collocations / expressions figées), we must allow them even if their meanings are somewhat tautological. This is especially true if they occur frequently in formal written speech.

All the messages have to be a lot softer. "Cette expression est considérée pléonastique, car une collation est toujours légère." (or something like that).  This rule about "little snacks" shows you why that is: If the speaker decides to say that they had not a regular "snack" but a "little snack", they might want to say that their snack was particularly tiny.

Pleonasm rules tend to be to prescriptive (and are usually not appreciated by users, just like anglicism rules). We should only write them for blatant, uncommon "abus de langage".

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: I understand the difference therefore, this rule works only for some cases but they are more cases where it doesn't. The phrasing is good but it is likely it would not be helpful in most cases for the users.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: I modified the message so it is more friendly and explains the rule.

udomai:
manually merged: 82b5a1077fd59e83957d03ad5e979c58b599d015

PR #4569, opened 2021-03-09
dans ou d'en rule

Rebecca-auque:
Confusion entre 'd'en' et 'dans'.

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -1130,7 +1130,7 @@ though, and has since been slightly modified)-->
               </marker>
               <token postag="V.*" postag_regexp="yes"><exception case_sensitive="yes" regexp="yes">[A-Z].*|puis|a</exception></token>
           </pattern>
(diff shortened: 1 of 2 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -1095,6 +1095,89 @@ though, and has since been slightly modified)-->
...
+      </rule>
+      <rule>
+          <pattern>
+              <marker>
+                  <token>d'</token>
+                  <token>en</token>
+              </marker>
+              <token postag="[DN].*" postag_regexp="yes"/>
+          </pattern>
+          <message>Une confusion entre la préposition 'dans' et la contraction 'd'en' semble être détectée. Souhaitez-vous écrire <suggestion>d'en</suggestion> ?</message>
```
udomai, line 1177: The suggestion is why the test fails ;)

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: Indeed. The problem is solved !

PR #4575, opened 2021-03-10
Enable depedency caching for travis

Taher-Ghaleb:
Would be interested to know why maven dependencies haven't been cached on Travis. Thank you.


danielnaber:
We don't use travis anymore... I think I'll remove the `.travis.yml`

PR #4589, opened 2021-03-10
Or-Antipattern bug fix

OswaldLuc:
this request intends to solve the problem of Antipattern not working properly in a rule with Or tag

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/AbstractCompoundRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/AbstractCompoundRule.java
@@ -56,7 +56,7 @@ public abstract class AbstractCompoundRule extends Rule {
   }
 
   /** @since 3.0 */
-  protected abstract CompoundRuleData getCompoundRuleData();
+  public abstract CompoundRuleData getCompoundRuleData();
 
   /**
    * @since 3.0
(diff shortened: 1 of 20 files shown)


danielnaber:
Thanks! Could you maybe add a unit test?

jaumeortola:
What is the bug fixed here?

OswaldLuc:
The bug is mentioned in languatool documentation :
[https://dev.languagetool.org/development-overview#antipatterns](https://dev.languagetool.org/development-overview#antipatterns) 
You can see that the last line of this paragraphs says :
*Be aware that if there is or in the pattern this can also lead to problems with antipattern matching (as of LT 5.1)*
This bug also leads to false suggestions.
In order to analyze the bug, I used simple french grammar.xml  tested with org.languagetool.rules.patterns.PatternRuleTest (cf attached file [grammar.xml-Bug-Or-Test.txt](https://github.com/languagetool-org/languagetool/files/6417764/grammar.xml-Bug-Or-Test.txt))
The problem was in the createRules function in the file PatternRuleHandler.java
With an Or tag this function creates one rule for each Or-altenative and therefore it goes into a loop of recursive calls to itself.
Somme variables are backup before it calls itself.
The problem is that only the address of the variables was saved which sometimes would lead to loss of data.
The solution implemented makes a copy of the variables to be saved.
Of course, it should be massively tested, especially by those who had met this problem.

A unit test would be interesting. But first of all, I don't really know how to do it. Simply adding the rule I was working on in the french rules might have been a good way achieve it. But while I was working on this bug, the french grammar.xml file has been rewritten in an interesting way. And I do not think that my rule fits in now.

Issue #4360, opened 2021-02-17
[en] a<->an 


Tex2002ans:
**a->an False Positive**

These sentences are already correct since those words start with a "you" sound:

> He can be a holder of **a** usufruct, but this is a matter of contract, not law.
> Next, I will propose **a** eudaemonistic definition of distributive justice.
> This law is needed for **a** universalized exchange economy.

So:

- usufruct
- eudaemonistic
- universalized

should be added to an a/an exception list.

* * *

**an->a Wrong Suggestion (American English)**

In American English, the word "herb" has a silent 'h', so is spoken more like "erb":

> This is **an** herb that causes catalepsy.

Currently, "a" is recommended.

In American English, "an" is correct. (Although it may depend on dialect (?).)

In British English, this rule can stay as is, since they use a hard 'h'.

Unsure of Australian/Canadian/Others.

Tex2002ans:
I also caught this today.

LT doesn't detect this in quotes:

a->an
> Hand me a "s".

but it does detect no-quote version:

a->an
> Hand me **a** s.

LT also correctly detects most vowel-sound alphabet letters within quotes too:

a->an (Correct)
> Hand me a "e".
> Hand me a "f".
> Hand me a "i".
> Hand me a "l".
> Hand me a "m".
> Hand me a "n".
> Hand me a "o".
> Hand me a "r".
> Hand me a "x".

But LT is wrong on "h" and "u".

a->an (Wrong recommendation)
> Hand me a "u".

a->an (Undetected)
> Hand me a "h".

- "h" is an odd one though.
 - In American English, it's like you're saying "eight" with a "ch" on the end: "aitch"
 - In British/Australian, I believe many accents pronounce an 'h' like: "haitch"
   - See Youtube video: ["Say letter H - #aitch​ vs #haitch​ H pronunciation"](https://www.youtube.com/watch?v=-EhSnNQ1So4)

Tex2002ans:
Fixing a/an issues...

Is it as simple as tossing the words into these 2 files?

- det_a.txt + det_an.txt

?

If so, I could create a pull request. (And I might have a lot of other words to correct too.)

A few I recommended in #4352 could go in there.

A:

- eudaemonism
- eudaimonism
- eudaemonist
- eudaimonist
- eudaemonists
- eudaimonists
- eudaemonistic
- eudaimonistic
- usufruct
- usufructs
- usufructuary
- usufructuaries
- universalized
- universalised

[...]

PR #4604, opened 2021-03-13
Correct 'An' for "ho-" + "hour-" words (#4360)

Tex2002ans:
Along with a few other rare ones.

'Y' words found in M-W using:

- \ ē
- \ ˈig

'H' words found in M-W using:

- \ ˈä
- \ ˈō

(Also removed a few duplicates.)


danielnaber:
Added manually

PR #4607, opened 2021-03-14
[ar] add ArabicTransVerbRule, ArabicPunctuationWhitespaceRule, improve tagger and add rules 

sohaibafifi:
This PR includes the following changes to the Arabic module:

- added and improved rules
- improve tagger and synthesizer to better tag pronouns
- add ArabicTransVerbRule and ArabicPunctuationWhitespaceRule


languagetool-language-modules/ar/src/main/java/org/languagetool/rules/ar/ArabicPunctuationWhitespaceRule.java
```diff
@@ -0,0 +1,228 @@
...
+      }
+      prevPrevToken = prevToken;
+      prevToken = token;
+      prevWhite = isWhitespace && !tokens[i].isFieldCode(); // LO/OO code before comma/dot
+    }
+
+    return toRuleMatchArray(ruleMatches);
+  }
+
+  private boolean isDomain(AnalyzedTokenReadings[] tokens, int i) {
```
danielnaber, line 159: There seems to be a lot of code duplication with `CommaWhitespaceRule`, could this be avoided?

sohaibafifi: This is because the `match` function of `CommaWhitespaceRule` is **final**.

sohaibafifi: May be if we duplicate rules with different type of commas.

danielnaber: Feel free to remove the `final` in `CommaWhitespaceRule.match()` if it helps to avoid duplication.

sohaibafifi: Actually I prefer the separate rules solution. It seems to be cleaner.

danielnaber:
Thanks, I've merged the PR. If you open e.g. `ArabicTransVerbRule` with IntelliJ, you see quite some "unused variable" warnings. It would be nice if you could clean them up.

PR #4626, opened 2021-03-16
[ar] add notashkeel to case conversions

sohaibafifi:
This is still a draft but is it possible to make these changes to the core module? 
[Tashkeel](https://en.wikipedia.org/wiki/Arabic_diacritics) are optional diacritics that may change the meaning of a word.


sohaibafifi:
This is still a draft for now.

danielnaber:
> This is still a draft but is it possible to make these changes to the core module?

Currently, the tests seem to fail, is that related to your change?

PR #4671, opened 2021-03-25
[en] its -> it's part 1

AzadehSafakish:
Does not cover many false positives YET; I have code that covers them, but it won’t build. However, this rule covers all of the examples in the text file you sent me EXCEPT:

Unless its something fairly impressive, I won’t remember it.
This isn’t my umbrella; its somebody else’s.
Study or don't--its up to you.
Time for a hot bath, and then its bedtime.

The first two exceptions can be easily detected by adding an <or> statement, but the online checker doesn’t allow me to do that. So, I’m sending this version of the rule (which I know works), and the next (untested) version I send you will have the <or> statement.

The third exception, I don’t know how to account for because it’s really a spelling error; ‘don’t—its’ is treated as one word.

I don’t know how to cover the fourth exception…


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -101590,6 +101590,27 @@ The accident victim died from her injuries.
...
+                <token>its</token>
+                <token regexp="yes">best|own</token>
+            </antipattern>
+            <antipattern>
+                <token regexp="yes">at|by|on</token>
+                <token>its</token>
+            </antipattern>
+            <pattern raw_pos="yes">
+                <marker>
+                    <token> its</token>
```
udomai, line 101605: `<token>its</token>`

udomai (changes requested):

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -101590,6 +101590,27 @@ The accident victim died from her injuries.
...
+            <antipattern>
+                <token postag="V.*" postag_regexp="yes" />
+                <token>its</token>
+                <token regexp="yes">best|own</token>
+            </antipattern>
+            <antipattern>
+                <token regexp="yes">at|by|on</token>
+                <token>its</token>
+            </antipattern>
+            <pattern raw_pos="yes">
```
udomai, line 101603: try it w/o `raw_pos`

PR #4672, opened 2021-03-25
[en] DRAFT its -> it's

AzadehSafakish:
An updated ‘its -> it’s’ rule that should(?) cover all of the example you sent me EXCEPT:

Study or don't--its up to you.
Time for a hot bath, and then its bedtime.

Again, ‘don’t—its’ is treated as one word, and I don’t know how to handle the last one. 

This rule doesn’t cover many false positives, specifically it doesn’t cover ‘its’ + ‘adjective’ + ‘noun’ (having trouble with the disambiguator). But, I’m still working on getting a better version to build :)


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -101590,6 +101590,30 @@ The accident victim died from her injuries.
...
+                <token>its</token>
+                <token regexp="yes">best|own</token>
+            </antipattern>
+            <antipattern>
+                <token regexp="yes">at|by|on</token>
+                <token>its</token>
+            </antipattern>
+            <pattern raw_pos="yes">
+                <marker>
+                    <token> its</token>
```
udomai, line 101605: There's an excess space ;)

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -101590,6 +101590,30 @@ The accident victim died from her injuries.
...
+                </marker>
+                <or>
+                    <token postag="NN|NNS|NN:U|NN:UN|SENT_END|PARA_END" postag_regexp="yes" negate_pos="yes">
+                        <exception>hindlimbs</exception>
+                    </token>
+                    <token regexp="yes">anything|everything|something|nothing|anybody|everybody|somebody|nobody</token>
```
udomai, line 101611: I've quickly tested this in the rule editor by using first one, then the other version of the token (thereby avoiding using `<or>`).

I really like the rule with the second token `<token regexp="yes">anything|everything|something|nothing|anybody|everybody|somebody|nobody</token>`. It's unbelievable we don't find those sentences yet!

With the first token (`<token postag="NN|NNS|NN:U|NN:UN|SENT_END|PARA_END" postag_regexp="yes" negate_pos="yes"><exception>hindlimbs</exception></token>`), there are many false positives, though:

- TransAlta Corp. sold its retail business last year

Always make sure that you test thoroughly. For each version of your pattern, put at least one example. Look for false positives and write antipatterns if needed. For each antipattern, put in an example to see whether it actually works.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -101590,6 +101590,30 @@ The accident victim died from her injuries.
             <example correction="">This <marker>is</marker> right.</example>
             <example>This <marker>is</marker> conforming with my current knowledge.</example>
         </rule>
+        <rule id="ITS_TO_IT_S_2" name="its -> it's [2]">
```
udomai, line 101593: These could be subrules of IT_IS_2, I think.

In any case, the previous commit on this topic and this one should be in the same rulegroup. You could call it "ITS_IT_IS". The first rule in it would automatically be designated ITS_IT_IS[1], the second one ITS_IT_IS[2], etc.

udomai (changes requested):

PR #4673, opened 2021-03-25
[en] BASIC START "worked here since 2002"

AzadehSafakish:
Just a very basic start on the “he worked here since 2002” issue. Obviously, there’s different tenses that you could use depending on the context that follows this phrase; that’s its own box of worms. Does not include the verb “to be,” I’ll add that on later. The logic is:

WORK HERE SINCE [DATE]  (e.g. 2002)

I work here since 2002 	—>	I have been working here since 2002

I worked here since 2002 —> 	I have worked here since 2002

Also inflected for 3rd person.

Next I want to tackle the phrase WORK HERE FOR [SPAN OF TIME e.g. 3 years]. Tenses are a bit different for that one…


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -101590,6 +101590,68 @@ The accident victim died from her injuries.
             <example correction="">This <marker>is</marker> right.</example>
             <example>This <marker>is</marker> conforming with my current knowledge.</example>
         </rule>
+        <rulegroup id="WORK_HERE_SINCE_DATE" name="present/past to progressive/perfect">
```
udomai, line 101593: Always remember `default="temp_off"`

udomai (changes requested):

udomai:
You can leave the "for" contexts for another time :)

AzadehSafakish:
Pulled a cleaner, updated version of this rule :)

PR #4684, opened 2021-03-26
[fr] nouveaux mots

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/multiwords.txt
```diff
@@ -1,4 +1,52 @@
...
+coupe bise N m s
+folk J e sp
+tarif entreprise N m s
+tarifs entreprise N m p
+plus démunis N m p
+par la suite A
+bande dessinée N f s
+bandes dessinées N f p
+double corps N m s
+court A
```
dpelle, line 42: Why do we have that?

languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/multiwords.txt
```diff
@@ -1,4 +1,52 @@
...
+fiches client N f p
+base de données client N f s
+bases de données client N f p
+un tantinet A
+latina J f sp
+langue maternelle N f s
+mauvais goût N s p
+tout de même A
+Troie Z f s
+a première vue A
```
dpelle, line 17: a → à

languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/multiwords.txt
```diff
@@ -1,4 +1,52 @@
...
+à plusieurs reprises A
+donnée personnelle N f s
+données personnelles N f p
```
dpelle, line 20: I don't understand why we need such entries.

udomai: We are going to be using `multiwords` a lot more to prevent FPs from agreement rules.

udomai:
The build fails because the tag (e.g. `N f s`) has to be separated from the (multi)word by a tab (I recommend copy-pasting a tab from a previous entry).

languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/multiwords.txt
jaumeortola: This should not be here. "Court" is obviously an ambiguous word (adjective, verb, adverb), and it should be disambiguated (if possible) in disambiguation.xml.

languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/multiwords.txt
jaumeortola: It is a simple way to disambiguate some phrases, but it should be used only when there is no possible ambiguity. 
Obviously, it is better to disambiguate in a way as general as possible. These entries are now unnecessary because the disambiguation is being done properly for the sentence `J'ai lu la politique de confidentialité des donnés personnelles du site`.

languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/multiwords.txt
jaumeortola: One token words like `Troie` should be added to added.txt, not here.

languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/multiwords.txt
Rebecca-auque: When you say 'These entries are now unnecessary because the disambiguation is being done properly for the sentence J'ai lu la politique de confidentialité des donnés personnelles du site.' you mean that the entries for 'donnée(s) personnelle(s) are now unecessary but the other entries are still needed, is that correct?

jaumeortola: Yes, the other entries are mostly good, with some exceptions. "forte puissance J f s" is not good. We need to disambiguate properly "forte" in disambiguation.xml.

PR #4718, opened 2021-03-30
[en-US] “the United States” verb agreement

AzadehSafakish:
For the American English .xml.

Handles phrases that contain either 'the United States' or 'the US' like:

The United States have --> has never had a female president.
The United States are --> is 245 years old.
What the United States need --> needs is a more competent congress.

The US do --> does not need another Trump.
It's crazy that the US do --> does not have tighter gun control laws.

etc...

No FPs with the source checker when using the attached .txt file (grepped for both 'United States' and 'US').
HOWEVER, I got 2 errors (and no warnings) when trying to commit the rule? I'm attaching a screenshot of the errors as well.
[US.txt](https://github.com/languagetool-org/languagetool/files/6232850/US.txt)
commit_errors


udomai:
To do: one more round of FP testing

AzadehSafakish:
Will get to this on Monday!

PR #4719, opened 2021-03-31
[en] comma before ending sentence with 'maybe'

AzadehSafakish:
VERY pedantic rule I know... but I saw there's a THOUGH_COMMA rule for premium, so I thought why not.

It could work maybe --> It could work, maybe.

No FPs that I can see... attaching a grepped text file for testing with source checker.
[maybe.txt](https://github.com/languagetool-org/languagetool/files/6233664/maybe.txt)


udomai:
To do: one more round of FP testing

AzadehSafakish:
Added some antipatterns, should be good to go.

AzadehSafakish:
moved to style

PR #4725, opened 2021-03-31
[fr] 3 rules cf/Bien_sûr/exception_accord_serie

remymartinDev:


udomai:
Three rule proposals? Excellent! But: Something went wrong. For now, there's only the tiny message change we made as a test.

Your rules are not part of this commit. Did you write the rules in a different branch?

Also: Please put only one rule(group) per Pull Request. That way, when one rule is ready, I can go ahead and merge it without waiting for the others.

PR #4727, opened 2021-03-31
[fr] add punctuation on abr. CF.

remymartinDev:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -36035,6 +36035,17 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_TYPOGRAPHIE" name="Typographie" type="typographical">
+      <rulegroup id="PONCTUATION_SUR_ABREVIATIONS" name="Abréviations" default="temp_off">
```
udomai, line 36038: I would make the name less general, since this rule targets one very specific (and very important!) aspect of abbreviations.

Why did you make this a `rulegroup`? As long as there's only one rule, you don't need to create a rulegroup. I.e., this would work without lines 36038 and 36048.
Rules within rulegroups do not have IDs, generally.

Test your rules with the Rule Editor – the rule editor tells you, in this case, that the correction causes the same rule to match again:

Before: `<example correction="cf.">Nous avons déjà vu cette notion, <marker>cf</marker> le cours précédent.</example>`
After: `<example correction="cf.">Nous avons déjà vu cette notion, <marker>cf</marker>. le cours précédent.</example>`
And so on: `<example correction="cf.">Nous avons déjà vu cette notion, <marker>cf</marker>.. le cours précédent.</example>`

How would you avoid this?

And then, there are false positives in the Rule Editor results. How about excluding "CF" (both uppercase)?

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
udomai: Also (as a routine), always run the mvn test before making a pull request. If the build doesn't succeed, you can make a pull request anyway, but give it the status of a "draft pull request".

Once the mvn test succeeds **and** you don't find any false positives in the Rule Editor, you can mark it as "ready for review".

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -36035,6 +36035,17 @@ though, and has since been slightly modified)-->
...
+          <rule id="PONCT_CF" default="temp_off">
+              <pattern>
+                  <marker>
+                      <token>cf</token>
+                  </marker>
+              </pattern>
+              <message>Si <suggestion>cf.</suggestion> désigne l'abréviation de "confère", il convient d'ajouter un point.</message>
```
udomai, line 36045: Ça existe, "conférer" ? C'est pas du latin _confer_ ?

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -36035,6 +36035,17 @@ though, and has since been slightly modified)-->
     </rule>
   </category>
   <category id="CAT_TYPOGRAPHIE" name="Typographie" type="typographical">
+      <rulegroup id="PONCTUATION_SUR_ABREVIATIONS" name="Abréviations" default="temp_off">
+          <rule id="PONCT_CF" default="temp_off">
+              <pattern>
+                  <marker>
+                      <token>cf</token>
```
udomai, line 36042: I'd also write an AP or an exception for CF (club de futebol etc.) to avoid FPs.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
remymartinDev: Effectivement, c'est une erreur de ma part, j'ai toujours pensé à tort que cela venait du verbe conférer

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
remymartinDev: Creation of the rulegroup : I have anticipated many addings to this rulegroup, including for example vol. for volume, coll. for collection, etc. for Et cetera/cætera, [...]

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
remymartinDev: for the moment, I did not find a way to take this into account in the rule

remymartinDev:
Due to confusion in branches, I could nio longer commit to this branch
I close it and re-create the rule in this commit :
[https://github.com/languagetool-org/languagetool/pull/4907](url)

PR #4728, opened 2021-03-31
[fr] add rule for sentence start with BIEN_SÛR on SUR_ACCENT

remymartinDev:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -3676,6 +3676,18 @@ though, and has since been slightly modified)-->
     </rulegroup>
     <rulegroup id="SUR_ACCENT" name="Sur au lieu de Sûr">
       <url>http://bdl.oqlf.gouv.qc.ca/bdl/gabarit_bdl.asp?id=2232</url>
+      <rule default="temp_off">
+          <pattern>
+             <token postag="SENT_START"/>
+             <token>Bien</token>
+             <marker>
+                 <token>sur[e?s?]</token>
+             </marker>
```
jaumeortola, line 3685: `<token regexp="yes">sur[e?s?]</token>`

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -3676,6 +3676,18 @@ though, and has since been slightly modified)-->
     </rulegroup>
     <rulegroup id="SUR_ACCENT" name="Sur au lieu de Sûr">
       <url>http://bdl.oqlf.gouv.qc.ca/bdl/gabarit_bdl.asp?id=2232</url>
+      <rule default="temp_off">
```
udomai, line 3679: If you want to add a new subrule to a rulegroup, it's best if you add it at the very bottom. Otherwise, all the other subrule IDs are going to change (leading to unnecessary diff entries: `SUR_ACCENT[1] → SUR_ACCENT[2]`, etc.).

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -3676,6 +3676,18 @@ though, and has since been slightly modified)-->
...
+          <pattern>
+             <token postag="SENT_START"/>
+             <token>Bien</token>
```
udomai, line 3682: FYI: tokens are always case-insensitive. This token could also be `<token>bien</token>` and would match the exact same words (bien, Bien, BIEN, etc.). When it's important, you can use `case_sensitive="yes"`, but in this case, I wouldn't recommend it (users don't always start their sentence with an uppercase letter).

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
udomai: I would go as far as writing

```
<marker>
    <token regexp="yes">sure?s?</token>
</marker>

File: languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml [hunk 4]
```diff
@@ -3676,6 +3676,18 @@ though, and has since been slightly modified)-->
...
+             <marker>
+                 <token>sur[e?s?]</token>
+             </marker>
+          </pattern>
+          <message>Voulez-vous écrire <suggestion>sûr</suggestion> ?</message>
```
udomai, line 3687: In general, messages have to give the user an idea why that is a mistake. They should never just be "Did you mean XYZ?". In this case, you could say `Le mot 'sur' est une préposition. Vouliez-vous dire <suggestion>bien sûr</suggestion>, au sens de 'certainement' ?` or something like that.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -3676,6 +3676,18 @@ though, and has since been slightly modified)-->
...
+          <example correction="sûr">Bien <marker>sur</marker>, on me dira que la vie est dure</example>
+          <example>Comfortablement assis, il se sentait bien sur cette chaise</example>
+        </rule>
```
udomai, line 3690: Small detail: The indent here does not match the indent in the next `<rule>`; you might want to check whether your tab is actually 4 spaces wide (IntelliJ: File > Settings > Editor > Code Style > XML)

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -3676,6 +3676,18 @@ though, and has since been slightly modified)-->
...
+          <pattern>
+             <token postag="SENT_START"/>
+             <token>Bien</token>
+             <marker>
+                 <token>sur[e?s?]</token>
+             </marker>
+          </pattern>
+          <message>Voulez-vous écrire <suggestion>sûr</suggestion> ?</message>
+          <example correction="sûr">Bien <marker>sur</marker>, on me dira que la vie est dure</example>
+          <example>Comfortablement assis, il se sentait bien sur cette chaise</example>
```
udomai, line 3689: Could you please make sure that your examples are "compelte" sentences (unless there is a technical reason for them not to be)? In this case, it's about the missing "." at the end.

udomai:
2nd version merged by Jaume

PR #4737, opened 2021-04-02
[en] tag questions

AzadehSafakish:
Kinda niche, L2 error but it was way too ugly not to fix:

You know the way, aren’t you?
I stopped here, wasn’t I?
You don’t know, are you?
It doesn’t look like it, is it?
It didn’t look like it, was it?
Well, you’re here, don’t you?
I was late, didn’t I?
I’m not in trouble, do I?
You’re not shy, do you?
He’s crazy, doesn’t he?
It was a little off, didn’t it?
It isn’t very helpful, does it?


AzadehSafakish:
already in the right place

PR #4741, opened 2021-04-03
[en] Style suggestion: replace the phrase "a whole other"

AzadehSafakish:
That's a whole other topic. → That's an entirely different topic.

The rule only replaces certain instances of the phrase; there are other instances (idiomatic expressions) where the replacement sounds bizarre, so I leave those alone (e.g. 'a whole other beast').

Do you think the message is okay?


AzadehSafakish:
Already moved to style earlier today

PR #4751, opened 2021-04-05
[en] let + infinitive

AzadehSafakish:
!!!

A very common L2(?) error, apparently?

let

LOTS of hits in the rule editor and source checker, but no FPs... :)


AzadehSafakish:
moved to grammar

PR #4759, opened 2021-04-06
[en] double negative + RP

AzadehSafakish:
LT Editor doesn't catch double negatives with an intervening particle, so I fixed that.

image


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -63936,6 +63936,63 @@ USA
                 <example>I <marker>haven't said no</marker> yet.</example>
             </rule>
         </rulegroup>
+        <rulegroup id="DOUBLE_NEG_RP" name="I didn't mention it to no one" default="temp_off">
+            <antipattern>
+                <token regexp="yes">could|would|should</token>
+                <token regexp="yes">not|n't</token>
```
udomai, line 63942: Did you know? This is totally correct, but you could have written: `<token inflected="yes">not</token>`

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -63936,6 +63936,63 @@ USA
...
+                </pattern>
+                <message>This phrase contains a double negative, or a comma may be missing. Try: <suggestion>anything</suggestion>.</message>
+                <example correction="anything">I won't stop for <marker>nothing</marker>.</example>
+                <example>From nothing there will always be something because a rule that “something can't come from nothing” would be something, and therefore it is not present in nothing.</example>
+            </rule>
+            <rule>
+                <pattern>
+                    <token regexp="yes">not|n't</token>
+                    <token postag="VB" />
+                    <token min="0" chunk_re="B\-NP\-(singular|plural)" />
```
udomai, line 63969: see my comment in the other rule, `B-NP.*`

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -63936,6 +63936,63 @@ USA
...
+                    <token min="0" chunk_re="E\-NP\-(singular|plural)" />
+                    <token postag="IN|RP|TO" postag_regexp="yes" />
+                    <marker>
+                        <token>nobody</token>
+                    </marker>
+                </pattern>
+                <message>This phrase contains a double negative, or a comma may be missing. Try: <suggestion>anybody</suggestion>.</message>
```
udomai, line 63976: In general, it's best if you can avoid putting the suggestion (or the suggested form) in the message, so that the suggestion doesn't appear twice (once in the message, and once as the suggestion itself).

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -63936,6 +63936,63 @@ USA
...
+                    <token postag="VB" />
+                    <token min="0" chunk_re="B\-NP\-(singular|plural)" />
+                    <token min="0" chunk_re="E\-NP\-(singular|plural)" />
+                    <token postag="IN|RP|TO" postag_regexp="yes" />
+                    <marker>
+                        <token>no</token>
+                        <token>one</token>
+                    </marker>
+                </pattern>
+                <message>This phrase contains a double negative, or a comma may be missing. Try: <suggestion>anyone</suggestion>.</message>
```
udomai, line 63991: We could hedge the message some more, like "might contain". Just in case somebody actually means to use litotes.

AzadehSafakish: Fixed 💯

PR #4760, opened 2021-04-06
[en] Style suggestion: tell it to X -> tell X

AzadehSafakish:
Tell it to somebody else. → Tell someone else.

I promised him that I would not tell it to anyone. → I promised him that I would not tell anyone.

I told it to my mother first and foremost. → I told my mother first and foremost.

Etc.

PLEASE let me know if you think this is a good idea or not...


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -102505,6 +102505,49 @@ The accident victim died from her injuries.
...
+                <token>to</token>
+                <token>Groucho</token>
+            </antipattern>
+            <pattern>
+                <marker>
+                    <token inflected="yes">tell</token>
+                    <token>it</token>
+                    <token>to</token>
+                </marker>
+            <token chunk_re="B\-NP\-(singular|plural)">
```
udomai, line 102538: no need to escape the `-` here, and since there are only singular and plural, you could write `B-NP.*`

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -102505,6 +102505,49 @@ The accident victim died from her injuries.
             <example correction="">This <marker>is</marker> right.</example>
             <example>This <marker>is</marker> conforming with my current knowledge.</example>
         </rule>
+        <rule id="TELL_IT" name="I [told it to] -> I [told] Leo" default="temp_off">
```
udomai, line 102508: I would put this in the category "style" rather than "creative writing"

udomai:
Nice!! I have only noted two minor things. Can you make those changes, please? Other than that, this rule is ready for merging.

AzadehSafakish:
> Nice!! I have only noted two minor things. Can you make those changes, please? Other than that, this rule is ready for merging.

Fixed 💯

PR #4765, opened 2021-04-06
[gl] Clean-up/Localization

xosecalvo:
Localization of many instances in grammar.xml where Portuguese and Galician differ (the previous maintainer was Portuguese). This affects the UI mostly, not the rules.
redundancies.txt has seen some expressions not used in Galicia removed and others added, and the list sorted alphabetically.


danielnaber:
Thanks, I've merged your changes.

PR #4768, opened 2021-04-07
[en] have 'following questions'

AzadehSafakish:
This wasn't the issue that I originally set out to fix, but in the rule editor I noticed a lot of these sentences that LT doesn't correct:

image

The proposed rule suggests a 'the' before 'following + noun.'

Under settings, I have following options → Under settings, I have the following options


AzadehSafakish:
moved to grammar category

PR #4779, opened 2021-04-08
[en] Extra Tags for PRP

udomai:
This is a draft for adding more detailed postags to personal pronouns (and to be able to synthesize them directly).

For now, the build breaks.

- some tests fail (can be fixed)
- some rules break (e.g. if they look for PRP, except for postag_negate PRP → can be fixed)
- synthesis of "mine" from "me" doesn't seem to work (see new rule BELONG_TO_PRP, expects "was me|were me")

I am starting to wonder whether this change is actually worth putting the necessary time in. @jaumeortola or @MikeUnwalla, this is not urgent. When you have the time, I would appreciate your take on this. Am I doing this right? I think these extra tags are harder to add than I thought, do you agree?

Here's an overview of the tags I wanted to add:

<details>
    <summary>Extra PRP tags</summary>

<br>**Subject pronouns (SUBJ):**

I	I	SUBJ_1S
you	you	SUBJ_2S, SUBJ_2P
he	he	SUBJ_3SM
she	she	SUBJ_3SF
it	it	SUBJ_3SN
they	they	SUBJ_3S, SUBJ_3P
we	we	SUBJ_1P

**Object pronouns (OBJ):**

me	I	OBJ_1S
you	you	OBJ_2S, OBJ_2P
him	he	OBJ_3SM
her	she	OBJ_3SF, POSS_3SF
it	it	OBJ_3SN
them	they	OBJ_3S, OBJ_3P
us	we	OBJ_1P

**Reflexive pronouns (REFL):**

myself	I	REFL_1S
yourself	you	REFL_2S
himself	he	REFL_3SM
herself	she	REFL_3SF
itself	it	REFL_3SN
themself	they	REFL_3S
ourselves	we	REFL_1P
yourselves	you	REFL_2P
themselves	they	REFL_3P

**Possessive attributes (POSS):**

my	I	POSS_1S
your	you	POSS_2S, POSS_2P
his	he	POSS_3SM
her	she	POSS_3SF
its	it	POSS_3SN
their	they	POSS_3S, POSS_3P
our	we	POSS_1P

**"Genitive" pronouns (GEN):**

mine	I	GEN_1S
yours	your	GEN_2S, GEN_2P
his	he	GEN_3SM
hers	she	GEN_3SF
its	it	GEN_3SN
theirs	they	GEN_3S, GEN_3P
ours	we	GEN_1P
</details>

Here's what the maven test currently says:

[build fail](https://github.com/languagetool-org/languagetool/files/6278577/PRPtest.txt)


MikeUnwalla:
I like the idea of postags. The alternative is to use entities (we already have object_pronouns and subject_pronouns in grammar.xml). Postags are better, because they are specified in one place only, and they can be used in all rules in all files.

Adding _only_ new tags should not cause Maven errors for disambiguation, except if in disambiguation, a marked word in an 'ambiguous' example is a word that has a new postag. That problem is easy to deal with (I do it each time I update the STE checker). Just add the new POS in the examples.

disambiguation.xml/rulegroup id="PRP_TAGS_EXTRA", rule 2 has 1 token in the pattern. But, the action is:
`<disambig action="add"><wd pos="OBJ_3SF"/><wd pos="POSS_3SF"/></disambig>`
As best I know, that is not correct. You must make 2 separate rules. One rule adds OBJ_3SF, another rule adds POSS_3SF.

But, what is the purpose of PRP_TAGS_EXTRA? There is no disambiguation. Isn't the rulegroup redundant? The tags are also in added.txt.

Maven analysis line 454 shows a problem with IN_PRP_THEN_IN_PRP. I _guess_ that the problem is caused by the exception with negate_pos:
`<exception postag="PRP|NN(:UN?)?" postag_regexp="yes" negate_pos="yes"/>`

The correction suggestion "between him than between she" is not good English. Refer also to the comment in the rule. The rule is a candidate for deletion.

udomai:
Thank you for the tips, @MikeUnwalla! I've corrected the disambiguation rules with more than one "add", dividing them in twixt. Or thrixt/quadrixt, respectively.

You were right about why the rules caused test failures, they were easy to fix. And then there was a test to fix. Now the build works fine.

However, synthesizing pronouns based on the new tags doesn't work yet (I've commented out my experimental rule).

If I add the new pronoun tags to added.txt, they will replace the old tagging (if I'm not mistaken). **How can I make it so I can actually synthesize pronouns in suggestions using the new tags?**

Also, I was thinking that maybe my tags are not practical. Would it make sense if, instead of "SUBJ_1S", I made them look like this "PRP_SUBJ_1S"? It's a longer tag, but it will be easier to formulate something like `<exception negate_pos="yes" postag="PRP_.*" postag_regexp="yes"/>` instead of the current `<exception negate_pos="yes" postag="PRP|(SUBJ|OBJ|REFL|POSS|GEN).*" postag_regexp="yes"/>`.

MikeUnwalla:
> If I add the new pronoun tags to added.txt, they will replace the old tagging (if I'm not mistaken). 

No. Refer to the contractions in added.txt (line 6 onwards). Also, lines 23 to 25 for o'er.

Unless you have a good reason, I suggest that it's better to add tags in added.txt rather than in disambiguation.xml. With the added tags in disambiguation, you don't disambiguate, you just add tags. You might as well do that with added.txt (it's shorter and clearer). Then, you can use the added tags anywhere in disambiguation.xml (currently, you don't, but someone might want to in the future).

> How can I make it so I can actually synthesize pronouns in suggestions using the new tags?

I don't know. I don't remember ever doing something like that.

> Would it make sense if, instead of "SUBJ_1S", I made them look like this "PRP_SUBJ_1S"? 

Yes. For the reason you give. Also, PRP_ makes it clear that the word to which the tag applies is a type of pronoun.

udomai:
I messed up the update of this branch, I think. A clean version is here: #4870

PR #4782, opened 2021-04-09
[en] this [will allow] -> this [allows]

AzadehSafakish:
In the redundancy category.

image


MikeUnwalla:
The rule is in the redundancy category and the message is "This phrasing could be wordy".

The examples are not examples of redundancy or unnecessary wordiness. Whether the 'corrections' are indeed correct is a matter of semantics. If the text is about repeated actions, then the simple present is a good suggestion. But, in many cases, one really does want to refer to a future action.

Possibly, better categories are 'semantics' or 'style'.

AzadehSafakish:
You're absolutely right; I've moved it to the style category and changed the message: 'Future tense may not be necessary in this context.'

PR #4783, opened 2021-04-09
[en] used for VBG

AzadehSafakish:
Because this is embarrassing...

In the style category.

image
image

They were used [to hunt] partridges → They were used [for hunting] partridges


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -96278,6 +96278,65 @@ The accident victim died from her injuries.
...
+                <marker>
+                    <token>to</token>
+                    <token postag="VB">
+                        <exception regexp="yes">be|date|do|lube|mean|slow|up</exception>
+                    </token>
+                </marker>
+            </pattern>
+            <message>Try using 'for' + the -ing form of the verb.</message>
+            <suggestion>for <match no="5" postag="VBG" /></suggestion>
+            <example correction="for hunting">They are used <marker>to hunt</marker> partridges.</example>
```
udomai, line 96329: I'm not sure at all, but "being used to VBG" is also correct, meaning "being accustomed to VBG"?

_Did you hear that guns are forbidden in this forest now?_ — _That's okay, we're used to hunting with bow and arrow._

Are sentences like that plausible at all?

AzadehSafakish: Yes, in my first draft of this rule I made exceptions for I, you, he, she, we, and they, because those are the contexts where it's almost certain "used to" means "being accustomed to" (I am used to, we are used to, etc). It seems I forgot to add them in again after switching to NP chunks. Will add them back in ASAP.

AzadehSafakish: Added in exceptions for previously mentioned pronouns + proper nouns.

Also added to exception list for verbs, none of which actually generate any FPs, they just sound a little awkward (but using for + VBG is still grammatically correct).

If you check in the rule in the editor, no FPs come up and I believe all of the sentences could be improved with for + VBG. However, it is a style rule and I've added the "picky" tag. 

BUT if you're still not too keen on it, I could scrap the rule and go with @MikeUnwalla 's suggestion to just patch up ADMIT_ENJOY_VB for passive voice, and leave it at that.

PR #4784, opened 2021-04-09
[en] improved WHOLE_OTHER

AzadehSafakish:
Updated message:

The expression 'a whole other' can seem colloquial. Consider replacing it with a more formal alternative.


AzadehSafakish:
Checking it out right now hold on...

PR #4822, opened 2021-04-14
[fr] nombres décimaux singulier

Rebecca-auque:


jaumeortola:
I add this pattern to the rule AGREEMENT_NUMERAL_SINGULAR, which is a simple rule. No need for another rule.

PR #4831, opened 2021-04-15
nombre début de phrase

Rebecca-auque:


udomai:
I will discuss with Jaume whether we can bring the `postag="_spell_number_"` to French easily.

PR #4832, opened 2021-04-15
[en] improved USED_FOR_VBG

AzadehSafakish:
Removed the "picky" tag and added a few more awkward sounding verbs (in the context of 'for + VBG') to the exception list. 
The list was getting kind of long, so I made it an entity. 

Even though the rule excludes a lot of verbs, I think it's still going to be opened a lot and I think having a good number of exceptions will increase the apply rate.


udomai:
This looks good! Can it be switched from "Draft" to "Ready for Review"?

AzadehSafakish:
Took care of the double construction issue

PR #4856, opened 2021-04-19
[en] LIMITED_TO

AzadehSafakish:
image

Just a small typo I noticed with the too vs either rule, however LT offers a wrong suggestion.


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -19037,6 +19037,24 @@ USA
...
+                <marker>
+                    <token>too</token>
+                </marker>
+            </pattern>
+            <message>Use 'to' in this context.</message>
+            <suggestion>to</suggestion>
+            <example correction="to">There are many activities, including, but not limited <marker>too</marker>, running, jumping, and swimming.</example>
+            <example correction="to">They're trying to deal with all the issues, which include but are not limited <marker>too</marker> updates, bugs, and overloaded servers.</example>
+            <example correction="to">They're trying to deal with all the issues, which include but are not limited <marker>too</marker>: updates, bugs, and overloaded servers.</example>
+            <example>The game/Dev team is limited too.</example>
```
MikeUnwalla, line 19056: Also should not suggest 'to':
It's rather limited too (in my opinion).

AzadehSafakish: I've updated the pattern to only target 'include-inflected but (be-inflected) not limited too...'

PR #4857, opened 2021-04-19
[en] TOO_TO_EITHER

AzadehSafakish:
image

I'm aware there's a rulegroup to deal with this issue, however it seems to miss contexts where adverbs are involved. So I patched that up.


udomai:
In this match in the Rule Editor, I think "to" would be the better suggestion. Maybe there is a series of verbs that risk ending a sentence in "too" (misspelling of "to"):

_I also got criticism for living with my mom and not paying rent even though i couldn't afford **too**._

Same might have happened with _dare, want…_

AzadehSafakish:
> Maybe there is a series of verbs that risk ending a sentence in "too" (misspelling of "to"):
> 
> Same might have happened with _dare, want…_

Updated entity + AP

udomai:
merged here: 5bf32908d9d445d319808afd4cb74c7b72ade686

PR #4858, opened 2021-04-19
[en] etc. punctuation

AzadehSafakish:
image

Where I come from, 'etc.' needs a period mid-sentence, which LT editor misses. 

However, the rule has a lot of hits; I'll have to look more into the specifics of 'etc.' punctuation but I think most of what the rule detects is accurate.


MikeUnwalla:
In BrE, a period (full stop) is not necessary: https://www.ldoceonline.com/dictionary/etc#etc__3

I suggest that you put the rule in `rules\en\en-US`.

tiff:
Great rule. I think this rule will have a high acceptance rate as the suggestion is simple and doesn't require much attention by the user

Issue #4841, opened 2021-04-17
[en] Not found: studies have show that lobyist have


MikeUnwalla:
Noted in https://internal1.languagetool.org/regression-tests/via-http/2021-04-17/en-US/result_grammar_PLURAL_VERB_AFTER_THIS[2].html

Before my change to the disambiguation for 'have', PLURAL_VERB_AFTER_THIS[2] gave a warning for 'have' and suggested 'has':

> ... studies have show that lobbyist **have** become less influential in ...

That suggestion is not correct, so I will not change PLURAL_VERB_AFTER_THIS[2] such that it gives a warning. A better correction is:

> ... studies **have shown that lobbyists** have become less influential in ...

AzadehSafakish:
I've updated the rule to suggest both forms:

lobbyist has
lobbyists have

PR #4871, opened 2021-04-20
[en] improve PLURAL_VERB_AFTER_THIS for #4841

AzadehSafakish:
Issue #4841 

PLURAL_VERB_AFTER_THIS[5] - ‘that' as an SBAR

Studies have shown that lobbyist have become less influential in state politics.

Will suggest:
lobbyist has
lobbyists have

Should I temp off a rule that’s already active?


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -34883,8 +34882,124 @@ USA
...
+                    <token postag="PRP" />
+                    <token>that</token>
+                </antipattern>
+                <antipattern>
+                    <token regexp="yes">and|see</token>
+                    <token>that</token>
+                </antipattern>
+                <pattern>
+                    <token>that</token>
+                    <token postag_regexp='yes' postag='NN|NN:U'>
```
tiff, line 34942: why not also "NN:UN"? (both uncountable and countable)

AzadehSafakish: I originally excluded it as it triggered an FP in rule [3], and I reused the code for [4]. However upon further inspection, the FP can be handled with a refined antipattern, so I’ve updated [3], [4], and [5] to include NN:UN. Also, I excluded NNP from [3] to be consistent with the other rules; including it brings up the issue surrounding company names (AKA collective nouns).

udomai:
If the rulegroup is already active, you can `<rule default="temp_off">` the single subrules.

PR #4874, opened 2021-04-20
[fr] nombre début de phrase rule improved

Rebecca-auque:


udomai:
Note to self: don't merge yet, will be split in twixt

Issue #2520, opened 2020-02-27
[en] wrong tagging PRPPOSS_VB_NN


tiff:
> The very fact that you suddenly thought about her **makes** it obvious that you like her.

"makes" is only NNS

PR #4893, opened 2021-04-23
[en] missing commas in progress

AzadehSafakish:
Premium issue #2520, which doesn't link here...


AzadehSafakish:
moving to premium

PR #4894, opened 2021-04-23
[en] either nor

AzadehSafakish:
image


udomai:
I like this rule!

Is this an FP or a TP? I've confused myself and cannot tell anymore.

```
<example>It should <marker>not overload</marker> either of these circuits, nor cause any other electrical problems.</example>
```

AzadehSafakish:
> <example>It should <marker>not overload either</marker> of these circuits, nor cause any other electrical problems.</example>
> ```

TP, however it brings up an important issue with 'either of'. As corrected by the rule:

It should overload **neither** of these circuits, **nor** cause any other electrical problems.

Ideally: It should overload neither [the circuits], nor [something else].
Currently (with 'neither of'): It should overload neither [this circuit] nor [that circuit], nor [something else].

In this case, I would prefer to correct the 'nor' rather than the 'either'. So, the preferred correction would be:

It should not overload **either** of these circuits, **or** cause any other electrical problems.

I've added an AP and another subrule to deal with this on another branch, because I'm having trouble with rebasing...

AzadehSafakish:
> ```
> <example>It should <marker>not overload either</marker> of these circuits, nor cause any other electrical problems.</example>
> ```

TP, however it brings up an important issue with 'either of'. As corrected by the rule:

It should overload **neither** of these circuits, **nor** cause any other electrical problems.

Ideally: It should overload neither [the circuits], nor [something else].
Currently (with 'neither of'): It should overload neither [this circuit] nor [that circuit], nor [something else].

In this case, I would prefer to correct the 'nor' rather than the 'either'. So, the preferred correction would be:

It should not overload **either** of these circuits, **or** cause any other electrical problems.

I've added an AP and another subrule to deal with this on another branch, because I'm having trouble with rebasing...

udomai:
Closing, merging follow-up PR

PR #4909, opened 2021-04-25
[fr] using "à" instead of "de"

remymartinDev:
this is a first shot for this topic, but this provides too much FP for now, that is why I push it as a draft-PR

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -41826,6 +41826,22 @@ though, and has since been slightly modified)-->
         </rule>
     </category>
     <category id="CAT_TOURS_CRITIQUES" name="Tours critiqués divers (barbarismes, impropriétés, solécismes, etc.)">
+        <rule default="temp_off">
+           <pattern>
+                <token postag="D.*" postag_regexp="yes"/>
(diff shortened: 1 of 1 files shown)


udomai:
The APs for "barbe à papa" and for "N.* à [V inf]" are missing. Pro-tip: if there are known FPs, instead of `correction="…"`, you can write `type="triggers_error"`. That way the build doesn't fail. Don't forget about those examples at the end, though ;)

The first AP is trivial, and the second one should be easy, too (no infinitives). Subsequently, you go hunting for the more difficult-to-get FPs.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -41826,6 +41826,98 @@ though, and has since been slightly modified)-->
...
+                    <unify>
+                        <feature id="number"/>
+                        <feature id="gender"/>
+                        <token postag="D.*" postag_regexp="yes" inflected="yes">le</token>
+                        <token postag="N.*" postag_regexp="yes"/>
+                    </unify>
+                    <token>à</token>
+                    <token>moi</token>
+                </pattern>
+                <message>" \1 \2 \3 \4 " est une expression fautive.</message>
```
udomai, line 41841: "…" peut être considéré du langage familier. Dans des contextes plus formels, évrivez …

`<url>` ?

udomai (changes requested):

PR #4915, opened 2021-04-26
Rm6 punctuation on cf

remymartinDev:


tiff:
"CF" could also be a file extension
https://fileinfo.com/extension/cf

Also, it's the top-level domain of Central Africa, a country that speaks French.

e.g.
my_file.cf

or 

www.domain.cf

Some antipattern like:
```
<antipattern>
  <token regexp="yes">[._]</token>
  <token spacebefore="no">cf</token>
</antipattern>
```

Should make the rule more robust

tiff:
Also, what about cases where an ellipsis comes after "cf".

E.g. 
> cf… 

Do we want to add punctuation here as well?

remymartinDev:
Thank you for the feedback, @tiff 
I added it in last commit

udomai:
Manually merged: 5405dd3b5cedeacccb8cb9b67f5ecae592887efb

Issue #4910, opened 2021-04-25
[el]  ReplaceHomonymsRule for Greek 


Nikos-Antonopoulos:
Apparently Greek does not have a rule that detects homonyms. There are many cases where replacements should be suggested!

Nikos-Antonopoulos:
I'll take care of this.

PR #4922, opened 2021-04-27
[el] Create rule for homonyms

Nikos-Antonopoulos:
This rule, detects words that share the same pronunciation but the spelling and the meaning of them is different.  It solves issue #4910.


jaumeortola:
Thank you. The rule seems okay. Have you tested for possible false positives? For example: 
```
μεγάλη λείπει=μεγάλη λύπη 
- Είναι εδώ τα κορίτσια;
- Η μεγάλη λείπει.
```
If there are false positives, then the replacement should not be here but in a pattern rule (with context or exceptions).

Nikos-Antonopoulos:
I've updated the replace.txt accordingly!

PR #4930, opened 2021-04-29
[en] either nor

AzadehSafakish:
Original branch AS-14 wasn't up-to-date; I rebased, fetched, pulled, now this. 

image

Pushing the rule on another branch to be safe.


udomai:
Let's look into that next week. My hypothesis is that the origin/master might not be in sync with the upstream/master...

PR #4938, opened 2021-04-30
[en] make HTML comment AP in ARROWS softer

udomai:
@tiff: I would like to make this AP softer, to match only real HTML comments. I think `-->` is a more common arrow replacement than `->` and I think that excluding it categorically would make `ARROWS` a lot less useful…

What do you think?


tiff:
Of course! Thanks!

Issue #4950, opened 2021-05-03
[en] EnglishWordRepeatBeginningRule suggestion of adverbs


giorgossideris:
I think it would be nice, if the same adverb is detected in the beginning of successive sentences, to suggest different ones to replace it.

giorgossideris:
I will add this feature

PR #4951, opened 2021-05-03
[en] Add adverb suggestions to EnglishWordRepeatBeginningRule (#4950)

giorgossideris:
Added feature to suggest different adverbs, when successive sentences start with the same adverb.

*Example*

image

Also, added a relative test case.

### DIFF
--- a/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/EnglishWordRepeatBeginningRule.java
+++ b/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/EnglishWordRepeatBeginningRule.java
@@ -45,6 +45,8 @@ public class EnglishWordRepeatBeginningRule extends WordRepeatBeginningRule {
     return "ENGLISH_WORD_REPEAT_BEGINNING_RULE";
   }
   
+  	//==================== ADVERBS ======================
+  
 	// adverbs used to add to what the previous sentence mentioned
 	private static final Set<String> ADD_ADVERBS = new HashSet<>();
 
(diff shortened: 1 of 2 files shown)


languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/EnglishWordRepeatBeginningRule.java
```diff
@@ -44,35 +45,84 @@ public String getId() {
...
+		// based on https://www.pinterest.com/pin/229542912245527548/
+		ADD_ADVERBS.add("Additionally");
+		ADD_ADVERBS.add("Besides");
+		ADD_ADVERBS.add("Furthermore");
+		ADD_ADVERBS.add("Moreover");
+		ADD_ADVERBS.add("Also");
+		CONTRAST_ADVERBS.add("Unlike");
+		CONTRAST_ADVERBS.add("Nevertheless");
+		CONTRAST_ADVERBS.add("Nonetheless");
+		CONTRAST_ADVERBS.add("Despite");
```
udomai, line 70: Thank you, @giorgossideris! This is a cool rule!

Just one thing: Could you take out "Unlike" and "Despite"? They are not adverbs and, most importantly, they cannot be used as sentence adverbs at the beginning of a sentence ("Nevertheless, …" is okay, "Despite, …" and "Unlike, …" are not). So if someone writes several sentences starting with "Nonetheless", they should not be offered as possible replacements.

They could go into a different set, e.g. CONTRAST_PREPOSITIONS…

udomai (changes requested):

languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/EnglishWordRepeatBeginningRule.java
giorgossideris: @udomai That's a pretty good observation, I didn't think of that. I will make the appropriate changes!

udomai (approved):

PR #4961, opened 2021-05-03
[en] draft: subject-focused S-V agreement rule

udomai:
This is a first draft to try out a subject-centred approach. We could work on it some more and write plenty of APs, until the FPs are gone.

Is there a better approach, @MikeUnwalla?


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -19232,6 +19232,46 @@ USA
...
+                <token chunk_re="B-VP" postag="VBP?" postag_regexp="yes">
+                    <exception scope="previous" regexp="yes">(some|any)(body|one)|i|you|we|they|this|that</exception>
+                    <exception scope="previous" postag="N.*S|UNKNOWN|WP|PCT|``" postag_regexp="yes"/>
+                    <exception regexp="yes">do|install|curl|renew|remove|create|set|run|make|please|thank</exception>
+                    <exception postag="VBD|MD" postag_regexp="yes"/>
+                    <exception scope="next" postag="PCT"/>
+                    </token>
+                </marker>
+            </pattern>
+            <message>This verb form does not seem to match the subject.</message>
```
MikeUnwalla, line 19265: The message is fine if the subject and the verb are close, as in the first example.

The message is not so good for the third example if the user does not have a good knowledge of grammar. What is the subject? A nice enhancement would be to include the subject 'lady' in the message.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -19232,6 +19232,46 @@ USA
...
+                <token skip="3" postag="SENT_START"/>
+                <token postag="SENT_END"/>
+            </antipattern>
+            <antipattern>
+                <token>there</token>
+                <token>are</token>
+            </antipattern>
+            <pattern>
+                <token postag="SENT_START"/>
+                <token skip="-1" chunk_re=".-NP-singular" max="10">
```
MikeUnwalla, line 19251: The chunker becomes confused if a plural noun is in the middle of a noun cluster. Example:
The **management resources test** in the new proposal show how errors occur.

'management resources test' is identified as a plural noun cluster.

udomai:
Thank you for your input, @MikeUnwalla! I've developed a rule to serve as a basis for our subject-verb agreement sprint. Details in the dedicated issue :)

Issue #4962, opened 2021-05-03
[en] Subject-Verb Agreement – which strategy is best?


udomai:
If we want to catch basic subject-verb agreement errors, I'd start with the safest scenarios.

For example: SENT_START + singular subject + plural verb form

How can we approach this most efficiently? Is the subject-centered approach the best one?

Here is a draft PR for that: https://github.com/languagetool-org/languagetool/pull/4961

udomai:
A different strategy could be: Going more fine-grained.

First word of the sentence: `DT|PRP\$`..., then min="0" RB, then min="0" JJ…

PR #4972, opened 2021-05-05
[en] sv agreement relative clauses #4962

AzadehSafakish:
image

Fixed.
Issue #4962 

The rule is ready to go out, save for two issues:

The first (minor) one: The suggestion for a VBD form for any inflection of 'to be' is going to suggest 'were', which we don't want here (ex. This red ball gown that supposedly goes well with your eyes **are** actually kinda ugly to me; suggest is, was, **were**, and will be). Do we need a separate rule for 'be' verbs?

The second issue, which irritates me to no end: No matter what I do, I cannot seem to get it to catch sentences with a subject in the relative clause (ex. The dusty book that **you** read about red ants and termites seem very informative). I'll keep playing with it, but I think it may have to do with the antipatterns at the top of the rulegroup (none of which my rule actually use). If that's the case, I may establish a separate rulegroup.


MikeUnwalla:
>  but I think it may have to do with the antipatterns at the top of the rulegroup (none of which my rule actually use). If that's the case, I may establish a separate rulegroup.

If the AP prevent your rule from finding errors, then you are correct to make a new rulegroup. To try to change the existing AP will cause you much grief.

AzadehSafakish:
image

Created a separate rulegroup that handles embedded subjects + longer verb phrases.
Will write the separate rule for ’to be’ inflections tomorrow, and if I get to it, the rules for plural subjects.
After that, I’ll start on clauses that don’t contain a prepositional phrase.

AzadehSafakish:
image

Handles plural subjects with relative clauses.

All that is left to do:
Fix [1] to catch single-word subjects + clause
Write two more rules (one for singular and one for plural subjects) to catch relative clauses WITHOUT a prepositional phrase.

udomai:
About the was/were conundrum: I'm still working on a solution -- more on Thursday 🙂

AzadehSafakish:
> About the was/were conundrum: I'm still working on a solution -- more on Thursday 🙂

Two new rulegroups in progress, hope to have them up and scrap this draft by Tuesday.

AzadehSafakish:
Pushed a new draft, closing this one.

PR #4974, opened 2021-05-05
[fr] activate rules

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/added.txt
```diff
@@ -1218,4 +1218,17 @@ tripé	triper	V ppa m s
...
+postdoc	postdoc	J e sp
+postélectoral	postélectoral	J m s
+postélectorale	postélectoral	J f s
+postélectorales	postélectoral	J f p
+postélectoraux	postélectoral	J m p
+postimpressionniste	postimpressionniste	J e sp
+postimpressionniste	postimpressionniste	N e sp
+postimpressionnisme	postimpressionnisme	N e sp
+postuniversitaire	postuniversitaire	J e s
+postuniversitaires	postuniversitaire	J e p
```
jaumeortola, line 1234: Are these plural forms correct? Shouldn't they be with 's'? postsoviétiques, postromantiques...

Rebecca-auque: Yes, apparently there are correct. I added them to the file.

PR #4978, opened 2021-05-06
[fr] add TIRET_BAS (WORD_CONTAINS_UNDERSCORE)

udomai:
@jaumeortola, when you have a minute (not urgent), could you look at this rule?

I was trying to port the WORD_CONTAINS_UNDERSCORE rule from English to French (TIRET_BAS). The main difference being that the underscore is not a token boundary in French. I thought I had adapted everything, but I'm getting a test failure I can't understand. I've run through a few possible solutions, but I cannot identify the problem.

Do you see what's going wrong? Hopefully I just oversaw something really simple...

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -39538,9 +39538,9 @@ though, and has since been slightly modified)-->
                 <token regexp="yes">&comp_mots;+_+&comp_mots;+<exception postag="UNKNOWN" negate_pos="yes"/></token>
             </pattern>
             <message>Un tiret bas (_) à l'intérieur d'un mot est atypique (sauf dans des contextes techniques, les pseudos dans les réseaux sociaux etc.). Veuillez vérifier que le mot '\1' soit correct.</message>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -39491,6 +39492,71 @@ though, and has since been slightly modified)-->
...
+                <token regexp="yes">&comp_mots;+</token>
+                <token regexp="yes" spacebefore="no">&comp_mots;+</token>
+                <token regexp="yes" spacebefore="no">['‘"“]</token>
+            </antipattern>
+            <antipattern>
+                <token skip="3" regexp="yes">=|\*|\/|\+|\-|%</token>
+                <token regexp="yes">.*_.*</token>
+            </antipattern>
+            <antipattern>
+                <token skip="3" regexp="yes">/=|\*|\/|\+|\-|%)_</token>
```
jaumeortola, line 39517: The `)` is unmatched. You need to escape it, or to add `(` previously.

udomai: Thank you so much!!! I was going crazy yesterday because I couldn't find the error.

🙇

PR #4981, opened 2021-05-07
[fr] trait d'union ou pas rule improved

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -23787,17 +23787,6 @@ though, and has since been slightly modified)-->
...
-                    <token>post</token>
-                    <token min="0">-</token>
-                    <token postag="[NJ].*|V ppa.*" postag_regexp="yes" regexp="yes">\p{Ll}.*</token>
-                </pattern>
-                <message suppress_misspelled="yes">Le préfixe '\1' est généralement associé au terme qu'il précède.</message>
-                <suggestion><match no="3" regexp_match="^(.*)$" regexp_replace="post$1" case_conversion="preserve"/></suggestion>
-                <example correction="postuniversitaires">Les <marker>post universitaires</marker>.</example>
-                <example>Les troubles du stress post-traumatique.</example>
-            </rule>
         </rulegroup>
```
jaumeortola, line 23790: This should not be removed. Otherwise, we don't detect an error in "post impressionniste". We can change the tagger if it accepts too many undesired words. Or we can add antipatterns and exceptions (that is what I have done for this prefix in Catalan and Spanish).

Rebecca-auque: Okay, I put it back. I would not choose antipatterns and exceptions as there is no clear pattern of why some words are with an hyphen and other are not. For example "le postimpressionnisme" is correct while "le postrepas" is not. With APs, all the wrong words would have to be looked for. So how can the tagger be changed?

Rebecca-auque: Okay, I put it back. I would not choose antipatterns and exceptions as there is no clear pattern of why some words are with an hyphen and other are not. For example "le postimpressionnisme" is correct while "le postrepas" is not. With APs, all the wrong words would have to be looked for. So how can the tagger be changed?

udomai: I could do a quick corpus research of the most frequent words following the pattern `post[^-].*` or `post-.*`…

udomai: If it can be useful, here are the results, ordered by frequency:

[post[^-]{7,}.txt](https://github.com/languagetool-org/languagetool/files/6441125/post.-.7.txt)
[post-….txt](https://github.com/languagetool-org/languagetool/files/6441126/post-.txt)

udomai: Cleaner versions (without numbers):
[post[^-]{7,}.txt](https://github.com/languagetool-org/languagetool/files/6441157/post.-.7.txt)
[post-….txt](https://github.com/languagetool-org/languagetool/files/6441158/post-.txt)

Rebecca-auque: Okay, I'll add them to the added and compound file.

PR #4988, opened 2021-05-09
[gl] added Acronyms - Plural; removed Puntual

xosecalvo:
Added a new rule for acronyms, which do not inflect for plural.
Removed rule about 'puntual', already set to off a long time ago: new meaning is now accepted in DRAG.


danielnaber:
Could you check your PR? I don't see any changed files at https://github.com/languagetool-org/languagetool/pull/4988/files

PR #4989, opened 2021-05-10
[en] improved VERB_HERE_SINCE

AzadehSafakish:
All of the deactivation contexts had to do with 'since' being used like 'because', so I rewrote the rulegroup to restrict the pattern to only 'since + dates/common expressions of time'.


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -20099,90 +20099,95 @@ USA
             </rule>
         </rulegroup>
         <rulegroup id="VERB_HERE_SINCE" name="perfect tense + since">
```
udomai, line 20101: Could you please temp_off it again?

AzadehSafakish: 💯

PR #4993, opened 2021-05-10
[fr] verbe pronom

Rebecca-auque:


jaumeortola:
Added manually as ACCORD_V_QUESTION2.

PR #4998, opened 2021-05-11
[fr] auxiliaire manquant

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -18435,6 +18435,72 @@ though, and has since been slightly modified)-->
...
+                </pattern>
+                <filter class="org.languagetool.rules.fr.FindSuggestionsFilter" args="wordFrom:2 desiredPostag:V.*(ind|sub).*2.s removeSuggestionsRegexp:^.*[éè]$"/>
+                <message>Un verbe conjugué est généralement attendu après le pronom personnel sujet « \1 ».</message>
+                <suggestion>as \2</suggestion>
+                <suggestion>avais \2</suggestion>
+                <suggestion>es \2</suggestion>
+                <suggestion>étais \2</suggestion>
+                <suggestion><match no="2" postag="V.* ppa.*" postag_regexp="yes" postag_replace="V.* ind psim 1 s"/></suggestion>
+                <example correction="as eu|avais eu|es eu|étais eu|eus|es|eus">Tu <marker>eu</marker> la chance d'exercer le métier de conseiller pendant 1 an.</example>
+            </rule>
```
udomai, line 18454: I like this rule!!

In this case, the filter is not needed, right? What was going on with those suggestions? There's something going wrong in the replacement process. I only recommend `postag_replace` if you want to recycle part of the original postag. Since this is not the case here, you can do it much more easily. Also, I've added an example with another verb, just to check if everything looks fine:


```xml
           <rule>
                <pattern>
                    <token>tu</token>
                    <marker>
                        <token postag="V.* ppa.*" postag_regexp="yes"></token>
                    </marker>
                </pattern>
                <message>Un verbe conjugué est généralement attendu après le pronom personnel sujet « \1 ».</message>
                <suggestion>as \2</suggestion>
                <suggestion>avais \2</suggestion>
                <suggestion>es \2</suggestion>
                <suggestion>étais \2</suggestion>
                <suggestion><match no="2" postag="V [a-z]* ?ind psim 2 s" postag_regexp="yes"/></suggestion>
                <example correction="as eu|avais eu|es eu|étais eu|eus">Tu <marker>eu</marker> la chance d'exercer le métier de conseiller pendant 1 an.</example>
                <example correction="as vu|avais vu|es vu|étais vu|vis">Tu <marker>vu</marker> la gueule du mec?</example>
            </rule>
```

udomai (changes requested):

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Rebecca-auque: I put the filter for the first subrule as I thought sometimes it can be one of the past tense suggested but it could also be another tense (past tense or not). For example, 'tu mangé' could be 'tu mangeais'. So I thought the suggestion scope could be improved with the filter.

udomai: I see! That makes sense. We need to prioritize the suggestions here, since the user will only see five of them. You could throw out the suggestions with "avais" and "étais" and put `{suggestion}` in fourth position. If my arithmetic is right, that should allow for two suggestions from the filter :)

PR #5007, opened 2021-05-14
Minor fixes on MCOMMA antipatterns

palasso:


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/remote-rule-filters.xml
```diff
@@ -73,7 +73,7 @@
                 <pattern>
                     <token regexp="yes">,|\.</token>
                     <marker>
-                        <token postag="CD"/>
+                        <token regexp="yes" spacebefore="no">\d+</token>
```
udomai, line 76: Why `spacebefore="no"`? If the model wants to put a comma between "7000 messages", for example, the AP wouldn't stop it from doing it…

palasso: This specific AP is only for numbers of the form `32,321` or `1.99` but not of the form `, 321` or `. 99`.

PR #5009, opened 2021-05-14
default priority for HydraLEO: -1

fabrichter:
prefer suggestions from more specific rules because of better messages


palasso:
Looks good to me. Thank you!

PR #5034, opened 2021-05-19
[fr] faire vppa rule improved

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -18790,6 +18791,18 @@ though, and has since been slightly modified)-->
                 <example>Je regarde la <marker>télé</marker>.</example>
                 <example>Je compte les <marker>invités</marker>.</example>
             </rule>
+            <rule default="temp_off">
+                <pattern>
+                    <token postag="V.*" postag_regexp="yes"/>
```
jaumeortola, line 18796: Perhaps this is too general, and we need to specify some verbs: finir, continuer etc.

PR #5047, opened 2021-05-20
[gl] clitics - rule completed

xosecalvo:
Rule CLITICS / Start of sentence is complete (antipatterns, url, short)


xosecalvo:
Looking at the diff I've noticed that entities and other references have expanded. This is not what I see in my local copy of the file. What's happened?

Issue #4585, opened 2021-03-10
[en] Wrong tagging CONTRACTIONS


MikeUnwalla:
CONTRACTIONS tags Sha and Ca as NNP:

- [ ] Shan't is almost never used in AmE; rather, it is almost invariably replaced by won't or am/are/is not going to.
- [ ] Can't "ditto" Credit on that one.
- [ ] Can't, I'm going to a recruiting dinner.

jaumeortola:
It is easy to do it with a simple disambiguation rule. I will do it.

PR #5058, opened 2021-05-21
Fix for issue #4585

l3043Y:


jaumeortola:
We can remove the undesired tags in `shan't` and `can't` with a simple disambiguation rule, and I think it is better to do it this way. 

We prefer not to add complexity to the tokenizer. It is critical for us that the tokenizer is as fast as possible.

jaumeortola:
Fixed here: https://github.com/languagetool-org/languagetool/commit/086385ba520ea5e35453759baa34aafcdf28370f

l3043Y:
Thank you for your comment. I hope to come up with elegant fixes later on, and minimal function complexity.

jaumeortola:
Thank you for your work.

PR #5101, opened 2021-05-31
[en] improved BE_PART_AGREEMENT_2

AzadehSafakish:
Scrapped the whole thing and started from scratch.
default="temp_off"


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -51559,336 +51559,170 @@ USA
...
-                    <token regexp="yes">I|you|we</token>
-                    <token regexp="yes" inflected="yes">be|think</token>
+                    <token chunk_re="E-NP.*" />
+                    <token chunk="B-ADVP" />
                 </antipattern>
                 <antipattern>
-                    <token>is</token>
-                    <token regexp="yes">I|you|we|they</token>
+                    <token chunk_re="E-NP.*" />
+                    <token regexp="yes">.*'d</token>
```
udomai, line 51574: `'d` is always going to be tokenized as its own token – so this AP cannot match…

PR #5103, opened 2021-05-31
[en] deactivate TAG_QUESTIONS_2

AzadehSafakish:
Doing poorly, can't find it on Matomo. It's a huge, clunky rulegroup for an error that people are unlikely to make.
Do I have to default="off", or can I just delete the code?


udomai:
If you plan on coming back to it in the future, you can leave it `off`. If you're not going to recycle this code, you can delete it.

PR #5130, opened 2021-06-02
store SpellingCheckRule.wordsToBeIgnoredDictionary* as sorted arrays …

donnerpeter:
…to reduce memory usage

80->74 MB for GermanyGerman

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/spelling/SpellingCheckRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/spelling/SpellingCheckRule.java
@@ -18,7 +18,7 @@
  */
 package org.languagetool.rules.spelling;
 
-import gnu.trove.THashMap;
+import com.google.common.base.Strings;
 import gnu.trove.THashSet;
 import org.apache.commons.lang3.tuple.Pair;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


languagetool-language-modules/de/src/test/java/org/languagetool/rules/de/AgreementRuleTest.java
```diff
@@ -54,7 +54,7 @@ public void testCompoundMatch() throws IOException {
     assertBad("Doch dieser sehr kleine Magnesium Anteil ist entscheidend.", "dieser sehr kleine Magnesiumanteil", "dieser sehr kleine Magnesium-Anteil");
     assertBad("Die Standard Priorität ist 5.", "Die Standardpriorität", "Die Standard-Priorität");
     assertBad("Die derzeitige Standard Priorität ist 5.", "Die derzeitige Standardpriorität", "Die derzeitige Standard-Priorität");
-    assertBad("Ein neuer LanguageTool Account", "Ein neuer LanguageTool-Account");
+    assertBad("Ein neuer LanguageTool Account", "Ein neuer LanguageToolaccount", "Ein neuer LanguageTool-Account");
```
donnerpeter, line 57: This doesn't look right at all, but that follows from the code's logic. `HunspellRule#addIgnoreWords` adds this `LanguageTool` word. Previously `updateIgnoredWordDictionary` wasn't called after that, so it wasn't considered in `startsWithIgnoredWord`, which looks like a bug to me. Now `startsWithIgnoredWord` always uses all ignored words, and the behavior is changed. It probably should be addressed in some way. If you believe that previous behavior was correct and Hunspell's ignored words shouldn't be checked as prefixes, I can adapt the patch to that.

donnerpeter:
Build failed with "no space left on device". Probably not this change's fault?

languagetool-language-modules/de/src/test/java/org/languagetool/rules/de/AgreementRuleTest.java
danielnaber: Indeed it doesn't look right, but I don't have time to look into this now.

danielnaber:
> Build failed with "no space left on device". Probably not this change's fault?

restarting the build seems to have fixed it

languagetool-language-modules/de/src/test/java/org/languagetool/rules/de/AgreementRuleTest.java
donnerpeter: OK, I'll work around that

donnerpeter: Done. Now there are no regressions in tests.

PR #5138, opened 2021-06-03
Update english-homophones.txt

seananocoistin:
Added the word ababúna as a translation of wow.


jimregan:
That file isn't for translations, though; it's for homophones, which in this case is 'bhabh'

jimregan:
...which is already there

PR #5139, opened 2021-06-03
Update about moilligh

seananocoistin:
Inserting the rule to change the error moilligh síos to moilligh.

PR #5140, opened 2021-06-03
Two more rules about style errors

seananocoistin:
Inserting the rule to change the error dúnadh síos to dúnadh.


jimregan (commented):
It looks like you want `<suggestion>\1</suggestion>` instead of `<suggestion>\1 \3</suggestion>` (which can't work: `\3` means 'the third `<token>` in `<pattern>`, and there are only two)

I think you want to use `inflected="yes"` on the first `<token>` in each of these: that would also match, e.g., 'moillithe síos'.

The red X warning about this is probably caused by having the same category multiple times (i.e., the rules have to be within a single `<category>`, but I think I already have a category for collocations -- and wouldn't it be better to consider these redundancies?

jimregan:
Includes #5139

PR #5167, opened 2021-06-07
[ga] ar an aonach/leibhéal (from @seananocoistin; closes #5165 and #5166)

jimregan:


jimregan:
> Did I enter the error in the wrong place? How soon will these be available? I should create another rule about "ar leibhéal". Ar leibhéal is the correct form but often people write ag leibhéal. Ar 07/06/2021 23:35, scríobh Jim Regan:

`<error>` needs to contain the error:

```
<example correction="ar"><marker>ag</marker> an aonach</example>
```

The examples are used in automatic testing, so if the example doesn't match the pattern, the test fails.

> […](#)
> @seananocoistin : <example> needs to contain the _error_

'ag leibhéal' is too broad, as it's valid in too many contexts: see, e.g., https://www.focloir.ie/en/dictionary/ei/level

jimregan:
@seananocoistin - back to the drawing board, I'm afraid.

- focloir.ie 
  - [ag an aonach](https://www.focloir.ie/en/dictionary/ei/ag+an+aonach)
  - [ag an leibhéal](https://www.focloir.ie/en/dictionary/ei/ag+an+leibh%C3%A9al)
- Gaois.ie
  - [ag an leibhéal](https://www.gaois.ie/ga/corpora/monolingual/?Query=ag+an+leibh%C3%A9al&SearchMode=exact&PerPage=50)
  - [ag an aonach](https://www.gaois.ie/ga/corpora/monolingual/?Query=ag+an+aonach&SearchMode=exact&PerPage=50) doesn't turn up too many times, the first two results are Joe Steve Ó Neachtain and Pádraig Standún

That is, the rules are too broad to be enabled by default. If you can think of ways to narrow them down to definitely incorrect contexts, that would help.

PR #5188, opened 2021-06-11
Save memory by interning strings and using trove

donnerpeter:


languagetool-core/src/main/java/org/languagetool/tagging/ManualTagger.java
```diff
@@ -79,7 +79,8 @@ public ManualTagger(InputStream inputStream, boolean internTags) throws IOExcept
         lemma = interned.computeIfAbsent(lemma, Function.identity());
 
         String tag = parts[2].trim();
-        map.computeIfAbsent(form, __ -> new ArrayList<>()).add(new TaggedWord(lemma, internTags ? tag.intern() : tag));
+        String internedTag = internTags ? tag.intern() : interned.computeIfAbsent(tag, Function.identity());
```
donnerpeter, line 82: You might decide that you don't need `internTags` parameter anymore

donnerpeter:
~5MB on de+en

danielnaber:
@donnerpeter This might cause a rare issue documented at https://github.com/languagetool-org/languagetool/issues/5224, do you have any idea?

donnerpeter:
This can be caused by a JMM violation. I'll take a look.

PR #5189, opened 2021-06-11
save memory by reusing EnglishTagger instance

donnerpeter:
~10MB!


jaumeortola:
This can be done in other languages where a tagger is initialized in different places: es, fr, ca...

jaumeortola:
The same idea could be applied to spellers?

donnerpeter:
@jaumeortola Quite likely. And I won't object if you join in on this effort :)

BTW `MorfologikSpeller` already has `dictCache` allowing to reuse the most expensive parts (that I've seen in the snapshots).

Issue #5209, opened 2021-06-16
[el] Absence of redundancy rule


giorgossideris:
There is no redundancy rule in Greek, a rule that I believe that would be useful in the specific language.

giorgossideris:
I will create this rule!

PR #5211, opened 2021-06-17
[el] Create GreekRedundancyRule (#5209)

giorgossideris:
**Example**
image

*You can see more about redundancy [here](https://en.wikipedia.org/wiki/Redundancy_(linguistics)).*


languagetool-language-modules/el/src/test/java/org/languagetool/rules/el/GreekRedundancyRuleTest.java
```diff
@@ -0,0 +1,56 @@
+package org.languagetool.rules.el;
```
danielnaber, line 1: Could you add the standard license header here, too?

danielnaber:
Thanks! Have you locally tested this with a large corpus, to make sure the matches make sense in all/most contexts?

giorgossideris:
> 
> 
> Thanks! Have you locally tested this with a large corpus, to make sure the matches make sense in all/most contexts?
@danielnaber  thank you for the review! Yes I tested it with the wikipedia corpus and no matches were found, that I believe is logical, as carefully written text ought not to contain redundancies (obviously no false positives were detected that is good).

languagetool-language-modules/el/src/test/java/org/languagetool/rules/el/GreekRedundancyRuleTest.java
giorgossideris: Done!

danielnaber:
@giorgossideris Maybe you could also test with https://tatoeba.org, just to be sure? Unlike Wikipedia, it also contains colloquial texts.

giorgossideris:
> 
> 
> @giorgossideris Maybe you could also test with https://tatoeba.org, just to be sure? Unlike Wikipedia, it also contains colloquial texts.

Of course, I will do it now.

giorgossideris:
I tested with the sentences and detailed sentences from https://tatoeba.org/en/downloads and no matches were found neither.

Issue #5227, opened 2021-06-20
[el] Specific-case-rule


Nikos-Antonopoulos:
A Specific case rule for Greek could be quite useful.

Nikos-Antonopoulos:
@Giorgossideris and I will create this!!

danielnaber:
I forgot about this: could you make sure the rule doesn't load the list each time? Depending on the LT configuration, a rule constructor might be called once per check. The constructor should thus be lightweight and not take a long time to initialize.

giorgossideris:
@danielnaber I'll open a new pull request soon to fix that, I didn't think of that and thank you for mentioning it.

PR #5288, opened 2021-07-03
Speed up the SpecificCaseRule (#5227)

giorgossideris:
The loading of the phrases is performed in every initialization, so this commit increases its speed.


giorgossideris:
@danielnaber hopefully this was what you meant.

danielnaber:
Thanks, that looks good.

PR #5310, opened 2021-07-07
[fr] titres en anglais rule improved

Rebecca-auque:


jaumeortola:
Added manually https://github.com/languagetool-org/languagetool/commit/a91e2666af8c61a29c5767ab9e5d5388b0d8059b

PR #5318, opened 2021-07-07
[fr] vous nous non verbe

Rebecca-auque:


jaumeortola:
Added manually https://github.com/languagetool-org/languagetool/commit/41de67772ab577ea6b3f23418d330af0da92a9b5

PR #5330, opened 2021-07-09
[fr] a accent rule improved

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -18172,27 +18172,129 @@ though, and has since been slightly modified)-->
...
-                    <token postag="V.*" postag_regexp="yes">
-                        <exception postag="[ZNRJ].*|V ppr|V.* inf" postag_regexp="yes"/>
-                        <exception regexp="yes">puis|née?s?</exception>
-                        <exception scope="previous" regexp="yes" inflected="yes">qu[ei]</exception></token>
-                    <token postag="D.*" postag_regexp="yes"/>
-                    <token postag="N.*" postag_regexp="yes"/>
-                    <token postag="[JPN].*|C coor" postag_regexp="yes" min="0" max="6"/>
+            <rule default="temp_off">
+            <antipattern>
+                <token>le</token>
```
jaumeortola, line 18177: The antipatterns seem too complex, and perhaps incomplete. I would try this logic: if there is a conjugated verb (`V.*[123].*`) before `a`, then it is probably `à`. But we need to take care of relative clauses (`Ce stade qui peut également accueillir des compétitions d'athlétisme a une capacité de places.`) or interrogative ones (`tu peux voir quelle gueule a ta carte.`)

Rebecca-auque: I don't understand the difference between what you suggest and what I did. The pattern is 'V + GN + a' which becomes 'à' and then I have taken the relative clauses and interrogative ones using antipatterns. Are you suggesting to broaden the pattern by taking out GN?

PR #5345, opened 2021-07-13
[fr] a accent rule improved

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -18323,6 +18339,22 @@ though, and has since been slightly modified)-->
...
+                <token postag="P|UNKNOWN" postag_regexp="yes"/>
+            </antipattern>
+            <antipattern>
+                <token postag="[NJ].*" postag_regexp="yes"/>
+                <token min="0" max="1">,</token>
+                <token>a</token>
+                <token postag="K" skip="1"/>
+                <token min="0" max="1">,</token>
+                <token>sr</token>
+            </antipattern>
```
jaumeortola, line 18357: Again, my impression is that these antipatterns are very specific for some sentences in the diffs, but probably they are not good for improving the results generally.

PR #5363, opened 2021-07-16
Rm17 cest ces

remymartinDev:


jaumeortola:
The rule CES_CEST has been already added in another pull-request, and this can be closed. Is that correct, @remymartinDev

remymartinDev:
This rule had to be corrected due to many FPs, this is the proper PR
I have another PR called "CEST_CETTE", which name sounds similar, but it is a different rule

jaumeortola:
Now I see the differences. Added manually.

PR #5374, opened 2021-07-19
[en] deactivate rules

AzadehSafakish:
Deactivating:

SV_AGREEMENT_NEG_QS - I completely rewrote this rule like, two weeks ago, and users still don't like it. Could be an AAVE thing. Low apply rate with no deactivation contexts.

TOO_TO_EITHER - I feel like I've improved this rule a million times, made it picky, etc. and I just CAN'T be bothered anymore.


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -22094,113 +22094,6 @@ USA
...
-                        <exception postag="JJ" />
-                        <exception>much</exception>
-                    </token>
-                </pattern>
-                <message>Did you mean <suggestion>to</suggestion>?</message>
-                <example correction="to">I hate not having <marker>too</marker>.</example>
-                <example correction="to">I hate not wanting <marker>too</marker>.</example>
-                <example>He isn't trying too hard.</example>
-            </rule>
-        </rulegroup>
```
tiff, line 22203: I don't really see a problem with this rule. Maybe it's rather a style rule than a grammar rule?
Bildschirmfoto 2021-07-19 um 18 43 47

AzadehSafakish: Well then style rule it shall be :)

PR #5396, opened 2021-07-21
[fr] CES_CEST fp correction

remymartinDev:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -8660,6 +8660,11 @@ though, and has since been slightly modified)-->
             <example correction="C'est"><marker>Cet</marker> un garçon !</example>
         </rule>
         <rule id="CES_CEST" name="CES|SES au lieu de C'EST">
+            <antipattern>
+                <token regexp="yes">[cs]es</token>
+                <token regexp="yes">même</token>
```
udomai, line 8665: This is not a `regexp`

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -8660,6 +8660,11 @@ though, and has since been slightly modified)-->
...
+                <token postag="[JN].p" postag_regexp="yes"/>
```
udomai, line 8666: The test fails because the example sentence causes an error. That's because the `postag_regexp` should use `.*` instead of `.` (which matches only exactly one character).

udomai (changes requested):

udomai (approved):

PR #5407, opened 2021-07-26
[fr] p v pas rule improved

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -33643,6 +33664,46 @@ though, and has since been slightly modified)-->
                 <url>http://www.academie-francaise.fr/ne</url>
                 <example correction="Ce n'en est pas"><marker>C'en est pas</marker> une !</example>
             </rule>
+            <rule>
```
udomai, line 33667: Should these two be `temp_off`?

PR #5411, opened 2021-07-27
[fr] useless anglicism + transfer picky tags in proper category

remymartinDev:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -37515,8 +37690,8 @@ though, and has since been slightly modified)-->
                 <token regexp="yes">tâches?</token>
             </pattern>
             <message>« Assigner (quelqu’un) à une tâche » peut être considéré comme un anglicisme.</message>
-            <suggestion>affecteur</suggestion>
-            <example correction="affecteur"><marker>assigner</marker> à une tâche</example>
+            <suggestion>affecter</suggestion>
```
udomai, line 37693: Well spotted, but this rule has more problems: it takes any form of "assigner", but it always only gives "affecter" as a suggestion. Can be resolved with

```xml
<suggestion><match no="1" postag="V.*" postag_regexp="yes">affecter</match></suggestion>
```

to give "affecter" the same form that "assigner" has

udomai (changes requested):

udomai:
From which category to which other category did you move these?

remymartinDev:
> From which category to which other category did you move these?

Indeed, I should have mentioned it. There are 2 "Anglicism" categories : 
⇾ CAT_ANGLICISMES
⇾ CAT_ANGLICISMES_PICKY

all the rules found in CAT_ANGLICISMES with the tag `tags="picky"` have been moved to CAT_ANGLICISMES_PICKY

PR #5413, opened 2021-07-27
[fr] SUREMENT_ACCENT

remymartinDev:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -4240,6 +4240,16 @@ though, and has since been slightly modified)-->
...
+            <pattern>
+                <marker>
+                    <token>surement</token>
+                </marker>
+            </pattern>
+                <message>Ce mot s'écrit avec un accent circonflexe.</message>
+            <suggestion>sûrement</suggestion>
+            <example correction="sûrement">Tu as <marker>surement</marker> raison.</example>
+        </rule>
         <rule id="POLE_2" name="pôle" default="temp_off">
```
jaumeortola, line 4253: "Surement" is "orthographe rectifiée de 1990". See: https://fr.wiktionary.org/wiki/surement
Perhaps we need coherency rules or user settings (new/traditional orthography).

remymartinDev: Should I add "En orthographe traditionnelle (non réformée), attention à l'accent circonflexe." in the message ?
Or should I drop this rule ?

jaumeortola: As far as I know, we just accept all forms (traditional and reformed). I don't know if there are plans for other things. @udomai

remymartinDev:
Rule abandonned -> irrelevant

PR #5447, opened 2021-08-02
call checkCanceled during sentence disambiguation

donnerpeter:


donnerpeter:
To decrease the probability of UI freezes in IntelliJ

PR #5472, opened 2021-08-09
[en] rules to picky/temp off

AzadehSafakish:
@udomai 
Please merge.


udomai:
Alright! If you can't work on a rule in the near future, please set it to `off` rather than `temp_off` (as a general rule of thumb)

PR #5473, opened 2021-08-09
[en] improve SUBJECT_DROP to avoid correction loop

AzadehSafakish:
A rule (MISSING_I) is going live in Premium after the sync.

I don't know for sure, but I would assume it's going to create a correction loop with this rule (inserting an 'I' and then suggesting that it be removed). 

To be cautious this rule will no longer pick up 'I', to avoid any conflict.

@udomai 
Please merge.


udomai:
Good thing you thought of that. Testing this (whether there will be a loop or not) can be pretty hard with the regular tools. There is a pretty comfy way, though, by setting up a local server. Imma show you that next time we meet!

PR #5485, opened 2021-08-09
[fr] MOTS_INCOMP improved

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -35785,6 +35785,41 @@ though, and has since been slightly modified)-->
...
+                <token regexp="yes">[st]on</token>
+                <token postag="V ppa m sp?" postag_regexp="yes"/>
+            </antipattern>
+            <pattern>
+                <token regexp="yes" case_sensitive="yes">[cdlmst]es|[nv]os|cette|ce[st]|[ts]a|ton</token>
+                <marker>
+                    <token postag="V ppa m sp?" postag_regexp="yes"><exception postag="N.*|V (ind|sub|con).*" postag_regexp="yes"/></token>
+                </marker>
+            </pattern>
+            <filter class="org.languagetool.rules.fr.FindSuggestionsFilter" args="wordFrom:2 desiredPostag:N.*"/>
```
udomai, line 35815: It would be perfect if we could tell the `FindSuggestionsFilter` that we're only looking for words that match `\1` in number and gender…

PR #5503, opened 2021-08-11
[fr] s'en suivre

remymartinDev:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -55049,6 +55049,37 @@ though, and has since been slightly modified)-->
...
+        <rulegroup id="S_EN_SUIVRE_2" name="s'en suivre(s'ensuivre)" default="temp_off">
+            <rule>
+                <pattern>
+                    <marker>
+                        <token postag_regexp="yes" postag="R pers obj 3 .*">s'</token>
+                        <token>en</token>
+                        <token postag_regexp="yes" postag="V .*" inflected="yes">suivre</token>
+                    </marker>
+                </pattern>
+                <message>On ne doit pas écrire 's’en suivre' en trois mots car si le verbe 'suivre' peut s’employer en ce sens, ce n’est pas le cas de la forme pronominale 'se suivre'.</message>
```
jaumeortola, line 55061: Could we use a clearer message? For example, ` Au sens de "découler, résulter", utilisez "il s'ensuit" or "il suit"`.

remymartinDev: thank you Jaume, I did not even consider the possibility of proposing "il suit" which is relevant in many cases

PR #5525, opened 2021-08-16
[fr] ABREVIATIONS_ET_PONCTUATION add 'id' to subrules to identify in …

remymartinDev:
…Grafana

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
@@ -50661,7 +50661,7 @@ though, and has since been slightly modified)-->
             <example>autoentrepreneur et microentrepreneur</example>
         </rule>
         <rulegroup id="ABREVIATIONS_ET_PONCTUATION" name="Abréviations et ponctuation">
-            <rule> <!-- 1 / CF -->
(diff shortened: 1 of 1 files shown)


udomai:
This is technically not wrong, but giving IDs to subrules has the consequence that they are no longer real subrules.
The consequences of this are subtle: Users might no longer be able to deactivate the rulegroup in one click; we will probably not see these rules' statistics in the Matomo/Grafana data for the rulegroup…

Ergo: This doesn't break anything, but if you can avoid giving your subrules their own IDs, we avoid a few complications.

remymartinDev:
I understand
I found 'open' statistics very low in Grafana for this rulegroup, I wondered if 1 subrule in particular was lowering the rate, then I thought it could be a good way to identify it
PR closed

udomai:
I see! Open rates are an indicator of how often the rule matches, not of how popular it is. The number of matches in the nightly regression test is an indicator (although not 100% reliable) of that. Maybe that could be sufficient?

The open rate per subrule is not available via Grafana, but the deactivation rate is, which is more important. At some point, we might figure out a way of tracking the open rate per subrule as well.

remymartinDev:
Sorry, I used the wrong term, I wanted to talk about 'apply' rate, not 'open'
there are currently 188 Applies on 432 Opens (43.52%)

udomai:
That is too low and hard to track. But normally, the deactivations are directly correlated with low applications.

In Grafana, in the last 7 days, there is only data about subrule [2] – so this is the only subrule that matched. That's where all the action seems to be.
(In the last 30 days, there are a few deactivations for [1] as well, by the way.)

PR #5563, opened 2021-08-24
[core] fix disappearing quotes CommaWhitespaceRule

udomai:
Problem:

> The coffee maker doesn't work anymore after you " fixed" it.

CommaWhitespaceRule deletes the quote that has spaces on both sides, as well as the space to its right. (result: `you fixed" it.`)

The change I made in line 108 works for spaces after opening quotes (result: `you "fixed" it.`), but has an undesired outcome for spaces before closing quotes:

> The coffee maker doesn't work anymore after you "fixed " it.

result: `you "fixed "it.`

### DIFF
--- a/languagetool-core/src/test/java/org/languagetool/rules/CommaWhitespaceRuleTest.java
+++ b/languagetool-core/src/test/java/org/languagetool/rules/CommaWhitespaceRuleTest.java
@@ -95,7 +95,13 @@ public class CommaWhitespaceRuleTest {
     matches = rule.match(lt.getAnalyzedSentence("This , is a test sentence."));
     assertEquals(1, matches.length);
     assertEquals(",", matches[0].getSuggestedReplacements().get(0));
-    
+
+    matches = rule.match(lt.getAnalyzedSentence("You \" fixed\" it."));
+    assertEquals(1, matches.length);
+    assertEquals("\"", matches[0].getSuggestedReplacements().get(0));
+    matches = rule.match(lt.getAnalyzedSentence("You \"fixed \" it."));
+    assertEquals(1, matches.length);
+    assertEquals("\"", matches[0].getSuggestedReplacements().get(0));
 
     assertMatches("Ellipsis . . . as suggested by The Chicago Manual of Style", 3);
     assertMatches("Ellipsis . . . . as suggested by The Chicago Manual of Style", 4);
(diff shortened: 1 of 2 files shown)


udomai:
The two cases I added to the test don't cause any problems. But I can't seem to tell the test where the space I want to delete is.

udomai:
What we could do is: extend the match to the space before the quote and the space after, and have two suggestions: (1) the quote with the space after (but not before) and (2) the quote with the space before (but not after).

Unless we teach the rule to look for other quotes in the text surrounding it, this seems the easiest way.

languagetool-core/src/test/java/org/languagetool/rules/CommaWhitespaceRuleTest.java
```diff
@@ -95,7 +95,17 @@ public void testRule() throws IOException {
...
+    matches = rule.match(lt.getAnalyzedSentence("You \" fixed\" it."));
+    assertEquals(1, matches.length);
+    assertEquals("\"", matches[0].getSuggestedReplacements().get(0));
+    assertEquals(4, matches[0].getFromPos());
+    assertEquals(6, matches[0].getToPos());
+    matches = rule.match(lt.getAnalyzedSentence("You \"fixed \" it."));
+    assertEquals(1, matches.length);
+    assertEquals("\"", matches[0].getSuggestedReplacements().get(0));
+    assertEquals(11, matches[0].getFromPos());
+    assertEquals(13, matches[0].getToPos());
```
udomai, line 108: This should be 10–13 if we match both spaces. The current value is not good, as it attaches the quote to the following word instead of the one before.

languagetool-core/src/test/java/org/languagetool/rules/CommaWhitespaceRuleTest.java
```diff
@@ -95,7 +95,17 @@ public void testRule() throws IOException {
     matches = rule.match(lt.getAnalyzedSentence("This , is a test sentence."));
     assertEquals(1, matches.length);
     assertEquals(",", matches[0].getSuggestedReplacements().get(0));
-    
+
+    matches = rule.match(lt.getAnalyzedSentence("You \" fixed\" it."));
+    assertEquals(1, matches.length);
+    assertEquals("\"", matches[0].getSuggestedReplacements().get(0));
+    assertEquals(4, matches[0].getFromPos());
+    assertEquals(6, matches[0].getToPos());
```
udomai, line 103: This one should be 3–6 if we matched both spaces.

jaumeortola:
I resolved the problem. But there is a further improvement that can be made. The quotation marks that have a direction (left, right) could be treated as brackets (left and right). It seems straightforward. Is there any unforeseen problem there?

jaumeortola:
Done here https://github.com/languagetool-org/languagetool/commit/4870d8ed9fda1c0ad37c349bbbf7dbf1b207f03d

PR #5573, opened 2021-08-25
[fr] remove "éprouvè" from french.dump

remymartinDev:
I've seen this error in french.dump that is providing false results in premium diff
Removing this from french.dump, and I'll check tomorrow whether it is the root cause of false results


udomai:
What is the problem caused by this? All verbs of the first group have this form, and it is correct:

"Qu'éprouvé-je ?" can be spelled "Qu'éprouvè-je?" after the réforme de l'orthographe, so it is indeed a `V ind pres 1 s`

udomai:
@jaumeortola, can you save the day? Example: https://regression.languagetoolplus.com/via-http/2021-08-25/fr/result_grammar-premium4_VERBES_PAUVRES[3].html # 26: "ai" (`V.* ind pres 1 s`) is getting two forms of the lemma "éprouver": "éprouve" (correct) and "éprouvè". Didn't you have a solution for those verb forms no longer being suggested?

PR #5614, opened 2021-09-01
[fr] MOTS_INCOMP improved

Rebecca-auque:


jaumeortola:
I get possible FPs here: https://community.languagetool.org/ruleEditor/expert?devMode=true Have you checked it? Make it "temp_off".

Rebecca-auque:
Indeed, there were FPs, it's corrected. It is already temp_off as the rulegroup is temp_off.

jaumeortola:
The whole rule group is temp_off. Then it is fine.

PR #5619, opened 2021-09-01
suppress misspelled for RemoteRules

fabrichter:
add options to RemoteRule to filter out matches (`suppressMisspelledMatch`) or suggestions with misspellings (`suppressMisspelledSuggestions`)
allow application to only selected rule IDs via regex
added getDefaultSpellingRule() to Language

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/Language.java
+++ b/languagetool-core/src/main/java/org/languagetool/Language.java
@@ -29,6 +29,7 @@ import org.languagetool.rules.RemoteRuleConfig;
 import org.languagetool.rules.Rule;
 import org.languagetool.rules.neuralnetwork.Word2VecModel;
 import org.languagetool.rules.patterns.*;
+import org.languagetool.rules.spelling.SpellingCheckRule;
 import org.languagetool.synthesis.Synthesizer;
 import org.languagetool.tagging.Tagger;
 import org.languagetool.tagging.disambiguation.Disambiguator;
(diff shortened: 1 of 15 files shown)


jaumeortola:
`getDefaultSpellingRule()` is great. I will use it in other places. 

A comment about the method `SpellingCheckRule.isMisspelled()`. It doesn't tokenize the input string. So, the results could be wrong or imprecise in hyphenated compounds or multiwords.  

The spelling rules use the dictionaries. I don't know if we could save memory doing something similar to this: https://github.com/languagetool-org/languagetool/blob/4f3d2b090eb2be87afad5f7ef98f80c2ba81b431/languagetool-language-modules/en/src/main/java/org/languagetool/tagging/en/EnglishTagger.java#L38

jaumeortola:
> What would you recommend to use instead of isMisspelled? The place where I am calling it doesn't have access to an AnalyzedSentence for the relevant input.
One alternative is to do the full analysis (tokenizer, tagger...) and call "match()". I don't know if it is worthwhile. Do you expect to have "multi words" (or "split words") or "hyphenated-compounds" in the suggestions?

Alternatively, with "multi words" you could use isMisspelled() for every single word, and accept it if evey word is correct. With "hyphenated-compounds" you could check the whole word and its components (but the criteria for accepting the words are not clear-cut).

fabrichter:
I'm now calling match() instead of isMisspelled and I'm using getRawAnalyzedSentence() to only tokenize the sentence.

PR #5651, opened 2021-09-09
[fr] SI_CI_SY & SI_CI & CI improvement + merge

remymartinDev:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -36059,56 +36007,6 @@ though, and has since been slightly modified)-->
             <example correction="y avait">Qui <marker>y-avait</marker>-il cette nuit-là ?</example>
             <example correction="y a">Combien <marker>y-a</marker>-t-il de zoos en Australie?</example>
         </rule>
-        <rulegroup id="CI" name="ci joint → ci-joint (etc.)">
-            <rule>
```
jaumeortola, line 36063: This rule seemed fine to me. Has it FPs? Are all the errors detected by the new rules?

remymartinDev: I copied 2 or 3 subrules from this rulegroup (and also from SI_CI) into SI_CI_SY, but a huge amount of [FP's were generated](https://internal1.languagetool.org/regression-tests/via-http/2021-09-09/fr/result_grammar_SI_CI[3].html) by the subrule considering any Name followed by "SI", to be replaced by "Name-ci"
_Regarde cet arbre si, qui ressemble à un chêne_ -> TP
_Regarde cet arbre si il ressemble à un chêne_ -> FP
I thought about a way to improve it, for example by adding <token>,</token> after si, but this would also generate false positives like in :
_M'auriez-vous posé cette question si, par exemple,  j'avais été un homme_ 
Considering the impossibility of improving it, and seeing that the 60 examples in the diff are all FP, that is why I prefered to delete it. (But I kept the original rulegroup in a draft file, in case I think about a way to treat that in the future)

Also, the original "CI" rulegroup contained errors in the subrules **"CI" followed by a list of word**, this list included `ceux|celles?|celui|par` [that were incorrect](https://internal1.languagetool.org/regression-tests/via-http/2021-09-09/fr/result_grammar_SI_CI[1].html), correcting "si ceux" by "ci-ceux" that does not exists ("ceux-ci" exists, but it was taken into account by a subrule in SI-CI_SY)

PR #5665, opened 2021-09-13
[fr] ADJ_INVARIABLE improved

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/added.txt
```diff
@@ -2140,4 +2140,5 @@ Kâmasûtra	Kâmasûtra	N m s
 Naruto	Naruto	Z m sp
 auto-tamponneuse	auto-tamponneuse	N f s
 autos-tamponneuses	auto-tamponneuse	N f p
-top	top	J e sp
\ No newline at end of file
+top	top	J e sp
+extra	extra	J e sp
```
jaumeortola, line 2144: It can be also "extras". https://fr.wiktionary.org/wiki/extra#Adjectif

Rebecca-auque: Apparently yes! I only checked on Larousse but Le Robert says likewise that it can be plural. So Larousse and I must be a bit old school then !

jaumeortola: Then, can we just remove the rule?

Rebecca-auque: Yes, I just thought that adding 'top' as J e sp in the added file could not be enough in order to correct a sentence like 'Ce sont des filles tops.'

jaumeortola: Okay. We'll keep the rule for "tops/top" but not for "extras/extra".

PR #5672, opened 2021-09-14
[fr] ACCORD_COULEUR improved

Rebecca-auque:


jaumeortola:
Thanks. We could put the lists of colors and adjectives in XML entities, so they can be reused in different rules. Currently we have different lists on different rules. We should unifiy them.

PR #5679, opened 2021-09-15
Language Resource module per each Language module

KratochvilLukas:
We have encounter a licensing issue, that prevents us from using the LanguageTool library. Some resouce files are covered by a license other than LGPL as LanguageTool code itself. 

We have found a solution and it's implementation is in this PR. Every Language module has its own resources module. This module is included by default, but there is a possibility to exclude it and add them by our own way. 

It would be nice to get it into the code or implement some other solution that allows us not to use resources that are not under LGPL license. 

Thanks


KratochvilLukas:
Because of frequency of fixes into /resources directories there will be conflicts almost everytime.  We can use this PR thread for discussion before finally resolve all conflits.

jimregan:
You should consider talking to an actual legal professional before deciding you have a licensing issue, but at the very least you should try to understand what a [derivative work](https://en.wikipedia.org/wiki/Derivative_work) is and *isn't*, and, as I'm guessing you're specifically looking at dictionaries that use the GPL, read [section 5](https://www.gnu.org/licenses/gpl-3.0.html#section5) of the GPL3 and/or [section 2](https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html#section2) of the GPL2.

KratochvilLukas:
Source code for languages is OK. It's good to have as many languages as possible supported by you and your community. The only problem is with resources undes GPL or onther license different from LGPL. We cannot distribute these files together with application.

jimregan:
> Source code for languages is OK. It's good to have as many languages as possible supported by you and your community. The only problem is with resources undes GPL or onther license different from LGPL. We cannot distribute these files together with application.

Yes, you can. The GPL2 calls this "mere aggregation", the GPL3 calls them "aggregates".

Again, talk to an actual lawyer.

KratochvilLukas:
> > Source code for languages is OK. It's good to have as many languages as possible supported by you and your community. The only problem is with resources undes GPL or onther license different from LGPL. We cannot distribute these files together with application.
> 
> Yes, you can. The GPL2 calls this "mere aggregation", the GPL3 calls them "aggregates".
> 
> Again, talk to an actual lawyer.

We solve it on the basis that historically we were solving similar issue with hunspell dictionaries. It was consulted with lawyer too.  But you're right. I'll get other info from more authorized person.

jezbera:
Hi @jimregan,
legal documents are usually ambiguous. According to this
https://www.gnu.org/licenses/gpl-faq.html.en#MereAggregation
I don't think, that program using jar with GPL dictionaries is aggregate.

> If modules are designed to run linked together in a shared address space, that almost surely means combining them into one program.

jimregan:
> Hi @jimregan,
> legal documents are usually ambiguous.

Sure; also, what one has not fully understood can seem more ambiguous than it truly is.

> According to this
> https://www.gnu.org/licenses/gpl-faq.html.en#MereAggregation
> I don't think, that program using jar with GPL dictionaries is aggregate.
> 
> > If modules are designed to run linked together in a shared address space, that almost surely means combining them into one program.

Key phrases: run _linked together_ in a shared _address space_ - so, what happens when you link pieces of software together, and very much **not** what happens when you load a dictionary into a piece of software.

A derivative work is one that's _based on_ another, not one that merely includes it. If you did a techno version of a pop song, that techno version would be a derivative. Bundling the pop song ("aggregation") on the MP3 player, however, **does not** make the MP3 player a derivative of the song.

bammalus:
Hi @jimregan, I will not argue with you and I appreciate your comments. But our lawyers said no to this case and that is the end for us. That's why we seek another way how to use  LanguageTool in our application/service.

jimregan:
> Hi @jimregan, I will not argue with you and I appreciate your comments. But our lawyers said no to this case and that is the end for us. That's why we seek another way how to use LanguageTool in our application/service.

Well, I know it can be just as difficult to explain to a lawyer the difference between a class and a resource file as it is to explain to a programmer the difference between a derivative and an aggregation. And step one is knowing what it is that needs to be explained, which I somehow doubt is what ended up happening.

Putting all that aside, this is still excessive. What drew my eye to this in the first place was seeing that `language-ga` had been moved, as I _know_ there are no non-LGPL components in it. Those are already in a separate repo entirely. If you could replicate that across the other language modules, that would be much less disruptive, likely of benefit to languagetool generally, but if nothing else, certainly more convincing than not really making any case at all.

PR #5690, opened 2021-09-16
[ru] remove deprecated scripts

yakovru:
[ru] remove deprecated scripts


yakovru:
[ru] remove deprecated scripts

PR #5691, opened 2021-09-17
[fr] P_V_PAS improved

remymartinDev:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -38141,36 +38131,36 @@ though, and has since been slightly modified)-->
             <rule>
                 <antipattern>
                     <token regexp="yes">fau[xt]</token>
-                    <token regexp="yes">pas|jamais|guère|aucune?s?|personne</token>
+                    <token regexp="yes">pas|aucun(e(ment)?)?s?|jamais|guère|nulle(ment)?|plus|rien|personne</token>
                 </antipattern>
```
jaumeortola, line 38135: This regular expression is repeated many times. It would be useful to have it in an XML entity, so it can be edited and re-used more easily.

PR #5709, opened 2021-09-20
[fr] A_A_ACCENT improved

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -9195,11 +9195,12 @@ though, and has since been slightly modified)-->
                 <token>
                     <exception regexp="yes">pour|contrario|fortiori|posteriori|priori|minima</exception>
                     <exception spacebefore="no">)</exception>
-                    <exception postag="V.* ppa.*" postag_regexp="yes"/></token>
+                    <exception postag="V.* ppa.*|UNKNOWN" postag_regexp="yes"/></token>
```
jaumeortola, line 9198: This has pros and cons. I am not convinced that we should do it.

PR #5731, opened 2021-09-23
[fr] add AVERER

remymartinDev:
I've seen that the app was correcting some uses of "avérer" with AVERE(premium1) instead of SAVERER_FAUX(open source), and some cases were not covered
I commented SAVERER_FAUX(os), SAVERER_VRAI(os) , AVERE(premium1) until the new merged rulegroup has been confirmed by night diff


udomai:
There's something wrong with the code, and the CI test result is unspecific.

It could be that there cannot be comment within a comment: `<!-- comment <!-- comment 2 --> -->` will break the build.

remymartinDev:
Again, I assume that I ran MVNtest on the wrong version. 
I'm trying to focus on avoiding this kind of errors

PR #5736, opened 2021-09-24
[fr] ADJ_ADJ_SENT_END improved

Rebecca-auque:


languagetool-language-modules/fr/src/main/resources/org/languagetool/resource/fr/added.txt
```diff
@@ -62,3 +62,4 @@ débunkées	débunker	J f p
 débunkées	débunker	V ppa f p
 débunkés	débunker	J m p
 débunkés	débunker	V ppa m p
+végane	végane	J e s
```
jaumeortola, line 65: But the dictionaries say "végan (masc.) végane (fem.)".

PR #5760, opened 2021-10-01
[en] add to dictionary for Premium issues 3121, 3192

AzadehSafakish:
@udomai 

Still fails and idk why :/


udomai:
So, I've checked out this branch locally, and in `added.txt`, they weren't tabs between the word, the lemma and the postag, but spaces (that, suspiciously, had the exact width as the tab would have had (1, 2, or 3 spaces)).

Maybe it is a setting in IntelliJ that makes spaces instead of a tab character?

udomai:
Does this happen even if you copy-paste the tab character? Please try once more in your next `added` addition 🙂

PR #5834, opened 2021-10-14
[de] remove obsolete SUB:DAT:SIN:MAS forms

udomai:
Test results:

[diff-results-2021-10-14T14 13.zip](https://github.com/languagetool-org/languagetool/files/7347554/diff-results-2021-10-14T14.13.zip)


udomai:
I'll do another PR with the spelling so the CI test doesn't fail

PR #5868, opened 2021-10-19
Fix Rules Match Count

xavivars:
Match Count logged stopped working after https://github.com/languagetool-org/languagetool/commit/1f6260a35b5df471414de619f2f5fa7d92a16799

What was happening instead (on open-source installs) is that it was always logged `0/1 matches`, because premium was off and also res was a list of 1 CheckResult


xavivars:
That was fast! 😄

PR #5966, opened 2021-11-01
allow tab or semicolon in files: multiword.txt, added.txt, removed.txt

jaumeortola:
Editing manually tab-separated files is annoying and frequently a waste of time because of errors and confusions. 
I suggest allowing both tabs and semicolons. It is not "beautiful", but it can be useful. In the future, we can replace all tabs with semicolons in all files.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/synthesis/ManualSynthesizer.java
+++ b/languagetool-core/src/main/java/org/languagetool/synthesis/ManualSynthesizer.java
@@ -85,9 +85,9 @@ public final class ManualSynthesizer {
           continue;
         }
         line = line.replaceFirst("#.*", "").trim();
-        String[] parts = line.split("\t");
+        String[] parts = line.split("[\t;]");
         if (parts.length != 3) {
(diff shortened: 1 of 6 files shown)


jaumeortola:
Unfortunately, the semicolon is used inside Arabic POS tags. We need to use another character, or make the separator configurable.

udomai:
Argh! I guess `#` and `%` are out of the question. Is `~` accessible enough on EN/ES keyboards? It's okay in DE & FR, I guess (ALT GR + + in DE and ALT GR + 2 in FR)

jaumeortola:
For me, `~` is ALT GR+4. Anyway, I am going to make it configurable, just in case it is needed in the future: 
`[\t;]` by default and `[\t]` for Arabic.

jaumeortola:
Now, a custom separator can be added to the file itself:
`#separatorRegExp=[\t;]`
The default is still the tab. 
See: https://github.com/languagetool-org/languagetool/commit/4c61abf0a79362f4d4ead02c26ea5753897c0ace

PR #5976, opened 2021-11-04
stop suggestion generation for too many errors

fabrichter:


SteVio89:
Is it possible or useful to take the MAX_ERROR_RATE value from properties? We already have maxErrorsPerWordRate.

fabrichter:
I don't see an easy/good way to pass this value here, and this is intended as a temporary workaround as we're moving away from using this class altogether.

PR #5977, opened 2021-11-04
[core] add error rate calc after each rule check based on hole text size

SteVio89:
[core] add logging to see how error rate is growing in case of an error


fabrichter:
s/hole/whole/; otherwise, looks good to me

PR #5987, opened 2021-11-08
[en] add one simple AP to HYDRA_LEO_ARE_IS as a test

AzadehSafakish:
@udomai 
@s-burst 

I have no clue if I'm doing this correctly, so I'm opening a PR with a super simple test AP.
There were a few disable contexts for 'thou', and even at the threshold that it's currently at, the rule suggests 'is' in this sentence:

> Some scholars rely on John 8:57: "thou are not yet fifty years old", making the earliest possible year for Jesus's birth c. 18 BC.

I have a lot more APs I want to write, one of them being to ignore things in quotes. 
So, this example will be subsumed into that anyway. But it is a test AP for now.

If this looks correct, please merge and I will add the rest. Thanks!


udomai:
Looks good! Possible pitfalls would have been:

- adding a rule(group) id="AI_HYDRA_LEO_CPT_ARE_IS" even though there already is one
- problems with the example (would have shown up when running the test in EnglishRemoteRuleFilterTest.java, line 35)

PR #6027, opened 2021-11-16
[core] add MAX_SUGGESTIONS to MorfologikMultiSpeller

SteVio89:


jaumeortola:
A potential problem is that we loose words that after resorting would get to the first positions. Have you checked that?

SteVio89:
There is a “Collections.sort(result);” before I create the sublist of reduced suggestions, but if you know a better place to reduce the suggestions, I am excited to see your idea.

There is already a similar implementation here https://github.com/languagetool-org/languagetool/blob/master/languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/CompoundAwareHunspellRule.java#L147

fabrichter:
LGTM

PR #6035, opened 2021-11-18
[en] improve EN_REPEATEDWORDS_NEED

AzadehSafakish:
Added a first test AP for 'need to', as 'require to' is an odd suggestion:

I still need to sign-in somewhere. You need to sign-in too. →
I still need to sign-in somewhere. You **require** to sign-in too.

However, I couldn't get tokenRegex("need(s|ed|ing)?") or even tokenRegex("need|needs|needed|needing") to work, so it currently only works for VBP.

Questions, comments, concerns, tips?
@jaumeortola @udomai

### DIFF
--- a/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/EnglishRepeatedWordsRule.java
+++ b/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/EnglishRepeatedWordsRule.java
@@ -44,7 +44,7 @@ public class EnglishRepeatedWordsRule extends AbstractRepeatedWordsRule{
 
   private static final List<List<PatternToken>> ANTI_PATTERNS = Arrays.asList(
     Arrays.asList(
-      token("need"),   // "I still need -> require to sign-in"
+      tokenRegex("need(s|ed|ing)?"),   // "I still need -> require to sign-in"
       token("to")
     ));
 
--- a/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/EnglishRepeatedWordsRule.java
+++ b/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/EnglishRepeatedWordsRule.java
@@ -44,7 +44,7 @@ public class EnglishRepeatedWordsRule extends AbstractRepeatedWordsRule{
 
   private static final List<List<PatternToken>> ANTI_PATTERNS = Arrays.asList(
     Arrays.asList(
-      tokenRegex("need"),   // "I still need -> require to sign-in"
+      token("need"),   // "I still need -> require to sign-in"
       token("to")
     ));
 
--- a/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/EnglishRepeatedWordsRule.java
+++ b/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/EnglishRepeatedWordsRule.java
@@ -18,22 +18,44 @@
  */
 package org.languagetool.rules.en;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 import java.util.ResourceBundle;
+import java.util.function.Supplier;
 
 import org.languagetool.AnalyzedTokenReadings;
 import org.languagetool.language.AmericanEnglish;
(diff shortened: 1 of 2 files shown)


jaumeortola:
For me, tokenRegex is working. The sentence "It needs to be done. That needs to be done." currently marked as incorrect, becomes correct. Isn't it the expected result?

languagetool-language-modules/en/src/test/java/org/languagetool/rules/en/EnglishRepeatedWordsRuleTest.java
```diff
@@ -49,6 +49,7 @@ public void testRule() throws IOException {
     assertCorrectText("This is a new experience. Happy New Year!");
     assertCorrectText("This was needed. There is a need to do it.");
     assertCorrectText("It needs to be done. That also needed to be done.");
+    assertCorrectText("I still need to sign-in somewhere. You need to sign-in too.");
```
udomai, line 52: I think it should be "sign in", just as a side note 🤓

languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/EnglishRepeatedWordsRule.java
```diff
@@ -18,22 +18,44 @@
...
 
 public class EnglishRepeatedWordsRule extends AbstractRepeatedWordsRule{
   
   private static final EnglishSynthesizer synth = new EnglishSynthesizer(new AmericanEnglish());
 
+  private final Supplier<List<DisambiguationPatternRule>> antiPatterns;
+
+  private static final List<List<PatternToken>> ANTI_PATTERNS = Arrays.asList(
+    Arrays.asList(
+      token("need"),   // "I still need -> require to sign-in"
```
udomai, line 47: you could make this line

```java
new PatternTokenBuilder().token("need").posRegex("VB.*").matchInflectedForms().build(),
```

but

```java
new PatternTokenBuilder().tokenRegex("need(s|ed|ing)?").build(),
```

should work, too

udomai (changes requested):

languagetool-language-modules/en/src/test/java/org/languagetool/rules/en/EnglishRepeatedWordsRuleTest.java
udomai: With one of the solutions in the other comment, this sentence should be correct, too:

```
assertCorrectText("I still needed to sign-in somewhere. She needs to sign-in too.");
```

udomai:
By the way, @jaumeortola, can you confirm the rule matches all forms of "need" (matching the tag and chunk specified in synonyms.txt)? Just making sure.

AzadehSafakish:
> 
> 
> For me, tokenRegex is working. The sentence "It needs to be done. That needs to be done." currently marked as incorrect, becomes correct. Isn't it the expected result?

Yes, that is the expected result. I'm not sure what I'm missing, but when I try it, I believe it no longer catches the error in the test sentences (...)

[ERROR] Failures: 
[ERROR]   EnglishRepeatedWordsRuleTest.testRule:80 expected:<1> but was:<0>

languagetool-language-modules/en/src/test/java/org/languagetool/rules/en/EnglishRepeatedWordsRuleTest.java
AzadehSafakish: Perhaps?
I'll drop the hyphen in 'sign-in' once you drop the hyphen in 'extra-careful' XD
What even is that emoji...

udomai:
Line 80 tests the sentence two lines above it: "It needs to be done. That needs to be done." After you insert the AP, this sentence will no longer be caught – i.e., this sentence needs to be changed to something your AP allows.

jaumeortola:
Merged manually.

jaumeortola:
https://github.com/languagetool-org/languagetool/commit/01323a4c23a105841701ddd3758931ad9d6aef0d

AzadehSafakish:
Wait, I see. I was looking in the wrong place.
Thank you so much @jaumeortola !!

Issue #2288, opened 2020-01-04
Case sen sitive Simple Replace


baarsrj:
It would be great if simple replace could be made (partly) case sensitive. By this I mean:
- detect if there is uppercase in part before =
- if so, change behaviour to suggest check left part case sensitively, and suggest the right parts the same; except for the first token when this is at the start of the sentence

anmol1vw13:
Hello @baarsrj 
I'm new to opensource and saw this project on code triage. I would like to contribute to this project. The only challenge here is I need to understand the code flow. I'm very hands-on with Java but have no knowledge of NLP. It would be a great help if you could guide me.

PR #6127, opened 2021-12-04
Case sensitive simple replace

feliciawang08:
This PR aims to contribute to [issue 2288](https://github.com/languagetool-org/languagetool/issues/2288) regarding making SimpleReplace partly case sensitive. Instead of directly changing the already existing functionality for Simple Replace rule within the AbstractSimpleReplaceRule2 class, I add a separate public method that attempts to follow the partly case sensitive route. This serves only as an idea for the above issue, not a permanent replacement.


jaumeortola:
For checking case (and only case), we are using `AbstractCheckCaseRule extends AbstractSimpleReplaceRule2`. So, this change is probably redundant.

Issue #5518, opened 2021-08-13
[en] Missing POS for 'long'


MikeUnwalla:
The POS (RB, RBS, RBR) for the adverbs long/longer/longest are missing.

https://www.lexico.com/definition/long
https://www.merriam-webster.com/dictionary/long

(Adding the POS causes test failures in Premium, and I don't have time to deal with them.)

atif-github-venture:
I wanted to explore a little here.. do you think you can point me to a unit test case which does POS tagging from code base? or any class or a document which can explain the usage of that class?
I want to understand first how the current logic of POS tagging works in this application. @MikeUnwalla

atif-github-venture:
@MikeUnwalla @jaumeortola can anyone of you please point me in a direction here? Knowing a class and the rule xml will be a good place to start working on this issue.

atif-github-venture:
Any tip here?

We added below in following file languagetool-language-modules/src/main/resources/org/languagetool/resource/en/added.txt:

longer	longer	RBS
longest	longest	RBR
long	long	RB

it passed the unit tests for this issue.. but its breaking at 2 places..

1. [ERROR] Tests run: 6, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 25.861 s <<< FAILURE! - in org.languagetool.rules.en.EnglishPatternRuleTest
[ERROR] testRules(org.languagetool.rules.en.EnglishPatternRuleTest)  Time elapsed: 21.13 s  <<< ERROR!
org.languagetool.rules.patterns.PatternRuleTest$PatternRuleTestFailure: Test failure for rule FROM_FORM[5] in file /org/languagetool/rules/en/grammar.xml: Example sentence marked with 'triggers_error' didn't actually trigger an error: 'Over one million Taiwanese participate in the 228 Hand-in-Hand rally form a long human chain to commemorate the February 28 Incident in 1947.'
        at org.languagetool.rules.patterns.PatternRuleTest.addError(PatternRuleTest.java:322)

2.
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 9.351 s <<< FAILURE! - in org.languagetool.rules.en.EnglishTest
[ERROR] testLanguage(org.languagetool.rules.en.EnglishTest)  Time elapsed: 9.307 s  <<< FAILURE!
[...]

PR #6141, opened 2021-12-09
Feature/missing pos for long 5518

atif-github-venture:


atif-github-venture:
Solve for the issue -> https://github.com/languagetool-org/languagetool/issues/5518

PR #6146, opened 2021-12-12
remove 'hard coded' from spell checker

SethFalco:
I feel like I might've done this wrong. 🤔 Please just let me know where I should add it instead if so!

Removes "hard coded" from the spell checker to prevent the 2 step correction.

image

image


tiff:
@SethFalco It belongs in the "prohibit.txt". The "prohibit_custom.txt" is only for people who want to host LanguageTool themselves and customize it for their needs.

SethFalco:
Sorry about that! Should be in the right place now!

PR #6162, opened 2021-12-17
fix for min_prev_matches to respect mode parameter

fabrichter:
introduce PatternRuleTransformer
transform pattern rules with min_prev_matches into TextLevelRules


SteVio89:
Looks good to me so far.

PR #6178, opened 2021-12-22
refactor PatternRuleTransformer interface

fabrichter:
allow grouping multiple rules in one rule,
including or excluding rules for further transforms, etc.

group in RepeatedPatternRuleTransformer so repetitions from one rulegroup match

TODO: 
- [x] add tests for new functionality in other languages / with more rules
- [x] improve/revert changes to REP_THANK_YOU_FOR

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
+++ b/languagetool-core/src/main/java/org/languagetool/JLanguageTool.java
@@ -680,25 +680,13 @@ public class JLanguageTool {
     List<AbstractPatternRule> rules = new ArrayList<>(patternRules);
     List<PatternRuleTransformer> transforms = Arrays.asList(new RepeatedPatternRuleTransformer(lang));
 
-    // PatternRuleTransformer could want to indicate if transformed rules
-    // should be removed from the pool for further transform or not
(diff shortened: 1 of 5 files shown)


jaumeortola:
I will add a test for an existing rule in Catalan. If the rule REP_THANK_YOU_FOR is just a duplicate, we can revert the changes.

jaumeortola:
To avoid undesired results, the subrules in the group rule have to be mutually exclusive. Another solution is to use a filter like `CleanOverlappingFilter`. I am not adding the filtering yet.

fabrichter:
>To avoid undesired results, the subrules in the group rule have to be mutually exclusive. 

Could you elaborate on that? I am not quite sure what you mean.

jaumeortola:
> To avoid undesired results, the subrules in the group rule have to be mutually exclusive.

In the Catalan rule REP_INICIAR (before changes), patterns matched by REP_INICIAR[1] were also matched always by REP_INICIAR[2]. In the regular pattern rules, that never happens because, after a match, the other rules are ignored. To avoid confusions, we should apply the same logic in PatternRuleTransformer. Do you know where the overlapping matches inside a group rule are removed (to copy it)? I don't know it. 

Anyway, it is not critical. We can merge what we have now.

PR #6270, opened 2022-01-21
[de] add grammar.xml for de-CH

SteVio89:


languagetool-language-modules/de/src/main/resources/org/languagetool/rules/de/de-CH/grammar.xml
```diff
@@ -0,0 +1,52 @@
...
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this library; if not, write to the Free Software
+  ~ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+  ~ USA
+  -->
+
+<!DOCTYPE rules [
+        <!ENTITY anfauf "„|&quot;|»|›|‚|'">
+        <!ENTITY anfzu "‘|“|‹|«|&quot;|'">
```
fabrichter, line 27: Shouldn't these entities be used in the pattern?

languagetool-language-modules/de/src/main/java/org/languagetool/language/SwissGerman.java
```diff
@@ -93,4 +93,13 @@ public boolean isVariant() {
     return newRuleMatches;
   }
 
+  @Override
+  public String getOpeningDoubleQuote() {
```
fabrichter, line 97: We need a separate PR that also adds this for SwissGermanPremium

fabrichter (approved):

languagetool-language-modules/de/src/main/resources/org/languagetool/rules/de/de-CH/grammar.xml
```diff
@@ -0,0 +1,52 @@
...
+    <!-- ====================================================================== -->
+    <!-- Tippfehler -->
+    <!-- ====================================================================== -->
+    <category id="TYPOS" name="Mögliche Tippfehler">
+        <rulegroup id="TYPOGRAPHIC_QUOTES_de_CH" name="Anführungszeichen: &quot;xyz&quot; -> «xyz»">
+            <rule>
+                <!-- https://forum.languagetool.org/t/typografische-anfuhrungszeichen-fur-de-ch-mit-giullemets/7526/3 -->
+                <pattern>
+                    <token>"</token>
+                    <token spacebefore="no"/>
```
danielnaber, line 43: This will only match a single word (yes, just like the German rule). Maybe we can find a better solution for phrases? Or will it automatically work because there's some other rule for that?

danielnaber (changes requested):

PR #6280, opened 2022-01-26
added two more JUnit Test Cases

aditidubey12:


danielnaber:
Thanks, but is there any reason these specific tests are useful? Each rule comes with its own tests anyway.

PR #6284, opened 2022-01-26
treat remote rule filters rule IDs as regexps

fabrichter:
Allow remote rule filters to match multiple rule IDs by treating their IDs as regular expressions


fabrichter:
Idea from @palasso

udomai:
This is a cool thing. Can I merge it? Does anyone else need to check it first? @fabrichter @palasso

fabrichter:
@udomai  I'd like @SteVio89 to have a look and approve it, then we can merge.

SteVio89 (approved):

PR #6322, opened 2022-02-14
reduce classloading (LtThreadPoolFactory, Prometheus) when there are no remote rules

donnerpeter:


donnerpeter:
This is needed to avoid bundling Prometheus with IntelliJ after updating LanguageTool to 5.6

fabrichter (approved):

PR #6384, opened 2022-02-28
[de] CH quotes rule to CH grammar

udomai:
@affemitkaraffe, hast du diese Regel schon im nightly diff getestet?


affemitkaraffe:
Ja

PR #6464, opened 2022-03-21
[de] improve SICH_GEGENSEITIG

St-ac-y:
Evtl. als subrule für SICH_GEGENSEITIG?


udomai:
@St-ac-y, could you re-do this PR on an up-to-date branch, please? Also, please put an exception/AP into the first subrule to make sure it doesn't steal matches from the second subrule. And only temp_off the second subrule 🙂

PR #6471, opened 2022-03-24
[de] add "löchen" to suggestions.txt

Discostu36:
I am uncertain if I understand this file correctly as it is not documented (or I haven't found the documentation). I want to improve suggestions for the typo "löchen".


danielnaber:
That file is only used in a test, it has no impact on the suggestions. You could edit `GermanSpellerRule.getOnlySuggestions`, but fixing the underlying issue would be better, of course.

Discostu36:
Yeah, I overlooked that this file is in the "test" folder, sorry. As I am not able to fix the underlying issue, I'll edit the GermanSpellerRule instead.

PR #6506, opened 2022-04-06
[de] add words to spelling.txt

Discostu36:
https://de.wikipedia.org/wiki/Syndikus


danielnaber:
You can use flags like `/S` to avoid some duplication, BTW.

Discostu36:
I know, but somehow forgot in this case.

PR #6526, opened 2022-04-11
[de] add AP for DE_AGREEMENT

BellaDiekmann:


udomai:
To test whether your AP actually works in the contexts it's made for, you can add a correct test sentence at line 108 in `GermanRemoteRuleFilterTest.java`. That way you can test if it works. You could also use the SentenceSourceChecker or HTTPServer to that purpose (will be mentioned at 4 pm).

BellaDiekmann:
There has been a newer commit on May 4th.

PR #6576, opened 2022-04-23
Update GenericUnpairedBracketsRule.java fix issue 6566

Porthoos:
add function isInterval() to detect interval if bracket not match
use function isInterval() in function fillSymbolStack()


danielnaber:
Thanks! Could you extend the tests accordingly?

Porthoos:
> Thanks! Could you extend the tests accordingly?
Yes, I will work on it.
Thank you for your reply.

PR #6629, opened 2022-05-03
[pt] Deletes suggestions of change from words used in Portugal to words used in Brazil

ricardojosehlima:
The file replace.txt for pt-BR brings words that are used in Portugal and should be replaced by words used in Brazil. Most cases involve the diacritics (abstémio = abstêmio). So, LT recognizes my location (Brazil) and if I write abstémio it correctly suggests abstêmio, for the latter is the brazilian form.
However, the file also brought some dozens of cases that are inappropriate. One is the suggestion to change 'avançado' for 'atacante'. In Portugal, a position in a soccer team is called 'avançado' while in Brazil it is 'atacante'. However, 'avançado' also and mostly means 'advanced' in Portuguese, so it doesn't make sense for LT to suggest every time that one writes 'avançado' in Brazil to change to 'atacante', because almost always it will not be the case .
The same for the other words deleted.


ricardojosehlima:
By making a contribution to this project, I certify that:
(a) The contribution was created in whole or in part by me and I have the right to submit it under the open source license indicated in the file; or
(b) The contribution is based upon previous work that, to the best of my knowledge, is covered under an appropriate open source license and I have the right under that license to submit that work with modifications, whether created in whole or in part by me, under the same open source license (unless I am permitted to submit under a different license), as indicated in the file; or
(c) The contribution was provided directly to me by some other person who certified (a), (b) or (c) and I have not modified it.
(d) I understand and agree that this project and the contribution are public and that a record of the contribution (including all personal information I submit with it, including my sign-off) is maintained indefinitely and may be redistributed consistent with this project or the open source license(s) involved.
Signed-off-by: Ricardo Joseh Lima ricardo.lima@uerj.br

marcoagpinto:
Sorry… I was cleaning my e-mails and saw it only now.

PR #6631, opened 2022-05-03
[core] add more (sub)tags

udomai:
@jaumeortola, I've added the new tags in `rules.xsd`, too. Look good?


udomai:
Merging, keeping an eye on it today...

udomai:
Merging, keeping an eye on it today...

Issue #1919, opened 2019-09-06
[en] A sentence with an initial lower-case letter is not always found


MikeUnwalla:
In yesterday's snapshot:
![image](https://user-images.githubusercontent.com/5816160/64413122-d342a280-d088-11e9-9a3c-986c8850467d.png)

The problem also occurs on languagetool.org.

Possible related issue: https://github.com/languagetool-org/languagetool/issues/248

Yan-ice:
Before fixing, "Good morning. \n it is a nice day." would not be recognized as a sentence with an initial lower-case letter, because Segment would not split the sentence. Now it works in PR#6654.
Also, I'v read the code in LanguageTool and found that a sentencethat end with comma would cause the initial lower-case letter (line breaks are ignored in LanguageTool), so maybe "Hello and welcome, \n\nsorry, I have to reply in English." is correct.

danielnaber:
There are two modes in `segment.srx`, `ByLineBreak` and `ByTwoLineBreaks`, so LT is basically prepared to deal with both ways to interpret `\n`. So, I'm not sure whether a change in the code is really needed. Sorry I don't have time to look at this more closely now.

Yan-ice:
I've seen these before, but they aren't work properly according to my JUnit test. Honestly, I really wonder why "Good morning. \n it is a nice day." is not splitted by Segment before I write new rules, though "ByLineBreak" already exists.

PR #6654, opened 2022-05-11
Fix issue #1919

Yan-ice:
Two or more line break should be considered as the end of a sentence.


Yan-ice:
This is the PR to fix issue #1919

QosmosTheQ (approved):

Issue #6568, opened 2022-04-22
[en] verb repetition false positive


donnerpeter:
`If this map maps one or more keys to the specified value...`

1371131436:
This problem can be solved by modifying the rule REPEATED_VERBS in grammar.xml, and I am considering a   
good antipattern for this case.

MikeUnwalla:
Possibly, you can add chunks to the pattern. For the test sentence, the chunker gives a correct analysis. (I used an LT snapshot from 6 April for this screenshot):

![image](https://user-images.githubusercontent.com/5816160/168417731-5185026f-b9a4-4051-b901-7a2ccadc3db9.png)

Another option is to make the disambiguation better so that it gives the correct POS (map/NN and maps/VBZ).

1371131436:
@MikeUnwalla Yes, I have solved this problem by adding chunks, and I am making a pull request to solve this issue.

MikeUnwalla:
Great, thank you. Just beware that the chunker isn't always correct.

I'll leave it to the regular maintainers to deal with your pull request.

PR #6671, opened 2022-05-14
[en] correct a grammar problem about rule of REPEATED_VERBS

1371131436:
This PR aims to contribute to #6568. I add a new antipattern using chunks in grammar.xml to avoid the warning of this kind of case. 
270OQLX`7S)L{()@%X8UVTI


danielnaber:
Thanks. When you add an antipattern, could you also add an `<example>` for it?

1371131436:
@danielnaber Thanks for your advice! An `<example>` for this antipattern has been added to grammar.xml.

AzadehSafakish (approved):
The AP should be fine, thanks for the contribution 👍

PR #6672, opened 2022-05-14
fix: add redisDatabase to known properties

justjanne:
The redisDatabase property already exists and works fine, it’s just missing from the known properties.

When deploying languagetool in a cluster with a shared redis database, this can be useful to avoid key conflicts


danielnaber:
Thanks!

Issue #6578, opened 2022-04-23
Problem about CD with NN:UN detection


Leosang-lx:
For this simple example:
`"I have two apple."`
![image](https://user-images.githubusercontent.com/62881345/164892385-dbed78f6-b07c-45ac-8bdc-703e4b2b0e0b.png)
The tag of "apple" is [apple/NN:UN,E-NP-singular], and the rule CD_NN cannot show the problem since the POS-tag of apple is NN:UN and the rule of pattern only match the last word with tag "NNP" or "NN". Besides, the exception of this token also make NN:UN an exception.
Below is the related token of pattern.
`<token chunk="E-NP-singular" postag="NNP?" postag_regexp="yes">
         <exception postag="NN:UN?|NNP?S|JJ.*|IN|CD" postag_regexp="yes"/>
......`
The problem is related to the issue [#3864](https://github.com/languagetool-org/languagetool/issues/3864). I'm trying to handle this problem. Looking forward to some feedback, thx!

Leosang-lx:
@danielnaber I have questions about how is the word `apple` in `I have two apple.` tagged into `NN:UN`. Since the tag is a new tag shown in `tagset.txt`, and I have no clear idea of the meaning of `NN:UN`. Until now, I can simply address the problem by small modifications on `grammar.txt`, and looking forward to address the problem from modifying the tokenization of word.

danielnaber:
You can see the tags here: https://github.com/languagetool-org/languagetool/blob/master/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/tagset.txt - "apple" as `NN:UN` looks like a bug to me, but an English native speaker should decide that.

MikeUnwalla:
'apple' as NN:UN is correct: https://www.ldoceonline.com/dictionary/apple

Leosang-lx:
> 'apple' as NN:UN is correct: https://www.ldoceonline.com/dictionary/apple



> You can see the tags here: https://github.com/languagetool-org/languagetool/blob/master/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/tagset.txt - "apple" as `NN:UN` looks like a bug to me, but an English native speaker should decide that.

Thanks, I will check whether my existing modification on rules will affect the test cases.

Issue #3864, opened 2020-11-16
[en] False warning EN_COMPOUNDS


MikeUnwalla:
EN_COMPOUNDS suggests **twenty-second**. (The real correction is **twenty seconds**.)

> Sami just sat there for about **twenty second**.

PR #6677, opened 2022-05-15
[en] Correct problem about detection on "... two apple."(pattern: CD and E-NP-singular)

Leosang-lx:
This PR address the problem on detection of pattern: CD(the postag of token) with E-NP-singular(the chunk of token). Issue https://github.com/languagetool-org/languagetool/issues/6578 showed the problem, which was created by myself. And issue https://github.com/languagetool-org/languagetool/issues/3864 also show the wrong correction on this pattern.
After updating grammar.xml, cases on this pattern(mention above) can be found and give corrections.
image

And the word "second" will be tagged with postag "NN:UN" and "JJ", so it is an exception to handle in this pattern. Some other word that with postag "NN:UN" and "JJ" should also be considered to be an exception in this pattern, so "JJ.*" was removed from the pattern exception. Otherwise, correction of "twenty seconds" in the following case won't be given.
image


danielnaber:
Could you add `<example>` sentences that check your change?

Leosang-lx:
> Could you add `<example>` sentences that check your change?

Do you mean add examples on the cases that I mentioned? Like words tagged with "NN:UN"?

danielnaber:
You should add examples that wouldn't have worked correctly before, but now do with your change.

Leosang-lx:
> You should add examples that wouldn't have worked correctly before, but now do with your changUn

Understood, thx!

Leosang-lx:
@danielnaber Examples have been added to `grammar.xml`.

PR #6689, opened 2022-05-17
[es] add MALO_AMIGO_MAL_AMIGO

BellaDiekmann:
@jaumeortola: I finally had time to finish the rule to find errors like "Eres un bueno amigo" where "bueno" should be reduced to "buen". As you can see, there are many antipatterns. @udomai mentioned that there is the possibility to use filters. Would this make sense in this case?

### DIFF
--- a/languagetool-language-modules/es/src/main/resources/org/languagetool/rules/es/grammar.xml
+++ b/languagetool-language-modules/es/src/main/resources/org/languagetool/rules/es/grammar.xml
@@ -12230,6 +12230,168 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
         </rule>
     </category>
     <category id="GRAMMAR" name="Gramática" type="grammar">
+        <rule id="MALO_AMIGO_MAL_AMIGO" name="«malo» y «bueno» se apocopan delante de sustantivos masculinos en singular" default="temp_off">
+            <antipattern> <!--1-->
+                <token postag="V.*|DE0CN0" postag_regexp="yes" skip="1" min="0"/>
+                <token regexp="yes" skip="1">malo|bueno</token>
+                <token postag="VMN.*" postag_regexp="yes"/>
+            </antipattern>
+            <antipattern> <!--2-->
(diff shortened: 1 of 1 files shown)


jaumeortola:
I don't see a way to use filters here. 

For native speakers, probably the most frequent error is a missing comma. Perhaps we could always suggest also a comma (at least for bueno, not malo) as second suggestion. ¡Qué bueno chico! -> ¡Qué buen chico! // ¡Qué bueno, chico! 

Other than that, the antipatterns or exceptions could be simplified and generalized a bit. Some seem redundant. 3/4 antipatterns should be enough: `capitalized Bueno/Malo` except at the sentence start (you can use the tag _possible_NP),  `NC[MC][SN]000 + bueno/malo`, `ser/qué + bueno/malo + inf.`...  It has to be tried.

BellaDiekmann:
> I don't see a way to use filters here.
> 
> For native speakers, probably the most frequent error is a missing comma. Perhaps we could always suggest also a comma (at least for bueno, not malo) as second suggestion. ¡Qué bueno chico! -> ¡Qué buen chico! // ¡Qué bueno, chico!
> 
> Other than that, the antipatterns or exceptions could be simplified and generalized a bit. Some seem redundant. 3/4 antipatterns should be enough: `capitalized Bueno/Malo` except at the sentence start (you can use the tag _possible_NP), `NC[MC][SN]000 + bueno/malo`, `ser/qué + bueno/malo + inf.`... It has to be tried.

Thank you for your feedback @jaumeortola. I'll have a look at it. Just to get it right: you mean that 3-4 antipatterns in total should be enough?

jaumeortola:
I have improved the disambiguation of "dejo". It is better to improve the disambiguation, when possible, than to write specific exceptions. The improvement in disambiguation will benefit other rules.

BellaDiekmann:
> I have improved the disambiguation of "dejo". It is better to improve the disambiguation, when possible, than to write specific exceptions. The improvement in disambiguation will benefit other rules.

I'll have a look at how to reduce all the antipatterns as soon as possible. The idea of suggesting a comma as second possibility sounds good. I'll consider that as well.

PR #6768, opened 2022-06-09
[de] add pos tags

affemitkaraffe:


udomai:
currently being tested

PR #6786, opened 2022-06-14
[de] HunspellRule: exceptions for split suggestions

jaumeortola:


danielnaber:
Thanks, could you also add a test case for this?

PR #6809, opened 2022-06-18
[LO extension] Update Polish translation

xesarni:
New strings and some fixes.


danielnaber:
Thanks, but we don't edit the `MessagesBundle_pl.properties` in git, we use webtranslateit.com instead. You can email us (see https://languagetool.org/legal/) and we can add you there.

PR #6838, opened 2022-06-27
Add username and apiKey parameters to RemoteLanguageTool

shihabuddin:
This would make it possible to access premium API using RemoteLanguageTool


shihabuddin:
Thanks @danielnaber. Can you please tell me in which version will this be released and when?

danielnaber:
It will be part of 5.9 (see https://dev.languagetool.org/roadmap)

PR #6845, opened 2022-06-28
[de] add new words to spelling.txt

Discostu36:


danielnaber:
Thanks, but isn't "Kita" the more common and correct spelling?

Discostu36:
> Thanks, but isn't "Kita" the more common and correct spelling?

Well, at least it is the spelling that Duden suggests, but "KiTa" is also correct. Do we have a rule for suggested spellings, where Kita/KiTa could be added so that it is not a spelling mistake any more but a style suggestion? Or is that only available in Pro?

PR #6855, opened 2022-06-30
[es] add improved MALO_AMIGO_MAL_AMIGO

BellaDiekmann:
@jaumeortola: I finally had time to improve the rule for finding errors such as „malo amigo“ (see also https://github.com/languagetool-org/languagetool/pull/6689/files):
- I included the use of a coma as a second suggestion
- I exlcuded upper case „Bueno“ and „Malo“ 
- I removed all superfluous antipatterns
What do you think? Is the rule ready to go or should I still improve it?
Thank you for your feedback!


jaumeortola:
Great. The rule seems ready.

PR #6875, opened 2022-07-06
Sb03

susanaboatto:
I have added a couple of antipatterns to this rule because it was throwing a few false positives. If @St-ac-y or anyone else could review it when/if you have time, I'd really appreciate it :)

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -9439,46 +9439,7 @@ Isto observa-se quando as forças avaliam mal o cenário.
       <example>…rimeira é eleita por sufrágio universal directo e tem poderes fundamentalmente legislativos, além de fiscalizar…</example>
 	  <example type='correct'>Que mal tem <marker>fazer chichi</marker> na piscina?</example>	
     </rule>
-    <rule id='PHRASAL_VERB_RESIDIR_EM' name='Regência verbal: Residir em'>
-    <!--      Created by Tiago F. Santos, 2018-09-19      -->
-    <!--antipattern>
-          <token regexp='yes'>residentes?|morador(?:es)?</token>
-          <token/>
-          <token regexp='yes'>com|em|n[ao]s?</token>
-      </antipattern-->
-      <antipattern>
-          <token regexp='yes'>residentes?|morad(?:as?|or(?:es)?)</token>
-          <token postag_regexp='yes' postag='A.+'/>
-      </antipattern>
-<!-- MARCOAGPINTO 2020-10-06 + 2020-10-27 (21-OCT-2020+) *START* --> 
-<!-- "No núcleo reside a força imprescindível dos modelos." -->
-      <antipattern>		  
-		  <token regexp="yes">residem?</token>
-		  <token regexp="yes">as?|os?|um|uns|umas?</token>
-		  <token postag='NC[CFM][SP]000|AQ0[CFM][SP]0' postag_regexp='yes'/>
-	  </antipattern>
-<!-- MARCOAGPINTO 2020-10-06 + 2020-10-27 (21-OCT-2020+) *END* -->
-      <pattern>
-          <token inflected='yes' regexp='yes'>residir|morar|residente|morador</token>
-          <token postag_regexp='yes' postag='[DN].+'>
-            <exception postag_regexp='yes' postag='[AP].+'/>
-            <exception postag_regexp='yes' postag='R.+'/>
-            <exception regexp='yes'>com|em|n[ao]s?</exception>
-            <exception scope='next' regexp='yes'>perigos?</exception></token>
-      </pattern>
-      <message>Erro de regência. Dependendo do significado, escolha:</message>
-        <suggestion>\1 em \2</suggestion>
-        <suggestion>\1 com \2</suggestion>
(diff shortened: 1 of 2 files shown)


St-ac-y:
each 'topic' should be on a different branch (also new PR).

susanaboatto:
@marcoagpinto Thanks for the tips! If there's somewhere you could share your rule creation wisdom, it will surely be much appreciated! My background is in linguistics, so it might take a while to adjust to the coding part of it all. 

I want to add antipatterns to this rule to fix the false positives in the case of numbers with percentages and where "fora" is not a verb. What do you think of the antipatterns below?       

      <antipattern case_sensitive='yes'>
          <token>Fora</token>
	  <token postag='AQ..P.+|NC.P.+' postag_regexp='yes'/>
	  <token postag='AQ..P.+|NC.P.+' postag_regexp='yes'/>
          <example>Fora casos de ausência por questões de saúde, não é permitido faltar.</example>
      </antipattern>
      <antipattern>
          <token postag=',|-' postag_regexp='yes'/>
          <token>fora</token>
	  <token postag='AQ..P.+|NC.P.+' postag_regexp='yes'/>
          <example>Não é permitido faltar, fora casos de ausência por questões de saúde.</example>
      </antipattern>
      <antipattern>
	    <token postag='V.+' postag_regexp='yes'/>
	    <token postag='NCMP000' postag_regexp='no'/>
          <example>Estou 100% com você.</example>
      </antipattern>

marcoagpinto:
@susanaboatto 

Always see the POS information at:
https://community.languagetool.org/analysis/analyzeText

Paste there:
```
Fora casos de ausência por questões de saúde, não é permitido faltar.
Não é permitido faltar, fora casos de ausência por questões de saúde.
```

You can see that "fora" can be a noun and also verb "ir" and "ser":

fora_postags_20220706

Using words should be avoided the most possible since they limit the rules.
This should work with your two first antipatterns:

```
  <antipattern>
    <token postag='NCCS000' postag_regexp='no'/>
    <token postag='AQ..P.+|NC.P.+' postag_regexp='yes'/>
    <token postag='SPS00' postag_regexp='no'/>
<example>Fora casos de ausência por questões de saúde, não é permitido faltar.</example>
<example>Não é permitido faltar, fora casos de ausência por questões de saúde.</example>
  </antipattern>
```

Then make some tests replacing:
`<token postag='NCCS000' postag_regexp='no'/>`
with:
`<token postag='NC.S.+' postag_regexp='yes'/>`

Use the corpus provided by Daniel Naber to check the results (unzip the Wikipedia tool into a folder):
`java -Dfile.encoding=UTF-8 -Xmx4500M -jar languagetool-wikipedia.jar check-data -l pt-PT -r GENERAL_NUMBER_AGREEMENT_ERRORS -f pt-PT.txt -f tatoeba-pt.txt --max-sentences 600000 --context-size 100 >0.txt`

You can replace here the rule ID and the number of sentences. I have at 600 000 because I am slightly mad.

The first check is before the creation of antipattern, use 0.txt as output.

As you make changes and test them, change the output to 1.txt 2.txt 3.txt , etc.

Then make a DIFF (on Windows you can use TortoiseSVN) to see the false positives and success.

If you have several .txts, rename the most correct one to something like "2stable.txt", which means that future outputs can be checked (diff) against 2stable.txt instead of 0.txt.

Only check the last output with 0.txt before comming the change, to make sure nothing escapes.

Also, unzip the standalone tool to the desktop to to a
`testrules pt`
or
`testrules pt-pt`
or
`testrules pt-br`
depending on what you are doing.


Ahhhh... on the Wikipedia and Standalone unzips, you need to write the commands from the shell (DOS).

You can simply open the folders in normal windows, copy the path, open the shell by writing CMD in the magnifying lent on Window 10 and type there
`cd path_here`
You should have two shells/CLI, one for Wikipedia and other to standalonetool

I advise you to use Notepad++ to edit the files, it is extremelly powerful, just remember to go to its menu that allows to select the language of the document and change to XML so that all tags appear with colours.

I have three grammar.xml in my notepad++
1. from the repository (checkout)
2.. standalone tool
3. wikipedia tool

Always make the changes first on the stand alone and do a TESTRULES PT, then copy the whole text into the wikipedia file (CTR+A, CTR+C) go to the wikipedia tab and press (CTR+A, CTR+V).

Then test the results.

If all is okay you can copy the file from the standalone tool and replace the repository one with it and commit.

ahhhhh... tell me if this helps. I really need to write some good documentation for all this with examples.

❤️❤️❤️❤️❤️❤️❤️

marcoagpinto:
Ahhhhhhhh…

I remembered one thing I always do.

Before starting to edit or create new rules, I first update the checkout files.

Only after it, I open Notepad++ to make sure I have the latest grammar.xml.

Then I copy and paste that grammar.xml into the standalone tool grammar.xml.

🙂

St-ac-y:
thank you @marcoagpinto for all you help 🥇

susanaboatto:
Hi @St-ac-y 

It seems like @marcoagpinto has corrected it already, so I think we're good.

> thank you @marcoagpinto for all you help 🥇 @susanaboatto once you have fixed and corrected everything it might be best to open a new pull request for INKING_VERB_PREDICATE_AGREEMENT. maybe you can get it done before noon so I can review and hopefully merge it 👍

@marcoagpinto thanks for all the help and detailed information! I might need some time to learn how to run the test the way you described, it's a lot to take in. I still need to go through [the dev page](https://dev.languagetool.org/) to learn all the tricks :sweat_smile:

PR #6908, opened 2022-07-13
[pt] created antipattern in rule DOUBLE PUNCTUATION to avoid FAs

susanaboatto:
Made some antipatterns to avoid annoying/unwanted corrections in the double punctuation rule. Let's see if this works @danielnaber

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -19160,6 +19160,26 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
         <example type='correct'>O quê<marker>?!</marker></example>
       </rule>
       <rule>
+        <antipattern>
+          <token regexp='yes' skip='-1'>\d+|&operadores_matematicos;</token>
+          <token min='2'>-</token>
+          <example>38 -- 29</example>
+          <example>--[5]--</example>
+        </antipattern>
+        <antipattern>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -19160,6 +19160,26 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
...
+        <example>38 -- 29</example>
+        <example>--[5]--</example>
+        </antipattern>
+        <antipattern>
+          <token>-</token>
+          <token spacebefore='no' min='2'>-</token>
+        </antipattern>
+        <antipattern>
+          <token min='2'>-</token>
+          <token min='2'>-</token>
```
danielnaber, line 19181: Isn't this equivalent to `<token min='4'>-</token>`? Then you could use just that.

susanaboatto: I figured that maybe it wouldn't detect patterns with space, like "-- -- -- --", which showed up when I was testing the rule. But if it doesn't make a difference, then I think I'll just exclude this and leave only one antipattern with ` <token min='2'>-</token>`

danielnaber: No, you're right, I forgot about that case. But could you add an `<example>`?

susanaboatto: Done

PR #6932, opened 2022-07-18
Improve annotated text

jaumeortola:
Hidden characters \u202d \u202C \u00AD (soft hyphen) are treated as markup.


tiff:
I have informed Google but they might need a while until they have fixed it. There are plenty of support requests caused by this issue with several symptoms:
* wrong language detected
* everything underlined in red
* weird false positives

jaumeortola:
@tiff Is there a way to reproduce the errors in Google Docs?

PR #6945, opened 2022-07-21
improve GermanSpellerRule.removeGenderCompoundMatches

SteVio89:


danielnaber:
Could you check (and add tests) whether this has other effects? e.g. was `richtig_stimmt` (i.e. underline but no suffix) accepted before? What about `flasch_nittrichtig`? Does that change now? Should we care if it does?

SteVio89:
`richtig_stimmt` and `flasch_nittrichtig` are treated as before. Because the regex checks for uppercase letter as first char.

PR #6957, opened 2022-07-26
Add initial priorities for HydraLEO for PT and NL

palasso:
## Description
This PR adds initial priorities for HydraLEO for Portuguese (PT) and Dutch (NL).

## Issues
There are some concerns about the exact values of the priorities.

### Spellchecker priority
I am not sure which is the spellchecker for PT and NL and what its priority is. So far, HydraLEO would have a lower priority than the spellchecker.

For PT I see:
```
case "PT_DIACRITICS_REPLACE":     return -45;   // prefer over spell checker
case "HUNSPELL_RULE":             return -50;`
```
thus `HUNSPELL_RULE` seems to be the spellchecker and has a priority of `-50`.

For NL I see:
```
case "KORT_2": return -5;  //so that spelling errors are recognized first
case "EINDE_ZIN_ONVERWACHT": return -5;  //so that spelling errors are recognized first
```
thus the spellchecker seems to have a higher priority than `-5` (probably `0` as it doesn't seem to be on the list).

Based on the above findings, I adjusted the priority of HydraLEO in general to be lower than the spellchecker (`-5` for NL and `-51` for PT).

The low priority value in the Portuguese spellchecker raises the question whether there is a reason for it to have such a low priority value. I don't know if specifically for Portuguese most rules, and thus HydraLEO as well, are expected to have a higher priority. That being said I notice that the spellcheckers for French and Spanish also have low priority values (specifically `FR_SPELLING_RULE` and `MORFOLOGIK_RULE_ES` have `-100`) thus it may not be a concern.

### Comma XML rules priorities (including style rules)
For the specific case of HydraLEO missing commas, I have added a priority of `-51` so that comma XML rules, including style rules, have a higher priority. I can't determine though what is the range of priorities comma rules have (and which their Rule IDs are) thus I can't know for sure what is the priority for comma style rules as well.

One way to find out comma XML rules and what their priorities are would be to introduce high priorities for HydraLEO missing commas initially and find via the nightly diff which ones are replaced with HydraLEO matches. Could that be a way to resolve the issue?


palasso:
@danielnaber @fabrichter please review this issue

palasso:
The following is unrelated to HydraLEO and to the PR.
For `Portuguese.java` in `getPriorityForId` I notice in the end instead of a `return super.getPriorityForId(id);`  it has a `return 0;` . I am not sure if that's intended for `id` for cases not defined within the method. For the other languages that I have seen (English, German, French, Spanish) it would use `return super.getPriorityForId(id);` for these cases.

danielnaber:
The Dutch speller rule has id `MORFOLOGIK_RULE_NL_NL` and indeed it seems to have priority 0. Starting with a low priority like `-51` for a new comma rule seems safe.

Thanks for noticing the `return 0`, that probably should be changed. I'll do that locally and run some tests first.

fabrichter:
Cherry-picked this: https://github.com/languagetool-org/languagetool/commit/2f03f684121865044b34224c1d17b63c2b4fc079 and pushed it

PR #6971, opened 2022-07-29
I am trying to add Kurdish

amedcj:
I would like to add Kurdish to LanguageTool. I would be grateful for any help.

### DIFF
--- a/languagetool-core/src/main/resources/org/languagetool/MessagesBundle.properties
+++ b/languagetool-core/src/main/resources/org/languagetool/MessagesBundle.properties
@@ -208,6 +208,7 @@ incorrect_case = This sentence does not start with an uppercase letter.
 is = Icelandic
 it = Italian
 km = Khmer
+ku = Kurdish
 lt = Lithuanian
 missing_space_after_comma = Put a space after the comma.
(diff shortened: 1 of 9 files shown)


danielnaber:
Thanks, however, we currently don't add new languages unless they have been actively maintained for some time. You can find more details at https://dev.languagetool.org/adding-a-new-language. So I suggest you create a fork and work in that fork, adding rules and getting it ready.

amedcj:
Hello Daniel.
Thank you for your answer.
I have already created a fork and i will add rules, as you suggested. In the meantime, i will translate UI.
But you have written "unless they have been actively maintained for some time". How long was "some time" for languages added earlier?
By the way, I heard about LanguageTool a few days ago and was amazed.
I have voluntarily translated Open Source programs into Kurdish for years. Now I'm actively translating "Firefox Android". I am honored to work for the outstanding LanguageTool.

danielnaber:
> How long was "some time" for languages added earlier?

We didn't have this rule in the past, resulting in many unmaintained languages. A fork should be maintained for at least 6 months, and it also helps a lot if there's more than one person working on it.

PR #6995, opened 2022-08-05
Update MessagesBundle_de.properties

Mr-Update:
Translation completed and improved


danielnaber:
Thanks, but we use https://webtranslateit.com for translations. Changes made directly to the files would be lost with the next update from webtranslateit.com. Let me know if you'd like to have an access there (I need your email for that).

PR #7069, opened 2022-09-05
[en] when_what sample first rule

evan-defran-lt:
Targets wrong WH-word confusion _when_ vs _what_
may incorrectly suggest _what_ when relative pronoun _which_ is called for.


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -133322,7 +133322,20 @@ The accident victim died from her injuries.
                 <example correction="agent's|agent’s">Go to the travel <marker>agent`s</marker> office.</example>
             </rule>
         </rulegroup>
-
+        <rule id="CONFUSION_OF_WHEN_WHAT" name="confusion of when/what">
```
danielnaber, line 133325: This should probably have `default="temp_off"` so it's in the nightly first and not immediately active.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -23318,6 +23318,20 @@ USA
...
+            <pattern>
+                <marker>
+                    <token>when</token>
+                </marker>
+                <token postag="MD"/>
+                <token>be</token>
+                <token postag="VBN"/>
+                <token postag="IN"/>
+            </pattern>
+            <!-- concerned this might also falsely flag relative pronouns -->
```
AzadehSafakish, line 23331: If you're concerned about relative pronouns, a solution might be to restrict the rule so that no noun phrases appear before "when". This would also take care of the one FP I see in the editor.
You can try writing an antipattern like how I showed you, using chunks. The chunks for noun phrases are:

`B-NP-singular` (beginning of singular NP)
`I-NP-singular` (middle of singular NP)
`E-NP-singular` (end of singular NP)
`B-NP-plural` (beginning of plural NP)
`I-NP-plural` (middle of plural NP)
`E-NP-plural` (end of plural NP)

AzadehSafakish:
Manually merged.

PR #7081, opened 2022-09-09
[en] refined rule

evan-defran-lt:
Added antipatterns to avoid various FPs. 
Still need to an add an antipattern to avoid 'Support the Ring Cam please'


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -2125,11 +2125,34 @@ USA
     <!-- ====================================================================== -->
     <category id="TYPOS" name="Possible Typo" type="misspelling">
         <rule id="CAN_MISSPELLING" name="misspelling can" default="temp_off">
+        <!-- TODO: work in progress, need to add AP to avoid 'Support the Ring Cam please' -->
             <antipattern>
```
AzadehSafakish, line 2129: To target capitalized words that are not sentence initial, try using:
`<token><exception postag="SENT_START /></token>` in the AP.
Matches any token that's NOT at the start of a sentence.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -2125,11 +2125,34 @@ USA
     <!-- ====================================================================== -->
     <category id="TYPOS" name="Possible Typo" type="misspelling">
         <rule id="CAN_MISSPELLING" name="misspelling can" default="temp_off">
+        <!-- TODO: work in progress, need to add AP to avoid 'Support the Ring Cam please' -->
```
AzadehSafakish, line 2128: name=" *incorrect sentence* (*correct suggestion*)"

For example:
name="It cam (can) happen anytime"

PR #7114, opened 2022-09-22
[en] added rule

evan-defran-lt:
https://github.com/languagetooler-gmbh/languagetool-premium/issues/4622


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -23467,6 +23467,15 @@ USA
     <!-- Grammar -->
     <!-- ====================================================================== -->
     <category id="GRAMMAR" name="Grammar" type="grammar">
+        <rule id="GOOD-EDUCATED" name="She is good-educated (well-educated)." default="temp_off">
```
AzadehSafakish, line 23470: Wait, one (1) naming convention I forgot to mention:
We tend to use underscores here, not hyphens.
Once that's changed, you may merge this yourself! 👍

evan-defran-lt: Even in instances when the hyphen is orthographically called for?

AzadehSafakish: Yes

PR #7335, opened 2022-10-25
[fr] WRONG_ABBREVIATION adding a specific rule to avoid FP for et al.

LucieSteib:
Adding a specific rule to avoid FP on arabic onomastics (al-Andalus) and italian borrowings (al dente) but allow a suggestion for the expression et al. (required in French). 
Examples from diff:
_Il est préférable qu'ils soient à « al dente »._ (FP: -> al. dente)
 _La frontière religieuse et les relations commerciales entre al-Andalus et les Carolingiens (IXe-Xe s.)_(FP: -> al.-Andalus)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -134443,6 +134443,18 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
                 <suggestion><match no="1" regexp_match="(?iu)s$" regexp_replace=""/></suggestion>
                 <example correction="PIB">Les <marker>PIBs</marker> des pays émergents sont croissants.</example>
             </rule>
+            <rule tags="picky"><!--rule12-->
+                <pattern>
+                    <token>et</token>
+                    <marker>
+                        <token regexp="yes" case_sensitive="yes" spacebefore="yes">al
```
jaumeortola, line 134450: `spacebefore="yes"` is the default option. There is no need to write it.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -134443,6 +134443,18 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+                            <exception scope="next">.</exception></token>
+                    </marker>
+                </pattern>
+                <message>Un point est requis.</message>
+                <suggestion>\2.</suggestion>
+                <example correction="al.">Docteur Hun, Docteur Deuh et <marker>al</marker> l'ont démontré.</example>
+                </rule>
```
jaumeortola, line 134457: It is always good to add some correct examples (al-Andalus, al dente...) to make sure that the problem is solved.

LucieSteib: Done, excellent idea!

PR #7375, opened 2022-10-31
[en] added rule

evan-defran-lt:
Rule group "EMIGRATE_IMMIGRATE_CONFUSION"

E.g., We immigrated (emigrated) from Canada in 1979. 
E.g., We emigrated (immigrated) to Canada in 1979.


danielnaber:
@evan-defran-lt Great rule! One of the cases where I always wonder what's the difference between these words...

evan-defran-lt:
@danielnaber Thanks! It's something that I've been thinking of often lately. And keep an eye out for Gina's upcoming blog article on it!

PR #7410, opened 2022-11-09
[fr] AVENT_AVANT rule creation

LucieSteib:
We had a rule to handle when "avent" (noun, time before Christmas) should be replaced by "avant" (preposition), but the opposite didn't exist.


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -18783,6 +18783,18 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
             <example correction="avant">C'était mieux <marker>avent</marker>.</example>
         </rule>
+        <rule id="AVENT_AVANT" name="confusion of avent/avant" default="temp_off"><!--mirror rule from AVANT_AVENT-->
+             <pattern>
+                 <token regexp="yes">bougies?|chants?|couronnes?|dimanches?|messes?</token>
+                 <token regexp="yes">de</token>
+                 <token regexp="yes">l'</token>
+                 <token>Avant</token>
+             </pattern>
+            <message>Le mot <suggestion>Avent</suggestion> est plus probable que "\4" dans ce contexte.</message>
```
danielnaber, line 18793: Maybe we can have a very short description (e.g. in parentheses) what both words mean?

LucieSteib: Done in https://github.com/languagetool-org/languagetool/pull/7412

PR #7419, opened 2022-11-12
save memory: create just one synthesizer per language

donnerpeter:
For German, this saves about 3MB

### DIFF
--- a/languagetool-server/src/test/java/org/languagetool/server/HTTPServerTest.java
+++ b/languagetool-server/src/test/java/org/languagetool/server/HTTPServerTest.java
@@ -21,6 +21,7 @@ package org.languagetool.server;
 import org.apache.commons.lang3.StringUtils;
 import org.junit.Before;
 import org.junit.Test;
+import org.languagetool.JLanguageTool;
 import org.languagetool.Language;
 import org.languagetool.Languages;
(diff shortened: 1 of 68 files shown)


languagetool-server/src/test/java/org/languagetool/server/HTTPServerTest.java
```diff
@@ -313,7 +314,11 @@ private void runDataTests() throws IOException {
   }
 
   @Test
-  public void testTimeout() {
+  public void testTimeout() throws IOException {
+    GermanyGerman lang = new GermanyGerman();
+    // avoid rule <clinit> being interrupted by timeout, resulting in NCDFEs in the following tests
+    lang.getRelevantRules(JLanguageTool.getMessageBundle(), null, null, Collections.emptyList());
```
donnerpeter, line 320: testHTTPServer started to fail because the initialization became lazier and testTimeout's cancellation of futures happened during GermanWordRepeatRule clinit. Such a hack seems OK for the test, but something might be needed to be fixed in production as well. E.g. using more graceful alternatives of `future.cancel`/`shutdownNow` in `TextChecker`.

SteVio89 (approved):

Issue #7045, opened 2022-08-24
Unsatisfied link error on M1 (ARM) mac


samhutchins:
We're getting the following stack trace on macOS running on Apple Silicon:

```
java.lang.RuntimeException: Failed to initialize BridJ (java.lang.UnsatisfiedLinkError: Can't load library: /var/folders/md/k981cwwj58n0sv3v2y529xtc0000gn/T/BridJExtractedLibraries4773592580203442708/libbridj.dylib)
	at org.bridj.Platform.initLibrary(Platform.java:444)
	at org.bridj.Platform.<clinit>(Platform.java:228)
	at org.bridj.Pointer.<clinit>(Pointer.java:208)
	at org.languagetool.rules.spelling.hunspell.DumontsHunspellDictionary.<init>(DumontsHunspellDictionary.java:37)
	at org.languagetool.rules.spelling.hunspell.Hunspell.getDictionary(Hunspell.java:50)
	at org.languagetool.rules.spelling.hunspell.HunspellRule.init(HunspellRule.java:502)
	at org.languagetool.rules.de.GermanSpellerRule.init(GermanSpellerRule.java:1447)
	at org.languagetool.rules.spelling.hunspell.HunspellRule.ensureInitialized(HunspellRule.java:476)
	at org.languagetool.rules.spelling.hunspell.HunspellRule.match(HunspellRule.java:156)
	at org.languagetool.JLanguageTool.checkAnalyzedSentence(JLanguageTool.java:1414)
	at org.languagetool.JLanguageTool.access$1500(JLanguageTool.java:75)
	at org.languagetool.JLanguageTool$TextCheckCallable.getOtherRuleMatches(JLanguageTool.java:1994)
	at org.languagetool.JLanguageTool$TextCheckCallable.call(JLanguageTool.java:1906)
	at org.languagetool.JLanguageTool$TextCheckCallable.call(JLanguageTool.java:1877)
	at org.languagetool.JLanguageTool.performCheck(JLanguageTool.java:1338)
	at org.languagetool.JLanguageTool.checkInternal(JLanguageTool.java:1016)
	at org.languagetool.JLanguageTool.check(JLanguageTool.java:935)
	at org.languagetool.JLanguageTool.check(JLanguageTool.java:920)
	at org.languagetool.JLanguageTool.check(JLanguageTool.java:910)
	at org.languagetool.JLanguageTool.check(JLanguageTool.java:892)
[...]

Issue #7054, opened 2022-08-30
[LibreOffice add-on] Crashed with latest nightly - 2022-08-30


marcoagpinto:
Hello @FredKruse 

I was revising my thesis:
![LT_crash_20220830](https://user-images.githubusercontent.com/5192600/187538527-c767b799-1195-4587-a65a-c563c28c91a5.png)

marcoagpinto:
Basically, I opened the .docx, changed all text language to Portuguese, then selected “Save As” and when I was selecting .ODT format, the window above popped-up.

FredKruse:
@marcoagpinto Your thesis seams to be a hardness test for the extension. :smile: 
Could you please reproduce the issue, close the document and send me the log file right after?
But, I can look at it, not before the beginning of next week.
Note: The check should run after closing the popup, but not all text level rules will be handled correct.

marcoagpinto:
@FredKruse 

If it crashes again, I will send you a log file.

How do I get this log file?

Thanks!

marcoagpinto:
Better yet, after the crash I only used LibreOffice a few times and on the thesis, so I can send you the log I have?

Please tell me where the log is for me to get it.

Thanks!

FredKruse:
The log file is written new every time you start LibreOffice again. So, I need the log-file from the session in which the crash appears.
The name of the log-file is: LanguageTool.log
The directory-path is: \\user\<YourUserName>\AppData\Roaming\languagetool.org\LanguageTool\LibreOffice

marcoagpinto:
@FredKruse 

It crashed again:
![lt_crash_20220831](https://user-images.githubusercontent.com/5192600/187745233-b0d07197-16aa-4a6b-91ba-109e902d99ae.png)

Same procedure causing it:
1. Open .docx;
2. Change all text language to Portuguese;
3. Save as;
4. Select extension .ODT
5. ... and the crash happens.

Regarding the log file it is 3 MB long, and it reveals tons of information (contents) of the thesis, and therefore I can't share it here in a public place.

marcoagpinto:
28878 lines of log.

marcoagpinto:
And it shows full paragraphs of the thesis, I can't share it here.

marcoagpinto:
Here is also a hang (LibreOffice stopped responding and I had to close it)
[...]

PR #7469, opened 2022-11-23
Update dumonts/hunspell to 2.1.1

oleschoenburg:
Updates the dependency on https://gitlab.com/dumonts/hunspell-java to the newly released 2.0.0
Removes the redundant re-implementation of `spell`, `suggest` and `add` methods.

relates to #7045

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/DumontsHunspellDictionary.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/DumontsHunspellDictionary.java
@@ -18,29 +18,19 @@
  */
 package org.languagetool.rules.spelling.hunspell;
 
-import dumonts.hunspell.bindings.HunspellLibrary;
-import org.bridj.Pointer;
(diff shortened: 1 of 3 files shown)


oleschoenburg:
I'm mainly opening this PR to test the waters. The release is still new and the darwin-arm64 binaries in particular are not tested. If someone has access to a M1 machine: Could you try out running languagetool with hunspell-java 2.0.0 to confirm that it works as expected and solves #7054? :pray:

oleschoenburg:
I think it's already usable with Rosetta x86 emulation but the M1 binary is in the wrong path. I'll fix this today and release a new version, it'd be great if your colleague could then test it again to confirm that the native binary also works as expected.

oleschoenburg:
@danielnaber updated to the newest version. As mentioned above, would be great if you could test again to ensure that it works without x86 emulation too

samhutchins:
@oleschoenburg I've tried this updated patch on an M1 machine without rosetta, and it seems to work fine

danielnaber:
Another test on an M1 and on production look good so far, I'll merge this.

PR #7577, opened 2022-12-14
[en] 12/14/2022 - add word

AzadehSafakish:
To address this feedback:
image (6)

Tested on the HTTPServer, result should be:
image


AzadehSafakish:
Oddly enough, adding "non-Zoox" or both words does not remove the offending suggestions:
image (7)
image (8)

I don't believe "non-Jew(s)" warrants an addition to the naughty list in `AbstractEnglishSpellerRule.java`.
>May a non-Jew wear a Talit?
Teaching the Torah to non-Jews

etc.

PR #7586, opened 2022-12-16
[en] added rule

evan-defran-lt:


AzadehSafakish (approved):
One quick comment:

`<match no="1" postag="V.*" postag_regexp="yes>`
is useful for when you want to take the postag from match 1 and transfer it to a new lemma; say you wanted to replace "knocked on the window" with "tapped on the window".

`<match no="1" postag="V.*" postag_regexp="yes>tap</match>`

However, if you're not changing the lemma and just want to make sure the suggestion has the same postag inflection as the text, you can just use \1.

`Did you mean <suggestion>\1 on</suggestion>?`

should yield the same results.

evan-defran-lt:
Oh neat, I didn't know that!

PR #7709, opened 2023-01-13
feat(core): added createDefaultJLanguageTool to Language

fabrichter:
Allow using shared JLanguageTool instances from rules for various purposes.

@jaumeortola Can you see if this works for you? And if so, switch the other places where this is already done to use the new method?

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/Language.java
+++ b/languagetool-core/src/main/java/org/languagetool/Language.java
@@ -73,6 +73,8 @@ public abstract class Language {
   private static final Pattern APOSTROPHE = Pattern.compile("([\\p{L}\\d-])'([\\p{L}«])",
     Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE);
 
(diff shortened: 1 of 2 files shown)


SteVio89 (approved):

jaumeortola:
Okay. I will try, and it will probably work. 

But I am not sure if this will create just one instance because inside a filter we do `new French()`, which is a new Language(). (Perhaps we can get the language from the Rule() inside the RueMatch()...).

See how it is done for the Tagger or the Synthesizer: 

In FrenchTagger.java:
```
public static final FrenchTagger INSTANCE = new FrenchTagger();
```
And in French.java:
```
  public Tagger createDefaultTagger() {
    return FrenchTagger.INSTANCE;
  }
```

jaumeortola:
Well. Probably it is fine. "static" means that there will be just one instance.

fabrichter:
Static and computeIfAbsent should mean there won't be duplicate instances.
Should I merge this then, or do you want to add something here @jaumeortola ?

jaumeortola (approved):

PR #7712, opened 2023-01-13
Improve Multilanguage detection

SteVio89:


languagetool-server/src/main/java/org/languagetool/server/TextChecker.java
```diff
@@ -747,34 +778,35 @@ private List<CheckResults> getRuleMatches(AnnotatedText aText, Language lang,
...
-            res.addAll(getPipelineResults(entry.getValue().build(), entry.getKey(), motherTongue, params, userConfig, listener));
-          }
-        } catch (Exception e) {
-          log.error("Problem with multilingual mode (preferredLangs=" + preferredLangs+ ", preferredVariants=" + preferredVariants + "), " +
-            "falling back to single language.", e);
-          res.addAll(getPipelineResults(aText, lang, motherTongue, params, userConfig, listener));
-        }
-      }
+      } 
+//      else {
```
fabrichter, line 784: Can we add a note about why this code is commented out, or should it be removed and replaced with an error when the multilingual parameter is set?

languagetool-server/src/main/java/org/languagetool/server/TextChecker.java
```diff
@@ -572,6 +577,32 @@ void checkText(AnnotatedText aText, HttpExchange httpExchange, Map<String, Strin
...
+      long startTimeRecheck = System.currentTimeMillis();
+      Map<String, List<Range>> rangesOrderedByLanguage = new HashMap<>();
+      res.forEach(checkResults -> {
+        checkResults.getIgnoredRanges().forEach(range -> {
+          List<Range> sentenceRanges = rangesOrderedByLanguage.getOrDefault(range.getLang(), new ArrayList<>());
+          sentenceRanges.add(range);
+          rangesOrderedByLanguage.put(range.getLang(), sentenceRanges);
+        });
+      });
+      rangesOrderedByLanguage.forEach((shortLangCode, ranges) -> {
```
fabrichter, line 592: I don't think appending all ranges together here is correct when checking the text, since these could be distinct sentences separated by text in other languages; this might lead to e.g. false positives from text-level rules.

fabrichter (changes requested):

languagetool-server/src/main/java/org/languagetool/server/TextChecker.java
SteVio89: You are right. As this should not run (the parameter MultiLangEnabled is not used now), I will remove this part.

fabrichter (approved):

PR #7899, opened 2023-03-02
[nl] added MEN_MIJN

mark-baas:


languagetool-language-modules/nl/src/main/resources/org/languagetool/rules/nl/grammar.xml
```diff
@@ -350,6 +350,17 @@ regexp case-insensitive maken: (?i)
...
 
     <category id="GRAMMAR" name="Grammaticaregels">
+        <rule id="MEN_MIJN" name="Appels zijn men (mijn) favoriete fruitsoort" type="misspelling" default="temp_off">
+            <pattern>
+                <token>zijn</token>
+                <marker>
+                    <token>men</token>
+                </marker>
+            </pattern>
+            <message>Misschien bedoelt u <suggestion>m'n</suggestion> of <suggestion>mijn</suggestion></message>
```
danielnaber, line 360: This is a full sentence, isn't it? Then it probably should end with a `.` (at least that's how we handle it for the other languages).

PR #7904, opened 2023-03-03
[nl] added rule JOU_JOUW_VNW

mark-baas:


languagetool-language-modules/nl/src/main/resources/org/languagetool/rules/nl/grammar.xml
```diff
@@ -384,6 +384,19 @@ regexp case-insensitive maken: (?i)
...
+        <rule id="JOU_JOUW_VNW" name="Waarom stuurt hij jouw (jou) geen berichtjes meer?" type="misspelling" default="temp_off">
+            <pattern>
+                <marker>
+                    <token>jouw</token>
+                </marker>
+                <token postag_regexp="yes" postag="VNW:OBP.*">
+                    <exception postag_regexp="yes" postag="VNW:OBP:ONV|VNW:OBP:VRB"/>
+                </token>
+            </pattern>
+            <message>U bedoelt waarschijnlijk '<suggestion>jou</suggestion>' of '<suggestion>je</suggestion>'.</message>
```
AzadehSafakish, line 396: Putting quotes directly around a `<suggestion>` element will cause the mvn test to fail (build break).
Removing these quotes should fix it.

AzadehSafakish (changes requested):

PR #7948, opened 2023-03-08
[nl] disabled rule KORT_1

mark-baas:
One of the lowest performing rules, disabled for now.


danielnaber:
Thanks, could you add an XML comment explaining why the rule is off now? That will help later when someone looks over the rules.

PR #7950, opened 2023-03-08
[nl] improved rule HEBT_ZONDER_U_JE

mark-baas:
Improved rule, got rid of some FPs

### DIFF
--- a/languagetool-language-modules/nl/src/main/resources/org/languagetool/rules/nl/grammar.xml
+++ b/languagetool-language-modules/nl/src/main/resources/org/languagetool/rules/nl/grammar.xml
@@ -34515,15 +34515,20 @@ steenwijk;Steenwijk
                     <token skip="-1">hebt</token>
                     <token postag="SENT_END"/>
                 </antipattern>
+                <antipattern>
+                    <token skip="-1" regexp="yes">je|jij|jezelf</token>
+                    <token skip="-1">hebt</token>
+                </antipattern>
                 <pattern>
                     <marker>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/nl/src/main/resources/org/languagetool/rules/nl/grammar.xml
```diff
@@ -34515,15 +34515,20 @@ steenwijk;Steenwijk
...
                     <marker>
                         <token case_sensitive="yes">hebt</token>
                     </marker>
                 </pattern>
-                <message>Omdat er 'hebt' staat, moet ergens in de zin ook je, jij, ge, gij of u staan. Of dit is een vervolgzin.</message>
+                <message>Omdat er 'hebt' staat, moet ergens in de zin ook je, jij, ge, gij of u staan. Anders is dit wellicht een vervolgzin.</message>
                 <example type="incorrect">Dan <marker>hebt</marker> je het fout.</example>
                 <example type="incorrect">Dan <marker>hebt</marker> hij het fout.</example>
                 <example>Bedoel je dat je Joe Loop doodgeknuppeld hebt, en zijn lijk in het park <marker>hebt</marker> gedumpt?</example>
+                <example>Als je zorgen <marker>hebt</marker></example>
```
danielnaber, line 34531: BTW, you can also move this example directly before the `</antipattern>`. This way, it's even more clear that the example belongs to a specific antipattern (won't make a difference to the logic, but increases readability of the rule).

PR #7977, opened 2023-03-09
[nl] improved rule DESBETREFFEND

mark-baas:


languagetool-language-modules/nl/src/main/resources/org/languagetool/rules/nl/grammar.xml
```diff
@@ -25758,9 +25758,8 @@ steenwijk;Steenwijk
                 <pattern>
                     <token regexp="yes">desbetreffend|desbetreffende</token>
                 </pattern>
-                <message>&vlotter_mess; 'over' of 'de … waarom het gaat', 'deze', 'die' of 'dit'.</message>
-                <url>https://onzetaal.nl/taaladvies/ouderwets-taalgebruik</url>
```
AzadehSafakish, line 25762: Before I merge this one, is the URL (https://onzetaal.nl/taaladvies/ouderwets-taalgebruik) still an appropriate page to link to for this rule?

If it is, it would be best not to delete it.

PR #8023, opened 2023-03-13
[en] Add upstream/VB

### LINKED ISSUE #8022: [en] recognize "upstream" as verb
Nowadays, “upstream” can be used as a verb, but LanguageTool won't recognize it as one.
An example sentence is: “I would like to upstream the colour theme I made for KDevelop.”

After looking around for almost two hours, I still have no idea where this change/addition would go.

Auswahl_066

JulianGro:
Fixes #8022


tiff:
Thanks! However, you will also have to add the other verb forms "upstreamed", "upstreams", etc.

tiff:
FYI I have manually added these tags

PR #8085, opened 2023-03-20
[en] added rule

evan-defran-lt:
I'm going to work on making this more generalizable on Tuesday.


AzadehSafakish:
Nice one! I can't believe we don't catch the drank/drunk example.
It might be worth it to prioritize this one over `HAVE_PART_AGREEMENT` because of the more specific message, it's up to you.

AzadehSafakish (approved):

PR #8118, opened 2023-03-23
[en] added APs for AI_EN_LECTOR_REPLACEMENT_NOUN_NUMBER

evan-defran-lt:
I'm not sure why the sentence __<example correction="">Movie <marker>film</marker> seen in theaters in the United States runs at 24 frames per second, which is sufficient to create the illusion of continuous movement.</example>__

is not picked up by this AP:
```
            <rule>
                <pattern>
                    <token chunk="B-NP-singular"/>
                        <marker>
                            <and>
                                <token chunk="E-NP-singular"/>
                                <token postag="NN:UN"/>
                            </and>
                        </marker>
                </pattern>
            <example correction="">We are gearing to win some Court cases, but the <marker>case</marker> is dragging on.</example>
            <example correction="">His <marker>composition</marker> is by far the best of all.</example>
            <example correction="">How are you getting along with your <marker>study</marker>?</example>
            </rule>
```
__movie__ is chunked as `B-NP-singular` and __film__ is chunked as `E-NP-singular` and tagged as `NN:UN`.

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/remote-rule-filters.xml
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/remote-rule-filters.xml
@@ -297,6 +297,48 @@ To ignore a remote rule match, set the <marker> so that it exactly covers the te
         <rulegroup id="AI_EN_LECTOR_REPLACEMENT_NOUN_NUMBER" name="">
             <rule>
                 <pattern>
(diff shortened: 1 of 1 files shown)


AzadehSafakish:
image
It finds the error, but it also finds two others (I'm assuming 'illusion' and 'movement').
You'll have to make the AP more specific to catch just 'film', or write a separate AP for this sentence.

PR #8181, opened 2023-04-03
[en] added rule

evan-defran-lt:


evan-defran-lt:
https://github.com/orgs/languagetooler-gmbh/projects/46/views/3?pane=issue&itemId=24330689

PR #8183, opened 2023-04-03
[en] fix message and update URL

AzadehSafakish:
not sure if message is okay


evan-defran-lt:
The message seems fine to me.

PR #8255, opened 2023-04-14
[fr] CONFUSION_E_ER (18) fix

GillouLT:
To fix: https://github.com/languagetooler-gmbh/languagetool-premium/issues/5017

AP modified on CONFUSION_E_ER[18], "skip" +1

Related issue sized as "small" instead of "tiny" because thinking about the possible consequences of raising "skip" from 5 to 6 takes time.


LucieSteib (approved):
Good moove I think, let's monitor that in the future!

PR #8256, opened 2023-04-14
[fr] REP_CERTES Rule creation

GillouLT:
Synonym recommendation

This synonym recommendation already existed, but was working with Java. We wanted to add a link from our FR blog in the suggestion message and it was easier to add a new XML rule for this purpose.

Rule linking issue: https://github.com/languagetooler-gmbh/languagetool-premium/issues/5043


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -178327,5 +178327,17 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
                 <example correction="de flagrant|d'indéniable|d'incontestable|de sûr|de certain">Il n'y a rien <marker>d'évident</marker>.</example>
             </rule>
         </rulegroup>
+            <rule id="REP_CERTES" name="certes (style)" min_prev_matches="1" tags="picky">
```
LucieSteib, line 178330: So I checked, the category is 
`category id="REPETITIONS_STYLE" name="repetitions style" type="style">` so it's indeed a style rule :p

PR #8279, opened 2023-04-18
[fr] Style rules migration Part1

GillouLT:
Context: We are moving all the style rules from Grammar.xml to Style.xml

Why is this Part1? On this first step of the migration, all the categories and rulegroups tagged as "style" are moved to the new Style.xml file

Next steps will involve migrating other rules that aren't tagged as style but are actually style rules.


LucieSteib (approved):
Here we go :) 
Serrons les fesses ;)

PR #8281, opened 2023-04-18
[fr] ALTERNATIVE_OU_POSSIBILITE Improvement

GillouLT:
To fix: https://github.com/languagetooler-gmbh/languagetool-premium/issues/5445

- The issue pointed out that the rule wasn't flagging the plural version "autres alternatives", the rule now does flag it as it should.

- Apart from the issue, the rule was only suggesting the singular version of "possibilité", even if it was preceded by "deux", now it has been improved and suggests the correct number form.
It wasn't a big problem, if the user selected LT suggestion "possibilité" where it should be "possibilités", the plural agreement rule triggered anyway, but it was a 2 steps interaction.


LucieSteib:
@GillouLT sorry again, still need this PR?

GillouLT:
I can merge if you don't see any issue with it :)

PR #8313, opened 2023-04-24
[fr] TOUT_LE_MONDE_SONT Rule creation

GillouLT:
To fix: https://github.com/languagetooler-gmbh/languagetool-premium/issues/5375

The rule flags the string "tout le monde sont" (except when preceded by "par").
If a wrong agreement exists with the following verb, another rule will take over and suggest the right correction.


LucieSteib (approved):
praise: ouh that's cool. LGTM 🥜🥜🥜

PR #8317, opened 2023-04-24
[de] edited regexp to avoid duplicates in mvn

Luisa-LT:
image


Luisa-LT:
Checked with Agnes - Agnes pushes changes

PR #8318, opened 2023-04-24
[fr] PAS_DE_VIRGULE AP added

GillouLT:
To fix: https://github.com/languagetooler-gmbh/languagetool-premium/issues/4995

- AP added, it may seem too specific, but the comma rule is critical and I don't feel confident modifying existing APs.


LucieSteib (approved):
praise: very good moove! I don't see any lost of TPs coming out of it. LGTM 🥜🥜🥜

PR #8371, opened 2023-04-30
[be] Fix orthography and update BE grammar

PlagaMedicum:
- Fixed some orthography issues in messages (Switched from the Classical orthography to the Academic),
- Updated orthography rules (RI, I, U_KAROTKAJE),
- Added rules when latin and cyrillic script are mixed together,
- Add new style and phraseology rules (Style rules based on the Russian analogues)


danielnaber:
Thanks. Could you remove the change to the `pom.xml` from the PR? The "maintainer" property is only meant for people who contribute regularly and over a long time and who have been designated maintainers.

PlagaMedicum:
Sure thing, here you go!

Issue #8362, opened 2023-04-28
[de] Partei/Partie


Discostu36:
LanguageTool should have rule that detects typos with "Partie/Partei".

> Welche **Partie** hast du bei der letzten Bundestagswahl gewählt?

> Lass uns eine **Partei** Schach spielen.

Don't know if this is a case for statistical detection or a manual rule.

Discostu36:
Is there a reason why this is still `temp_off`?

danielnaber:
Maybe it was just forgotten, sorry. Luisa should be back in a few days to have a look. There's one false alarm, BTW: https://internal1.languagetool.org/regression-tests/via-http/2023-05-23/de-DE_full/result_grammar_PARTIE_VS_PARTEI%5B2%5D.html

PR #8387, opened 2023-05-03
[de] added rule group (temp_off)

Luisa-LT:
https://github.com/languagetool-org/languagetool/issues/8362


Luisa-LT:
https://github.com/languagetool-org/languagetool/commit/831f093204173c7cbf1367adba1e38bf556c1b1f

PR #8392, opened 2023-05-03
[fr] PEUT_PEUX_PEU fix

GillouLT:
To fix: https://github.com/languagetooler-gmbh/languagetool-premium/issues/5574

- "Petit" has been added to all APs to avoid FPs
- 2 Tokens added with min="0" in the pattern to take in account both "tout" and "petit" (is it a bad practice?)
- Example added

Will need review after nightly diff


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -23219,6 +23219,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
                 <pattern>
                     <token regexp="yes">très|trop|un|de|quelque|pour|à|depuis|sous|avec|le
                         <exception scope="previous" regexp="yes">.+'</exception></token>
+                    <token min="0">tout</token>
```
LucieSteib, line 23222: thought(unblocking): I don't see any problem with that right now. As you said, let's see the output of the nightly! 🥜

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -23227,6 +23229,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
                 <message>Un adverbe semble plus approprié qu'un verbe.</message>
                 <suggestion>peu</suggestion>
                 <example correction="peu">Il travaille trop <marker>peut</marker>.</example>
+                <example correction="peu">Juste un tout petit <marker>peut</marker>.</example>
```
LucieSteib, line 23232: Yeah indeed, that's needed, one way or another

LucieSteib (approved):
praise:this is great! 🥜🥜🥜

PR #8394, opened 2023-05-04
[fr] Style migration batch1

GillouLT:
Related to: https://github.com/languagetooler-gmbh/languagetool-premium/issues/5459

First batch migration for untagged style rules from grammar.xml to style.xml.
This first batch contains the rules that we consider 100% style related.

- Total rules migrated: 15
- New IDs and Names created for rules that were separated from a rulegroup


LucieSteib (approved):
praise: excellent job! I do think all the rules are moved, no one was left over. :D 🥜
though: I liked the choices of new names, and actually if we need to change them, we could do it while tagging the rules in a near future.

PR #8439, opened 2023-05-10
[fr] SOCIOCULTUREL rule creation

LucieSteib:
To fix [[fr] TIRET[15] sociaux-culturels❌, socio-culturels❌, socioculturels✅#4926](https://github.com/languagetooler-gmbh/languagetool-premium/issues/4926).


DominiquePelle-TomTom:
From the screenshot, I see that "sociaux-culturels" is tokenized as 2 words, but your pattern is trying to match it as a single word, so it will never match.

LucieSteib:
> From the screenshot, I see that "sociaux-culturels" is tokenized as 2 words, but your pattern is trying to match it as a single word, so it will never match.

Ah! Yes, indeed. So annoying it can't be treated as a string ^^'
Maybe if I add sociaux-culturel to the dictionary files, as one token, that'll become a valid option.

DominiquePelle-TomTom:
> Maybe if I add sociaux-culturel to the dictionary files, as one token, that'll become a valid option.

Why not modify your pattern to have multiple tokens instead?
Not tested but judging from other existing rules, something like this maybe:
```
            <pattern>
                <token>sociaux</token>
                <token spacebefore="no">-</token>
                <token>culturels</token>
            </pattern>
```

LucieSteib:
> > Maybe if I add sociaux-culturel to the dictionary files, as one token, that'll become a valid option.
> 
> Why not modify your pattern to have multiple tokens instead? Not tested but judging from other existing rules, something like this maybe:
> 
> ```
>             <pattern>
>                 <token>sociaux</token>
>                 <token spacebefore="no">-</token>
>                 <token>culturels</token>
>             </pattern>
> ```

Yes, better idea :) 
Thanks!

PR #8622, opened 2023-05-15
[fr] Style_Tagging First part Repetitions

GillouLT:
- To address: https://github.com/languagetooler-gmbh/languagetool-premium/issues/5658
- The rules from the repetition category have been tagged
- Off and Temp_off rules haven't been tagged, that would be done later on
- Other rules from the bottom part of the document have been tagged


LucieSteib (commented):
praise:Good job! 🥜🥜🥜
though(unblocking): maybe we should one day check all the REP_SOMETHING to see how bad/good they behave in real life (but I do guess most of them are okay).

PR #8627, opened 2023-05-16
[pt] Modify pt grammar to address issue #5572

p-goulart:
Closes languagetooler-gmbh/languagetool-premium#5572.

 - when replacing incorrect prepositions, the corrected forms were not taking into account obligatory contractions;

 - as such, 'com a' was being corrected to 'a a', instead of 'à';

 - the solution required many different rules, as the substitutions here are not trivial (a + a = à, but a + o = ao), and this was further complicated by special forms of the 'com' preposition ('comigo', 'conosco', etc.).

I do still have some questions, though, so bear with me:
* the suggestions frequently include a lot more than what is to be corrected, e.g. `habituar-me com a` -> `habituar-me à`; when writing rules, I've changed those to simply `com a` -> `à`, but this could have been part of a broader policy – is it?
* there's behaviour for the 'com calma' phrase I can't quite explain – I expect it to be flagged as an error, but it somehow isn't... I suspect something was specifically done to achieve this behaviour, as this phrase is given in the examples;
* ~I'm still not super happy about the results of the 'sobrepor' rule, suppose I'll need to keep tweaking that~ changing the skip value from `-1` to `3` seems to have done the trick.

... also the issue tagging in the commit message is incorrect, I wanted to refer to [this](https://github.com/languagetooler-gmbh/languagetool-premium/issues/5572); will need to work out a way to refer to issues from that board smartly, cos the GH markdown for that is massive.

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -11889,12 +11889,25 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                     <token min='0' regexp='yes'>me|te|se|nos|vos</token>
                     <token>com</token>
                     <token regexp="yes">a(quel[ea])?s?|os?</token>
+                    <example>Já me habituei com a mudança.</example>
+                    <example>Já me habituei com as mudanças.</example>
+                    <example>Você vai se habituar com o novo comandante.</example>
+                    <example>Você vai se habituar com os novos recrutas.</example>
+                    <example>Nós nos habituamos com aquela rotina.</example>
+                    <example>Nós nos habituamos com aquelas pessoas.</example>
+                    <example>Nós nos habituamos com aquele equipamento.</example>
+                    <example>Nós nos habituamos com aqueles computadores.</example>
                 </antipattern>
                 <antipattern>
                     <token inflected='yes'>habituar</token>
                     <token min='0' regexp='yes' spacebefore='no'>&hifen;</token>
                     <token min='0' regexp='yes'>me|te|se|nos|vos</token>
                     <token regexp="yes">co(m|n[st])igo|conv?osco</token>
+                    <example>Ela já se habituou comigo.</example>
+                    <example>Ela já se habituou contigo.</example>
+                    <example>Ela já se habituou consigo.</example>
+                    <example>Ela já se habituou conosco.</example>
+                    <example>Ela já se habituou convosco.</example>
                 </antipattern>
                 <pattern>
                     <token inflected='yes'>habituar</token>
@@ -12014,6 +12027,10 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                     </token>
                     <token>sobre</token>
                     <token regexp="yes">a(quel[ea])?s?|os?</token>
+                    <example>Está sobreposto sobre a segunda camada.</example>
+                    <example>Eu a sobrepus sobre as outras.</example>
+                    <example>Agora sobreponha esta sobre aquela.</example>
+                    <example>Devemos sobrepor o primeiro círculo sobre aquele quadrado.</example>
                 </antipattern>
                 <pattern>
                     <token inflected='yes' regexp='yes' skip="3">sobrep[oô]r
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -11882,7 +11882,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                 <example>A Literatura obedece a leis inflexíveis: a da herança, a do meio, a do momento.</example>
             </rule>
             <!-- Moved from the PHRASAL_VERBS_ERRORS rulegroup above and split to handle contractions. -->
-            <rule>
+            <rule default="off">
                 <antipattern>
                     <token min='0' regexp='yes'>me|te|se|vos</token>
                     <token inflected='yes'>habituar</token>
@@ -11913,7 +11913,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                 <example correction='a'>Tenho dificuldades em habituar-me <marker>com</marker> novas rotinas.</example>
                 <example>Consegues habituar-te com calma ao clima.</example>
             </rule>
-            <rule>
+            <rule default="off">
                 <pattern>
                     <token min='0' regexp='yes'>me|te|se|vos</token>
                     <token inflected='yes'>habituar</token>
@@ -11929,7 +11929,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                 <example correction='à'>Você conseguiu se habituar <marker>com a</marker> nova regra?</example>
                 <example correction='àquele'>Eles já se habituaram <marker>com aquele</marker> professor.</example>
             </rule>
-            <rule>
+            <rule default="off">
                 <pattern>
                     <token min='0' regexp='yes'>me|te|se|vos</token>
                     <token inflected='yes'>habituar</token>
@@ -11944,7 +11944,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                 <suggestion>a\6</suggestion>
                 <example correction='ao'>Você conseguiu se habituar <marker>com o</marker> clima?</example>
             </rule>
-            <rule>
+            <rule default="off">
                 <pattern>
                     <token min='0' regexp='yes'>me|te|se|vos</token>
                     <token inflected='yes'>habituar</token>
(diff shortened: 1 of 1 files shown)


p-goulart:
@maphjo @susanaboatto 🙏 ⬆️

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -11906,9 +11881,198 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                 <example>Torna-se necessário obedecer a normas de conduta.</example>
                 <example>A Literatura obedece a leis inflexíveis: a da herança, a do meio, a do momento.</example>
             </rule>
-        </rulegroup>
+            <!-- Moved from the PHRASAL_VERBS_ERRORS rulegroup above and split to handle contractions. -->
+            <rule default="off">
```
maphjo, line 11885: Please change to `<rule default="temp_off">` for all instances. This tag allows us to let the rule run over a big corpus internally to find additional false positives, before we publish it for all users.
`default="off"` completely deactivates the rule.

maphjo (changes requested):
Let's have @susanaboatto properly review the rules, when she's back next week.

maphjo:
@p-goulart: great work!! :tada: 
@susanaboatto will be back on Monday to complete the review.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: Oops, my bad! Fixed.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -11906,9 +11881,193 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
...
+                </pattern>
+                <message>Neste contexto, o verbo &quot;habituar&quot; rege-se com a preposição &quot;a&quot;.</message>
+                <suggestion>a</suggestion>
+                <example correction='a'>Ainda não me habituei <marker>com</marker> ele.</example>
+                <example correction='a'>Tenho dificuldades em habituar-me <marker>com</marker> novas rotinas.</example>
+                <example>Consegues habituar-te com calma ao clima.</example>
+            </rule>
+            <rule>
+                <pattern>
+                    <token min='0' regexp='yes'>me|te|se|vos</token>
```
susanaboatto, line 11918: I think the first `<token min='0' regexp='yes'>me|te|se|vos</token>` in the rules are useless. If min="0" is active, then it doesn't matter what comes before the verb "habituar". We could remove this line altogether and test it like this, or we could create further subrules to filter cases including pronouns.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -11906,9 +11881,193 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
...
+                    <token min='0' regexp='yes' spacebefore='no'>&hifen;</token>
+                    <token min='0' regexp='yes'>me|te|se|nos|vos</token>
+                    <marker>
+                        <token>com</token>
+                        <token regexp="yes">a(quel[ea])?s?</token>
+                    </marker>
+                </pattern>
+                <message>Neste contexto, o verbo &quot;habituar&quot; rege-se com a preposição &quot;a&quot;.</message>
+                <suggestion><match no="6" regexp_match="a(.*)" regexp_replace="à$1"/></suggestion>
+                <example correction='à'>Você conseguiu se habituar <marker>com</marker> a nova regra?</example>
```
susanaboatto, line 11929: The `<marker>` in the examples should include the determiners, e.g. `<example correction='à'>Você conseguiu se habituar <marker>com a</marker> nova regra?</example>`, or else we go back to the initial problem where multiple rules are triggered separately.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -11906,9 +11881,193 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
...
+                    <token min='0' regexp='yes'>me|te|se|nos|vos</token>
+                    <marker>
+                        <token>com</token>
+                    </marker>
+                </pattern>
+                <message>Neste contexto, o verbo &quot;habituar&quot; rege-se com a preposição &quot;a&quot;.</message>
+                <suggestion>a</suggestion>
+                <example correction='a'>Ainda não me habituei <marker>com</marker> ele.</example>
+                <example correction='a'>Tenho dificuldades em habituar-me <marker>com</marker> novas rotinas.</example>
+                <example>Consegues habituar-te com calma ao clima.</example>
```
susanaboatto, line 11914: There is an antipattern/exception missing for `<example>Consegues habituar-te com calma ao clima.</example>`.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -11906,9 +11881,193 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
...
+            <!-- Moved from the PHRASAL_VERBS_ERRORS rulegroup above and split to handle contractions. -->
+            <rule>
+                <antipattern>
+                    <token min='0' regexp='yes'>me|te|se|vos</token>
+                    <token inflected='yes'>habituar</token>
+                    <token min='0' regexp='yes' spacebefore='no'>&hifen;</token>
+                    <token min='0' regexp='yes'>me|te|se|nos|vos</token>
+                    <token>com</token>
+                    <token regexp="yes">a(quel[ea])?s?|os?</token>
+                </antipattern>
```
susanaboatto, line 11893: You can add `<example></example>` to your antipatterns to see if they work. This helps a lot with testing!

susanaboatto (changes requested):
Hi @p-goulart, great work! I left some comments for review 😄

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: Fair – these were there from the previous version of the rule, which, admittedly, repeated the whole structure in the suggestion.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: Screenshot 2023-05-22 at 10 25 53 AM

susanaboatto: I tested this locally, and the error is still triggered for this sentence:

```
org.languagetool.rules.patterns.PatternRuleTest$PatternRuleTestFailure: Test failure for rule PHRASAL_VERB_A2[1] in file /org/languagetool/rules/pt/style.xml (line 21639): Did not expect error in:
  Consegues habituar-te com calma ao clima.
  Analyzed token readings: [/SENT_START*] Consegues[conseguir/VMIP2S0*]  [ /null*] habituar[habituar/VMN0000,habituar/VMN01S0,habituar/VMN03S0,habituar/VMSF1S0,habituar/VMSF3S0] -[-/_PUNCT*] te[te/PP2CS000*]  [ /null*] com[com calma/RG]  [ /null*] calma[com calma/RG]  [ /null*] ao[a:o/SPS00:DA0MS0]  [ /null*] clima[clima/NCMS000] .[./SENT_END*,./_PUNCT_PERIOD*,./_PUNCT*]
```

So it needs `<token negate="yes">calma</token>` after the marker. The rule editor isn't always accurate. In this case, it could have something to do with the multi-word tagging of `com calma` (tagged as `RG`, not `SPS00` + `NCFS000`).

You can search for `PortuguesePatternRuleTest` on Intellij to run the test locally.

p-goulart: Right, I *did* run the local tests earlier, though, and everything passed. I've just tried changing `calma` to something else *and only then* does it actually fail. Likewise, the same tests were run on (Circle)[https://app.circleci.com/pipelines/github/languagetool-org/languagetool/22048/workflows/efc218cc-351e-404d-8fcf-175792e2c025/jobs/22182], and they're also fine 🤔 Are you running the tests on my branch?

It will probably not *hurt* to negate `calma`, sure, it just doesn't seem like things work very consistently for this one construction and I couldn't find anything specific to it... I mean, there are tonnes of adverbial phrases beginning with `com` and apparently they're tagged differently.

p-goulart: It's done, along with a few other common nouns to be found after `com` in adverbial phrases.

susanaboatto: Ah, found the source. I only ran the test on your changes, not the whole rule. 
`com calma` (`RG` + `RG`) apparently was the only exception that worked because of this antipattern, which is valid for the entire rule group:

 ```
<antipattern>
         <token postag_regexp='yes' postag='R.+'>
          <exception negate_pos='yes' postag_regexp='yes' postag='R.+'/></token>
          <token postag_regexp='yes' postag='R.+'>
          <exception negate_pos='yes' postag_regexp='yes' postag='R.+'/></token>
</antipattern>
```

We are good to go. We should get some diff results by Wednesday, so perfect timing 😄

p-goulart:
Okay, now I think I understand. Did you merge it because I don't have write perms? I was actually curious about what the merge policy here is – I'm personally used to squashing.

PR #8682, opened 2023-05-23
[pt] Add DOUBLE_INFINITIVE rulegroup

p-goulart:
- closes languagetooler-gmbh/languagetool-premium#5538;

 - really simple initial version taking care only of clearly ungrammatical epistemic modal double infinitives;

 - e.g. 'ele dever ter ido embora', ~'poder ter chovido mais cedo'~

 - methinks it's safe to assume these are typos.

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -20319,7 +20319,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                     <example>A cidade ficou-lhe a dever não ter sido invadida.</example>
                 </antipattern>
                 <pattern>
-                    <token postag_regexp="yes" postag="SENT_START|[DNPARZ].+" min="1" max="-1"/>
+                    <token postag_regexp="yes" postag="SENT_START|[DNPARZ].+"/>
                     <marker>
                         <token postag_regexp="yes" postag="VMN.+">dever</token>
                     </marker>
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -20287,7 +20287,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                 <antipattern>
                     <token regexp="yes">fac?to|questão|possibilidade|hipótese|apesar</token>
                     <token regexp="yes" skip="-1" postag_regexp="yes" postag="SPS00(:.+)?">
-                        d([eoa]|es[st][ea]|aquel[ea]|uma?|is[st]o)
+                        d([eoa]|es[st][ea]|aquel[ea]|um(as?)?|uns|is[st]o|aquilo)
                         <exception postag_regexp="yes" postag="V.+"/>
                     </token>
                     <token postag_regexp="yes" postag="VMN.+">dever</token>
@@ -20319,7 +20319,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                     <example>A cidade ficou-lhe a dever não ter sido invadida.</example>
                 </antipattern>
                 <pattern>
-                    <token postag_regexp="yes" postag="SENT_START|[DNPARZ].+" min="1"/>
+                    <token postag_regexp="yes" postag="SENT_START|[DNPARZ].+" min="1" max="-1"/>
                     <marker>
                         <token postag_regexp="yes" postag="VMN.+">dever</token>
                     </marker>
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -20273,7 +20273,46 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
             </rule>
         </rulegroup>
 
-
+        <!-- My assumption here is that these are mostly typos since they're pretty obvious ungrammatical,
+             from the descriptive point of view. So it doesn't seem to be worthwhile to account for non-3SG
+             inflections. I mean, 'dever' -> 'deve' is pretty straightforward, but 'dever' -> 'devo' less so.  -->
+        <rulegroup id="DOUBLE_INFINITIVE" name="Infinitivo duplo">
+            <short>Possível erro de digitação.</short>
+            <!-- This contains the original structure as described in premium issue 5538
+                 All sentences herein are of the shape: 'dever ter feito', and as such will usually indicate
+                 some kind of past tense epistemic modality, for which the double infinitive is very probably ungrammatical. -->
+            <rule default="temp_off">
+                <!-- The only antipattern here refers to constructions of the type 'the fact that...', for which
+                     a personal (i.e. inflected) infinitive can be used. -->
+                <antipattern>
+                    <token regexp="yes">fato|questão|possibilidade</token>
+                    <token regexp="yes" skip="-1">
+                        d[eoa]
+                        <exception postag_regexp="yes" postag="V.+"/>
+                    </token>
+                    <token postag_regexp="yes" postag="VMN.+" regexp="yes">dever|poder</token>
+                    <token>ter</token>
+                    <token postag_regexp="yes" postag="V.P.+"/>
+                    <example>A questão de o segredo dever ter sido revelado não nos afeta.</example>
+                    <example>O fato do contrato já dever ter sido assinado é irrelevante.</example>
+                    <example>A possibilidade de ele poder ter ido mais cedo certamente existe.</example>
+                </antipattern>
+                <pattern>
+                    <token postag_regexp="yes" postag="SENT_START|[DNPARZ].+" min="1"/>
+                    <marker>
+                        <token postag_regexp="yes" postag="VMN.+" regexp="yes">dever|poder</token>
+                    </marker>
+                    <token>ter</token>
+                    <token postag_regexp="yes" postag="V.P.+"/>
+                </pattern>
+                <message>O duplo infinitivo em &quot;<match no="2"/> <match no="3"/> <match no="4"/>&quot; parece indicar um error de digitação. Se for o caso, substituta &quot;<match no="2"/>&quot; por &quot;<suggestion><match no="2" postag_regexp="yes" postag="VMN.+" postag_replace="VMIP3S0"/></suggestion>&quot;.</message>
+                <example correction="deve">Ele já <marker>dever</marker> ter ido embora.</example>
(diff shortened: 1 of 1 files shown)


p-goulart:
(also I still can't add reviewers, labels, etc. 😭)

danielnaber:
> also I still can't add reviewers, labels

Please try again now.

p-goulart:
> > also I still can't add reviewers, labels
> 
> Please try again now.

Works a charm, thanks!

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -20273,7 +20273,66 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
...
+            <!-- This contains the original structure as described in premium issue 5538
+                 All sentences herein are of the shape: 'dever ter feito', and as such will usually indicate
+                 some kind of past tense epistemic modality, for which the double infinitive is very probably ungrammatical. -->
+            <rule default="temp_off">
+                <!-- This antipattern refers to constructions of the type 'the fact that...', for which
+                     a personal (i.e. inflected) infinitive can be used. -->
+                <antipattern>
+                    <token regexp="yes">fac?to|questão|possibilidade|hipótese|apesar</token>
+                    <token regexp="yes" skip="-1" postag_regexp="yes" postag="SPS00(:.+)?">
+                        d([eoa]|es[st][ea]|aquel[ea]|uma?|is[st]o)
```
susanaboatto, line 20290: We can add more instances: `d([eoa]s?|es[st][ea]s?|aquel[ea]?|aquilo?|uma?|uns|um|is[st]o)`. This could also become an entity, as there are more rules which might require limiting the `SPS*` tag to this.

susanaboatto: Checking the pattern now, this rule does not trigger `d([eoa]s?|es[st][ea]s?|aquel[ea]?|aquilo?|uma?|uns|um|is[st]o)` anyway because the first token in the pattern is limited to `SENT_START|[DNPARZ].+`. So are the antipatterns necessary?

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -20273,7 +20273,66 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
...
+                <antipattern>
+                    <token postag="SPS00">a</token>
+                    <token postag_regexp="yes" postag="VMN.+">dever</token>
+                    <token postag_regexp="yes" postag="R." min="0"/>
+                    <token>ter</token>
+                    <token postag_regexp="yes" postag="V.P.+"/>
+                    <example>A cidade ficou-lhe a dever não ter sido invadida.</example>
+                </antipattern>
+                <pattern>
+                    <token postag_regexp="yes" postag="SENT_START|[DNPARZ].+" min="1"/>
```
susanaboatto, line 20322: `min="1"` is not really necessary here

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -20273,7 +20273,66 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
...
+                    <marker>
+                        <token postag_regexp="yes" postag="VMN.+">dever</token>
```
susanaboatto, line 20324: Can we add more verbs? I can think of other fitting examples: `dever|poder|querer|pretender`

Note that `VMN.+` will also trigger a mistake like `deverem ter chegado lá`, which would likely need a different correction. `VMN0000` might be more fitting (but we could also leave it like this at first and see what kind of results we get in the diff, then add changes later)

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -20273,7 +20273,66 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
...
+                    </marker>
+                    <token postag_regexp="yes" postag="R." min="0"/>
+                    <token>ter</token>
+                    <token postag_regexp="yes" postag="V.P.+"/>
+                </pattern>
+                <message>O duplo infinitivo parece indicar um error de digitação. Se for o caso, substituta &quot;\2&quot; por <suggestion><match no="2" postag_regexp="yes" postag="VMN.+" postag_replace="VMIP3S0"/></suggestion>.</message>
```
susanaboatto, line 20330: More suggestions (such as `deveria`) would likely increase apply rates. But this is also something we could check in the diff results. I'll leave it up to you if you want to leave it as it is.

susanaboatto (changes requested):
🙌

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: I started with `dever|poder` and there were some instances of 'poder' that just didn't quite fit the structures I had designed for epistemic modals. (I mean, 'poder' is also an epistemic modal often, but more frequently it's simply a potential mood marker...)

As such I thought it was a better idea to start small with `dever` only, as in the original issue, and then gradually expand to other verbs.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: My initial assumption is that these instances of `dever` are typos for `deve`, not `deveria`. That's just a hunch, we'd need some real-world usage data to confirm, of course. I could totally add `deveria` to the suggestions, though, it's a quick thing to add.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: Not sure what you mean. The first token of the pattern includes `N.+` postags, which should match the words in the first antipattern token. Or am I seeing things? Seemed to work as expected as I was testing it.

susanaboatto: Nope, at least here `N.+` doesn't show up, only` SPS*` and `V*`:
image

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
susanaboatto: We could leave it as it is and try to expand the rule later. `poder` would require some antipatterns or another opening token (`negate` + `SPS.+`, for instance).

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
susanaboatto: I think this is mostly the case, but in the era of “auto complete” features, I feel some people don't always type whole words all the time. We can see what kind of results we get and decide later.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: anigif_sub-buzz-15209-1490296003-28

susanaboatto: Just realized I missed the `skip="-1"` in the antipattern. 

Just adding `d([eoa]s?|es[st][ea]s?|aquel[ea]?|aquilo?|uma?|uns|um|is[st]o)` should do the deed 😄.

p-goulart: I really don't see what wouldn't be working. The tests pass, and when I change it they don't.

susanaboatto: Just realized I completely missed `skip="-1"` in the antipattern. Let's just add `d([eoa]s?|es[st][ea]s?|aquel[ea]?|aquilo?|uma?|uns|um|is[st]o)`.

susanaboatto: And also `um` and the plural form of the others (this is what we have right now:` d([eoa]|es[st][ea]|aquel[ea]|uma?|is[st]o`)

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: It __seems__ to work as intended, i.e. at least one and at most infinity of tokens tagged with that regex. AFAIU the `skip` attribute would better fit the **next** token – but I think when I tried that it didn't quite do what I wanted? 😠

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: Oh, right, now I remembered. I had originally chosen not to accept the plural forms precisely because I assumed that `dever` was a misspelling for `deve`. Adding plural also tacks on the extra complication of there being a further number agreement error.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
susanaboatto: Did you test `min="1" max="-1"` instead of `min="1"`? If it works, let's see what we get.

p-goulart: I did test it, yes, but now the more I think about it the less sense it makes. Maybe you're right that matching a single one of those tokens should be enough for a healthy match. Sehr verwirrend.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
susanaboatto: True, let's leave it as it is and wait for the results 😄

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
susanaboatto: Two possibilities here:

**1.** add `skip="-1"` and see what happens (and if we'd miss any matches without it)
**2.** remove all the skips in the first token altogether (or limit them to just a few tokens) and just go with that for now

We could also try 1 and then change to 2 if results get too crazy.

p-goulart: Occam's etc., removed the silly min/max in the new commit. As far as I can tell we don't lose any coverage and don't require any new antipatterns.

PR #8683, opened 2023-05-23
set all tone_tags as default

SteVio89:


languagetool-core/src/main/java/org/languagetool/ToneTag.java
```diff
@@ -20,6 +20,13 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 public enum ToneTag {
-  clarity, formal, professional, confident, academic, povrem, scientific, objective, persuasive, informal, povadd, positive, NO_TONE_RULE;
+
+  clarity, formal, professional, confident, academic, povrem, scientific, objective, persuasive, informal, povadd, positive, NO_TONE_RULE, ALL_TONE_RULES;
+
+  public static List<ToneTag> REAL_TONE_TAGS = Arrays.asList(clarity, formal, professional, confident, academic, povrem, scientific, objective, persuasive, informal, povadd, positive);
```
fabrichter, line 31: It would be better to automatically generate this from all values, just excluding special ones.

languagetool-server/src/test/java/org/languagetool/server/HTTPServerTest.java
```diff
@@ -235,25 +235,32 @@ void runTestsV2() throws IOException, SAXException, ParserConfigurationException
 
     Language xxlang = Languages.getLanguageForShortCode("xx");
     String test1 = checkV2(xxlang, allToneTagRulesMatchingSentence, "&disabledRules=test_unification_with_negation");
-    assertTrue(test1.contains("Formal_Clarity_TONE_RULE")); //Clarity is enabled by default if no toneTag is specified
+    assertTrue(test1.contains("Formal_Clarity_TONE_RULE"));
     assertTrue(test1.contains("NO_TONE_RULE"));
-    assertFalse(test1.contains("CONFIDENT_ACADEMIC_SCIENTIFIC_TONE_RULE"));
-    assertFalse(test1.contains("CONFIDENT_ACADEMIC_TONE_RULE"));
+    //assertFalse(test1.contains("CONFIDENT_ACADEMIC_SCIENTIFIC_TONE_RULE"));
```
fabrichter, line 240: Can't the commented out statements be removed?

PR #8704, opened 2023-05-25
Doppeltes verb - added APs (still temp_off)

Luisa-LT:
https://github.com/languagetool-org/languagetool/commit/131b9256f6578e854277a64aa1d7d39984dd774b
Changed prio and added 3 APs (more APs to be added)

Diff: https://internal1.languagetool.org/regression-tests/via-http/2023-05-23/de-DE_full/result_grammar_DOPPELTES_VERB%5B1%5D.html


Discostu36:
I am currently doing a total rewrite of this rule. It will maybe take a few days.

PR #8705, opened 2023-05-25
[pt] Add words to spellcheck

p-goulart:
All of these occurred 15+ times in w20.

- [ ] how crucial is it to keep the file organised? (both in terms of the categories provided and in terms of stuff like alphabetical sorting);
- [ ] how much justification/research do you need for decisions made here? I've added comments only in cases where I myself was kind of like 'huh? what is this?'
- [ ] I've skipped `Profa` and `Drª`, as I think we need to work out a better solution here using the abbreviation rules we already have;
- [ ] I've noticed `hahaha`, and I'm certain we can add some kind of smart-ish 'laughing' rule that ignores stuff like `(h[ea])+h?` – just a thought... it's annoying having to manually ignore laughter... could be a style thing;
- [ ] the 'adequa' -> 'adéqua' thing is straight up prescriptivist, but not in a cute way; IMO this should be a style choice, with `adequa` available in all non-ultra-formal registers?
- [ ] `okay` is weird, but I'm not sure where we stand on that type of thing; do we ask to correct to `OK`? I personally write 'okay' in pt-BR but I might just be abroad too long;
- [ ] what in the universe is the PoS of initialisms like `LGBT`? they're used almost exclusively adjectivally, but they're not quite adjectives... they're not exactly adjectival phrases, either.

Some more general comments:
- [ ] the letter-case issue must be addressed somehow... I know I just got here but having to manually (well, I did record Vim macros for this because come on) add `lower`, `Title`, and `UPPER`case variants of every single proper noun does not seem like the most efficient thing;
- [ ] already briefly discussed with @susanaboatto the possibility of keeping a list of words that users keep ignoring but we should continue not to accept, so as not to repeat this every week/month;
- [ ] I'm definitely going to be working on a script to automate at least part of this at some point.


susanaboatto:
1) **Prof, Dr**: Yes, we can add the abbreviation inconsistencies to our list of issues in the feedback board;
2) **hahaha**: That would probably be necessary if we were to add `hahaha` to the `added.txt` file. As of now, I believe only `ha` is in the dictionary. Again, we can add this to the feedback board as a to-do task.
3) **adequa/adéqua**: add an issue to the feedback board, so we can tackle this when creating style rules. But it's okay to add `adéqua` to the tagset, as it is not incorrect.
4) **okay**: personally, I think it sounds like an _estrangeirismo_, but we can add it to the dictionary and make a `formality/shorten_it/foreign_words` rule in the style file if we notice it's causing problems.
5) **LGBT**: It functions as an adjective, so we tag it as a neutral adjective → `AQ0CS0` // `AQ0CP0` (`LGBTs`), or else unwanted rules will be triggered. But we should also tag those as nouns. We can do the same for other variants (LGBTQ, LGBTQIA+, etc.).

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/added.txt
```diff
@@ -8752,3 +8753,106 @@ aliançazinha	aliançazinha	NCFS000
...
+ZAP	ZAP	NPMSO00
+##
+Izabela	Izabela	NPFSS00
+IZABELA	Izabela	NPFSS00
+Avon	Avon	NPFSO00
+AVON	Avon	NPFSO00
+Mark	Mark	NPMSS00
+MARK	Mark	NPMSS00
+Alison	Alison	NPMSS00
+ALISON	Alison	NPMSS00
```
susanaboatto, line 8803: I feel like Alison is a unisex name? Best to tag it as NPCSS00

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/added.txt
```diff
@@ -8752,3 +8753,106 @@ aliançazinha	aliançazinha	NCFS000
 aliançazinhas	aliançazinha	NCFP000
 descompressor	descompressor	NCMS000
 descompressores	descompressor	NCMP000
+home	home	NCMS000
+homes	home	NCMP000
+PIN	PIN	NCMx000
```
susanaboatto, line 8758: Why not tag it as singular? There is a plural form (`PINs`).

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/added.txt
```diff
@@ -8752,3 +8753,106 @@ aliançazinha	aliançazinha	NCFS000
...
+SWITCH	switch	NCMS000
+switches	switch	NCMP000
+Switches	switch	NCMP000
+SWITCHES	switch	NCMP000
+prompt	prompt	NCMS000
+Prompt	prompt	NCMS000
+PROMPT	prompt	NCMS000
+prompts	prompt	NCMP000
+Prompts	prompt	NCMP000
+PROMPTS	prompt	NCMP000
```
susanaboatto, line 8770: I've never used this word in Portuguese, but I'm not sure about its gender. `O prompt` or `a prompt`? Does it even matter? Could also be tagged as `common`. Maybe we should make this a standard for all barbarisms?

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/hunspell/spelling.txt
```diff
@@ -6511,6 +6521,8 @@ putinha
 putinhas
 putona
 putonas
+buceta
+bucetas
```
susanaboatto, line 6525: The correct spelling for this word is with an `o`, not an `u`. I see it does not appear as a suggestion in the editor.

susanaboatto (changes requested):
👀

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/added.txt
p-goulart: Always masculine for me, but no harm in `C`ing it.

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/hunspell/spelling.txt
p-goulart: Should it even appear? Tbh the policy on vulgarity isn't super clear to me, either. I can delete it, I suppose, but I feel like policing the 'proper' spelling of a very common variant of such an informal word is a bit too prescriptivist? I can delete it if that's the policy, but weird.

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/added.txt
p-goulart: Typo

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/added.txt
p-goulart: tbh in Rio I think I only ever met male Alisons, but fair

p-goulart:
also re:`adéqua`, the form that's marked as incorrect **is** actually `adequa`, which is super 🤯 to me; `adéqua` is currently the only correct form (and the one suggested to users)

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/hunspell/spelling.txt
p-goulart: I have, of course, completely neglected the erotica authors!

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/added.txt
```diff
@@ -8752,3 +8753,106 @@ aliançazinha	aliançazinha	NCFS000
 aliançazinhas	aliançazinha	NCFP000
 descompressor	descompressor	NCMS000
 descompressores	descompressor	NCMP000
+home	home	NCCS000
+homes	home	NCCP000
+PIN	PIN	NCMS000
+PINs	PIN	NCMP000
+switch	switch	NCMS000
+Switch	switch	NCMS000
+SWITCH	switch	NCMS000
```
jaumeortola, line 8762: This triple version of words (lower case, first upper case, all upper case) shouldn't be necessary. It will probably create problems in the synthesis of suggestions. 
Is anything not working as expected if these three versions are missing? The speller? Then add the three words to spelling.txt, but not to added.txt.

p-goulart: Right, they're added to the hunspell files but IMO they shouldn't even be required there, which is already something I brought up with @susanaboatto when she was describing to me what was to be done here. I'll describe it in an issue. If this variation causes issues in `added.txt`, then removed they shall be.

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/added.txt
susanaboatto: Yes, but as an international name, it also applies to women, so many people could be using it in pop culture references, etc. 😉

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/added.txt
susanaboatto: Thanks for noticing this @jaumeortola!

jaumeortola: This would eliminate the need to add capitalized and all upper case words to speller.txt. https://github.com/languagetooler-gmbh/languagetool-premium/issues/4703

PR #8706, opened 2023-05-25
[fr] moves DOUBLES_ESPACES from premium to OS

maphjo:


maphjo:
@LucieSteib @GillouLT @danielnaber for visibility

PR #8720, opened 2023-05-29
[pt] Add VERBOS_DE_MOVIMENTO_EM_BR

p-goulart:
Closes languagetooler-gmbh/languagetool-premium#5502.

For now we are contemplating only 'chegar' (the original verb in the grammar file) and 'ir' (the verb from the issue description).

Due to the tricky nature of the morphology of 'ir' (suppletion with 'ser'), as well as its flexible semantics, the actual structures that we cover here are quite restricted. It should address the most obvious errors in very simple sentences, but we may be missing errors in more complex ones. Oh well.

A few comments, *comme d'hab*:
* having to duplicate rules just because of different contractions feels very silly, surely there's a way we can improve this? Can we add some kind of function in the Java somewhere that performs all the necessary substitutions intelligently?
* I strongly believe that, especially in the case of 'chegar', tagging this only with more formal writing styles would be ideal – many of the examples I found in the corpus feel 100% more natural to me with 'em' instead of the technically correct 'a'; wht say you?


susanaboatto:
Yes, this should be in the style.xml file and be tagged as formal. Are you able to open/edit the style-premium.xml file?

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/pt-BR/grammar.xml
```diff
@@ -527,17 +528,374 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+        <!-- situar-se-ia em X: problems with mesoclisis, just being safe -->
+        <antipattern>
+            <token postag_regexp="yes" postag="VMN.+"/>
+            <token spacebefore="no" regexp="yes">&hifen;</token>
+            <token spacebefore="no" postag_regexp="yes" postag="PP.+"/>
+            <token spacebefore="no" regexp="yes">&hifen;</token>
+            <token spacebefore="no" inflected="yes" postag_regexp="yes" postag="VMII.+" skip="-1">
+                ir
+                <exception scope="next" postag_regexp="yes" postag="V.+"/>
+                <exception regexp="yes" scope="next" postag_regexp="yes" postag="SP.+(:.+)?">
```
susanaboatto, line 550: Here, just `SP.+` should be enough, as it already covers `:.+?`

PR #8721, opened 2023-05-29
[pt] Add 'foi' exception to AUXILIARY_VERB_INFINITIVE

p-goulart:
- fixes languagetooler-gmbh/languagetool-premium#5491
 - the false negative was 'vocês virem foi um erro' – added to the examples of the rule after the fix, tests pass and no new false positive as far as I can see; should be safe.
 - sorry about the cosmetic changes in indentation – the only proper changes here are on lines 741 and 779.

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -731,14 +731,17 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                     <token spacebefore="no"/>
                 </antipattern>
                 <pattern>
-                    <token inflected="yes" regexp='yes'>&verbos_auxiliares_aspectuais;|&verbos_auxiliares_modais;
+                    <token inflected="yes" regexp='yes'>
+                        &verbos_auxiliares_aspectuais;|&verbos_auxiliares_modais;
                         <exception inflected='yes'>ser</exception>
                         <exception negate_pos='yes' postag_regexp="yes" postag="V.+"/>
-                        <!--exception regexp='yes'>quer|fora</exception--></token>
+                        <!--exception regexp='yes'>quer|fora</exception-->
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -772,6 +785,16 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 
                 <example>Qualquer um que queira vir será bem-vindo.</example>
                 <example>Até essa altura chegar decorrerão inúmeros conflitos.</example>
+
+                <!-- p5491 -->
+                <example>Vocês virem foi um erro.</example>
+                <!-- p5490: errors are intentional, do not fix -->
+                <example>Ele virá está noite?</example>
+                <example>Quero está.</example>
+                <example correction="Deixe estar"><marker>Deixe está</marker>.</example>
```
susanaboatto, line 794: If we are removing matches with `está`, we still need a separate rule/suggestion for it. The current correction we have may be wrong, but it points out an oddness about it and might still lead users to manually correct the sentence or notice what they wrote is wrong. I can add this to the issue's DoD, if you'd like.

susanaboatto: In any case, we should have it documented somewhere 🙂

p-goulart: I'll just keep adding stuff to the confusion pairs, but I have some ideas about what we can do with diacritic-only confusion pairs based on past experience working with Spanish.

p-goulart: Fixed!

p-goulart:
@susanaboatto anything else to review here? if not, can you please approve before it gets stale? 🥺

susanaboatto (approved):

PR #8745, opened 2023-05-31
[pt] Add and expand misc rules

p-goulart:
- mostly stuff seen in the corpus that I noticed we needed to expand, nothing ground-breaking;

 - really surprised about the lack of 'sera'->'será' correction, but I've triple-checked and it really doesn't get flagged anywhere 🤯

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -20889,33 +20889,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
             </rule>
         </rulegroup>
 
-        <rulegroup id="NO_ACCENT_ON_VERBAL_E" name="Falta de acento agudo sobre 'é'." default="temp_off">
-            <!-- the only consistent false positive here refers to mathematical texts where
-                 'e' is used as a variable, or physics where it is a constant;
-                 IMO that's a an okay thing to have, as those are super specialised
-                 and those people will need to have all sorts of rules turned off
-                 anyway... -->
-            <rule>
-                <pattern>
-                    <token regexp="yes">qu[eê]|quando|qual|onde|quem|quant[oa]s?|como|por[ ]?qu[êe]</token>
-                    <marker>
-                        <token case_sensitive="yes">e</token>
-                    </marker>
-                    <token negate="yes" regexp="yes">qu[eê]|quando|qual|onde|quem|quant[oa]s?|como|por[ ]?qu[êe]</token>
-                </pattern>
-                <message>Possível erro de acentuação. Verifique se não quis dizer <suggestion>é</suggestion>.</message>
-                <example correction="é">Acha que <marker>e</marker> melhor?</example>
-                <example correction="é">O que <marker>e</marker> o problema?</example>
-                <example correction="é">Onde <marker>e</marker> a padaria?</example>
-                <example correction="é">Qual <marker>e</marker> o pior deles?</example>
-                <example correction="é">Quando <marker>e</marker> a festa?</example>
-                <example correction="é">Quanto <marker>e</marker> a medida ideal?</example>
-                <example correction="é">Por que <marker>e</marker> pior fazer do meu jeito?</example>
-                <example>Onde e como assinar?</example>
-            </rule>
-        </rulegroup>
-
-
         <!-- ELE SEGUI OS ele segue os -->
         <rulegroup id='CONFUSÃO_VERBOS_TIPO_SEGUI_SEGUE_SEGUIU' name="Confusão do tipo: segui/segue/seguiu (Verbos)">
             <!-- Created by Marco A.G.Pinto with Ricardo Joseh Lima suggestions, Portuguese rule 2022-11-09 (25-JUL-2022+) -->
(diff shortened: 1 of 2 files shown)


languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -20889,6 +20889,32 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
             </rule>
         </rulegroup>
 
+        <rulegroup id="NO_ACCENT_ON_VERBAL_E" name="Falta de acento agudo sobre 'é'." default="temp_off">
```
susanaboatto, line 20892: There's a similar existing rule that could perhaps be edited CONFUSÃO_E_É. This is part of issue [5564](https://github.com/languagetooler-gmbh/languagetool-premium/issues/5564). If you find more confusion pairs, you can check this issue and/or expand it.

susanaboatto:
There are many uncorrected confusion pairs cases relating to diacritics, like `sera → será`. I have compiled some [in this issue](https://github.com/languagetooler-gmbh/languagetool-premium/issues/5564). You can add more pairs to the issue if you find them.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: Moved it to the already existing rule at d57ece80cf

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/pt-BR/grammar.xml
```diff
@@ -887,34 +889,103 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
-      <pattern>
-        <token>duzentas</token>
-        <token>gramas</token>
-      </pattern>
-      <message>Se grama se refere a peso substitua &quot;duzentas gramas&quot; por <suggestion>duzentos gramas</suggestion>.</message>
-      <example correction="duzentos gramas">Comprei <marker>duzentas gramas</marker> de carne.</example>
+    <rule id="DUZENTAS_GRAMAS_NUMERAL" name="Duzentos gramas" default="temp_off">
+        <pattern>
+            <token regexp="yes">(duz|trez|quatroc|quinh|seic|setec|oitoc|novec)entas</token>
+            <token>gramas</token>
```
susanaboatto, line 895: Would it be too risky to broaden the numbers even more here? `Grass` is not really countable, is it? I feel like if you want to count it, `gramado` is more suitable (possibly a style rule idea?). Anyway, just some thoughts.

p-goulart: Umm... what 'grass' are you referring to? Isn't this about 'grams' as the unit of mass? I personally feel like 'trezentas gramas' isn't particularly less likely to occur than the original 'duzentas gramas', so I just expanded to the other numbers between 300 and 900.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/pt-BR/grammar.xml
```diff
@@ -887,34 +889,103 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+        <pattern>
+            <token>não</token>
+            <marker>
+                <token>e</token>
+            </marker>
+            <token postag_regexp="yes" postag="R." min="0" max="-1"/>
+            <token postag_regexp="yes" postag="A.+"/>
+        </pattern>
+        <message>Possível erro de digitação: será que você quis escrever <suggestion>é</suggestion>?</message>
+        <short>Possível erro de digitação.</short>
```
susanaboatto, line 981: Maybe we could have a comma suggestion here as well to deal with possible ambiguous sentences. 

(But we could also wait and see if such FPs will appear in the diff results in the first place, before we get ahead of ourselves)

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/pt-BR/grammar.xml
susanaboatto: Yes, this is about units, but the point of the rule is to avoid confusion between the two types of `gramas`, isn't it? 

I assumed the matched numbers had been limited to avoid flagging correct sentences where the user actually meant to say `grass` instead of `grams` (e.g. `meu bebê comeu três gramas e um punhado de terra`). 

But apart from cases where there is a determiner (e.g. `Vou colocar uma grama no meu quintal`), that just seems super unlikely to occur. So I wonder if we could attempt expanding the matches and see where that leads.

susanaboatto: I'm thinking of recipes and calorie counting texts, but maybe I'm overthinking this. Anyway, just a thought. The rule is fine.

susanaboatto (approved):

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/pt-BR/grammar.xml
p-goulart: As far as I understood it, the point of the rule was to fix the agreement issue, as 'grama' (the unit) is masculine, not feminine. It didn't even cross my mind that anyone would be counting blades of grass. We can just roll it out and see if it affects anyone – like you, I seriously doubt it. There aren't __that many__ more numerals that would require agreement here. Maybe anything ending with '1' and '2' (for 'vinte e uma gramas' or 'vinte e duas gramas') – but that's not the kind of numeral that's very likely to be written out in full, so I wanted to leave it restricted only to do the full forms of numerals between 200 and 900, which seem more likely.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/pt-BR/grammar.xml
p-goulart: Yes, fair, I'll give myself a to-do for this.

PR #8769, opened 2023-06-02
[pt] Fix spelling warnings in replace files

p-goulart:
- addresses languagetooler-gmbh/languagetool-premium#5759 (the `pt` bit at least);
- in our case, since this also required splitting some of the locale's replace files by specific culture, we also needed minor changes to the code;
- some (though not all) of the issues were caused by actually misspelt words in the source files, those were obvs fixed;
- afaict it looks alright – tests pass and everything compiles locally.

This is, as discussed, in no way the thorough clean-up these files actually merit – that *is* a goal for this sprint, though.

A bit of a to-do for later here is definitely writing some more thorough tests for some of the Portuguese stuff.


languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/added.txt
```diff
@@ -8831,3 +8831,86 @@ mutagênico	mutagênico	AQ0MS0
 mutagênicos	mutagênico	AQ0MP0
 mutagênica	mutagênico	AQ0FS0
 mutagênicas	mutagênico	AQ0FP0
+consomê	consomê	NCMP000
+consomês	consomê	NCMS000
+brailes	brailes	AQ0CP0
```
p-goulart, line 8836:
```suggestion
brailes	braile	AQ0CP0
```

languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/PortugueseClicheRule.java
```diff
@@ -44,16 +41,19 @@ public class PortugueseClicheRule extends AbstractSimpleReplaceRule2 {
 
   public static final String PORTUGUESE_CLICHE_RULE = "PT_CLICHE_REPLACE";
 
-  private static final String FILE_NAME = "/pt/cliches.txt";
+//  private static final String FILE_NAME = "/pt/cliches.txt";
```
danielnaber, line 44: Better to delete this line instead of commenting it out (it's still in the git history, should we ever need it again).

PR #8772, opened 2023-06-04
fix: crash when closing and getting hunspell object

miurahr:
It changes HunspellDictionary interface to add  `HunspellDictionary#isClosed` method to indicate a status of implementation of  `HunspellDictionary`. Hunspell class will check its status. When it is closed status, Hunspell class invalidate cache and create new object. Otherwise, user cause a JVM crash (segmentation fault) by  accessing to freed JNA object. 

Here is an issue on a user project, OmegaT that report crash when reloading a session.
https://sourceforge.net/p/omegat/bugs/1182/

A log message from JVM is here
https://gist.github.com/mohad12211/fe2ebfd71469b876c41a32e84c661e12

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/DumontsHunspellDictionary.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/DumontsHunspellDictionary.java
@@ -40,16 +40,25 @@ public class DumontsHunspellDictionary implements HunspellDictionary {
 
   @Override
   public boolean spell(String word) {
+    if (closed) {
+      throw new RuntimeException("Attempt to use hunspell instance after closing");
+    }
     return hunspell.spell(word);
   }
 
   @Override
   public void add(String word) {
+    if (closed) {
+      throw new RuntimeException("Attempt to use hunspell instance after closing");
+    }
     hunspell.add(word);
   }
 
   @Override
   public List<String> suggest(String word) {
+    if (closed) {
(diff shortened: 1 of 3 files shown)


mohad12211:
Having a `Closeable` resource and a cache that can't be invalidated is problematic. This solution makes sense to me.

miurahr:
> Having a `Closeable` resource and a cache that can't be invalidated is problematic.

Yes, right. Hunspell class make a cache of `Closeable` object using `HashMap` without a way of invalidation. It is a root cause of  the problem.

miurahr:
There are several problems also in dumont `hunspell-java` project.

1. `suggest` method does not check whether `handle` is `null`. `spell` and `add` does.
2. `close` method check nullity of `handle` before calling `HunspellLibrary#close` but it does not set `null` after destroying it.
3.   `HunspellLibrary#close` is a binding to `libhunspell` which release a resource pointed by handle (pointer). It does not do any manipulation of the pointer itself.

No.2 and No.3 lead  null checks in methods to be non-sense.

1. https://gitlab.com/dumonts/hunspell-java/-/blob/master/src/main/java/dumonts/hunspell/Hunspell.java#L76
2. https://gitlab.com/dumonts/hunspell-java/-/blob/master/src/main/java/dumonts/hunspell/Hunspell.java#L95
3. https://github.com/hunspell/hunspell/blob/master/src/hunspell/hunspell.cxx#L2275

miurahr:
Just proposed a fix in upstream project.
https://gitlab.com/dumonts/hunspell-java/-/merge_requests/17

PR #8799, opened 2023-06-08
[pt] Update dictionary for w22

p-goulart:
Week 22 dictionary adds; not all of them, obvs, but a good number. The blacklist is for local use and/or reference for now.

Some of the spelling issues could also be addressed by tweaking abbreviation rules, which I also do here.


languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/blacklist.txt
```diff
@@ -0,0 +1,64 @@
...
+food	foreign term
+Food	foreign term
+Forms	foreign term
+forms	foreign term
+Front	foreign term
+front	foreign term
+Full	foreign term
+full	foreign term
+from	foreign term
+Gonzalez	wrong spelling in foreign language
```
susanaboatto, line 54: I wonder if some of the foreign terms shouldn't be added to the dictionary anyway. For example, I imagine `book` might be used a lot (or relatively often) in the marketing/influencer/photography industries. It's pretty common to come across sentences like `Gostaria de fazer um book de fotos?`

Or proper names like `Gonzalez` could have multiple spellings, so shouldn't we have as many of them as possible be in the dictionary?

PR #8800, opened 2023-06-08
[fr] QUE_SE_PASSE rule creation

maphjo:
I haven't had a good idea for a suggestion, so I used one of the standard phrases from grammar.xml.


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -47892,6 +47892,21 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+        <rule id="QUE_SE_PASSE" name="que/qui se passe" default="temp_off">
+            <pattern>
+                <token>qu'</token>
+                <token>est</token>
+                <token>-ce</token>
+                <marker>
+                    <token>que</token>
+                </marker>
+                <token>se</token>
+                <token>passe</token>
```
LucieSteib, line 47904: suggestion(unblocking): maybe consider turning "passe" to an active flexion like `<token postag="V.*" postag_regexp="yes" inflected="yes">faire</token>` ? "qu'est-ce qui se passait/passera"could also happen 🥜

LucieSteib (approved):

PR #8808, opened 2023-06-10
disambiguation: add chunktags

jaumeortola:
An easy way of adding chunk tags from disambiguation.xml.
This will be very useful for all Romance languages.

`action="addchunk"` will be similar to `action="add"` but the tag will be added to the chunk tags.


danielnaber:
This is now running on `https://api12.languagetool.org/v2/`, please check if it works as expected. The server is stable so far.

jaumeortola:
There is nothing to check online yet. It is not used in any rule so far. Anyway, it works as expected in the rules I tried locally. 

I created the PR to see if it's okay to use this new `action="addchunk"` in disambiguation files. It seems easy and clear enough. We'll just need to add it to the documentation.

Issue #8770, opened 2023-06-03
[pt] False spelling error about "raios X" and "raios-X" (the X-ray wave; X-ray machine)


brunnopleffken:
As a physicist, I deal much about electromagnetic waves. In portuguese, there's a spelling difference between the X-ray used in radiography and tomography, and the X-ray electromagnetic wave.

<img width="453" alt="Screenshot 2023-06-03 at 15 03 57" src="https://github.com/languagetool-org/languagetool/assets/3427344/39b672eb-166d-43ea-bb59-b3f7052b6a0a">

_Above, is the electromagnetic wave, and I spelled it correctly._

To clarify:

- "Raios X": is the X-ray electromagnetic wave, and there's **NOT** a hyphen in it;
- "Raios-X": is the X-ray machine in hospitals and clinics, although is recommended to use the word "radiografia" (radiography).

p-goulart:
@brunnopleffken , determining which one is meant by the user isn't super trivial, but we've removed `raio-x` from the automatic hyphenation rule and added a separate rule that tells users about the different possibilities. At least now you should be able to disable this one particular rule without losing support for the more general hyphenation rule. Hope it's an improvement!

PR #8819, opened 2023-06-13
[pt] Add X_RAY_MACHINE rule

p-goulart:
Silly rule for the weird corner case of #8770 .


languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -38542,6 +38542,42 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
     <category id='COMPOUNDING' name="Palavras Compostas" type="misspelling">
+        <rulegroup id="X_RAY_MACHINE" name="Raio X com hífen" default="temp_off">
+            <rule>
+                <pattern>
+                    <token case_sensitive="yes" regexp="yes">[rR]aios?</token>
+                    <token case_sensitive="no">X</token>
```
susanaboatto, line 38549: `case_sensitive="no"` is not necessary

p-goulart: I know, but I wanted to make it explicit, since in other cases it is.

susanaboatto:
We could revisit this later at some point and see if we can better differentiate between `raio X` and `raio-x` cases. But it's certainly an improvement to the current suggestion 💡

susanaboatto (approved):

PR #8823, opened 2023-06-14
[fr] Style general tone tag applied

GillouLT:
Related to: https://github.com/languagetooler-gmbh/languagetool-premium/issues/5766

Style rules have been reviewed and the new `general` tone tag has been applied where it's compatible (mostly replacing the “clarity” tone tag)

General tone tag count: 90 (including a few rulegroups)

Example of rules that have been tagged as general:

- Rules related to conventions, Kwh, Kelvin, Knot, etc.
- FR repeated words which include things like "C'est pour le <marker>b b</marker> de Caroline."
- Some rules addressing completely incorrect structures and can be found in the `"CAT_ANGLICISMES" name="Anglicismes (calques, emprunts directs, etc.)"` category and `id="CAT_REGIONALISMES"`
- Some rules from `id="CAT_REGLES_DE_BASE`

Additionally, the rule FILER_PARTIR has been deleted due to bad results in Grafana and to the limited improvement window.

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
@@ -843,21 +843,6 @@ USA
             <example correction="de la Covid">Les répercussions économiques <marker>du Covid</marker> ne se sauront pas avant 2022.</example>
             <example correction="à la COVID">Les dommages humains et économiques dus <marker>au COVID</marker> sont considérables.</example>
         </rule>
-        <rule id="FILER_PARTIR" name="filer au lieu de partir" tone_tags="formal">
-            <antipattern>
-                <token inflected="yes" case_sensitive="yes">filer</token>
-                <token regexp="yes">un|du</token>
-                <token>mauvais</token>
-                <token>coton</token>
-            </antipattern>
-            <pattern>
-                <token inflected="yes" case_sensitive="yes">filer</token>
-            </pattern>
-            <message>Le verbe est familier.</message>
-            <suggestion>courir</suggestion>
-            <suggestion>partir</suggestion>
-            <example correction="courir|partir">Il est venu pour finalement <marker>filer</marker> une heure après.</example>
-        </rule>
     </category>
     <category id="CAT_PLEONASMES" name="Pléonasmes" type="style" tone_tags="clarity">
         <!--The following rules were created by Hugo Voisard
@@ -3245,7 +3230,7 @@ USA
             <example correction="défilé de mode"><marker>parade de mode</marker></example>
             <example><marker>défilé de mode</marker></example>
         </rule>
-        <rulegroup id="A_PART_DE_CA" name="à part de ça" tone_tags="clarity">
+        <rulegroup id="A_PART_DE_CA" name="à part de ça" tone_tags="general">
             <rule>
                 <pattern>
                     <token>à</token>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
```diff
@@ -13001,7 +12986,7 @@ USA
             <example correction="affecter"><marker>assigner</marker> à une tâche</example>
             <example correction="affecté">Il a été <marker>assigné</marker> à une tâche.</example>
         </rule>
-        <rule id="AMI_DE_GARCON" name="ami de garçon/gars" tone_tags="clarity">
+        <rule id="AMI_DE_GARCON" name="ami de garçon/gars" tone_tags="general">
```
LucieSteib, line 12989: ahaha wow, cute, this rule has only 11 opens ever :D, AMI_DE_FILLE just below reaches 61, strangely. The rule was apparently created in march, 2022, but the originam file (grammar.xml) is too large to be displayed in github blame mode, sadly. Guess we will never know who created it ^^'

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
```diff
@@ -14362,7 +14347,7 @@ USA
                 <example><marker>leçon particulière</marker></example>
             </rule>
         </rulegroup>
-        <rule id="PORTER_FRUIT" name="porter fruit" tone_tags="clarity">
+        <rule id="PORTER_FRUIT" name="porter fruit" tone_tags="general">
```
LucieSteib, line 14350: believe it or not, this rule has only TPs in diff.
improvement(unblocking): rules CENTRE-JARDIN,TOUR_A_FEU, TOUR_DEAU;POTEAU_DE_BARBIER and maybe ENSEMBLE_DE_PATIO; MEUBLE_DE_PATIO should have the same tag than those? 🥜

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
```diff
@@ -14374,7 +14359,7 @@ USA
             <example correction="porter ses fruits|apporter ses fruits|être fructueux"><marker>porter fruit</marker></example>
             <example><marker>porter ses fruits</marker></example>
         </rule>
-        <rule id="CUIR_PATENT" name="cuir patent(e)" tone_tags="clarity">
+        <rule id="CUIR_PATENT" name="cuir patent(e)" tone_tags="general">
```
LucieSteib, line 14362: that's a weird rule, basically no opens ever

LucieSteib (commented):
praise:legit great work :) 🥜🥜🥜
I just left a few un-prefixed comments, mostly "for fun", and one that could be an actionnable (question:). Feel free to merge in case you think that question and suggestion aren't worth more work.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
GillouLT: It's the second time I go through this rule, it made me smile on both occasions hahaha I guess some L2 users can make that mistake but it's very very rare according to Grafana

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
GillouLT: Added :D

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
GillouLT: I just want to make the mistake in the editor so it can have at least 1 open :(

PR #8848, opened 2023-06-19
[pt] Require upcasing after initial dashes

p-goulart:
- pt-BR introduces dialogue with dash chars, so we need to check if the first char after a sentence-initial dash is uppercase;

 - add hyphen, n-dash, and m-dash to `UppercaseSentenceStartRule` logic *only if* the locale is Portuguese (must add all three because the rule to convert them to m-dashes is (and should be) `picky`);

 - we recognise other locales may not want that behaviour, so that's PT-only, which requires some goofy logic, but I always prefer goofiness to repeating strings, where applicable;

 - add a PT test class for this, and add a couple of tests to EN just to be sure everything is as it should be.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/UppercaseSentenceStartRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/UppercaseSentenceStartRule.java
@@ -217,7 +217,20 @@ public class UppercaseSentenceStartRule extends TextLevelRule {
   }
 
   private boolean isQuoteStart(String word) {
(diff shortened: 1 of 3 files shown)


jaumeortola:
The code is okay. 

Probably, this is not specific to PT-BR, and it is common to most languages. Let's try it with PT. 

I think we already have exceptions for bullet points (not sure if including dashes) in enumeration lists, so that if the sentence doesn't end with a full stop (or question mark, etc.) the capitalized word is not required at the start.

p-goulart:
I didn't want to change anything in the other languages, I didn't know if there were tests somewhere else that might be affected. Anyway, yeah, thanks, let's see how this works for PT first.

PR #8862, opened 2023-06-21
[pt] Add INFORMAL_2SG_BR style rule

p-goulart:


jaumeortola (approved):

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/pt-BR/style.xml
```diff
@@ -952,6 +955,154 @@ USA
...
+                        <exception regexp="yes" inflected="yes">&dative_person_verbs;</exception>
+                    </token>
+                    <marker>
+                        <token regexp="yes" spacebefore="no">&hifen;</token>
+                        <token spacebefore="no">te</token>
+                    </marker>
+                </pattern>
+                <message>No português brasileiro, o uso da segunda pessoa do singular não é consistente em todos os dialetos. Seu texto soará mais neutro se substituir por &quot;você&quot;.</message>
+                <suggestion> você</suggestion>
+                <example correction=" você">Conheço<marker>-te</marker> desde 2002.</example>
```
susanaboatto, line 1077: I think "lhe" would be a better correction for this. It's not grammatically ideal, but neither is the original sentence for this example. Check diff results [here](https://regression.languagetoolplus.com/via-http/2023-06-30/pt-BR/result_style_INFORMAL_2SG_BR%5B10%5D.html)

"Conheço você" as a replacement for "te conheço", on the other hand, works better (and I see there is a subrule for this already)

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/pt-BR/style.xml
```diff
@@ -952,6 +955,154 @@ USA
...
+            <rule> <!-- reflexive proclitic -->
+                <pattern>
+                    <token>te</token>
+                    <token postag_regexp="yes" postag="V...2S."/>
+                </pattern>
+                <message>No português brasileiro, o uso da segunda pessoa do singular não é consistente em todos os dialetos. Seu texto soará mais neutro se substituir por &quot;te&quot;.</message>
+                <suggestion>se <match no="2" postag_regexp="yes" postag="(V...)2S(.)" postag_replace="$13S$2"/></suggestion>
+                <example correction="se esqueça">Não <marker>te esqueças</marker>!</example>
+                <example correction="se esqueceu">Acho que <marker>te esqueceste</marker>...</example>
+            </rule>
```
susanaboatto, line 1038: As per the [diff](https://regression.languagetoolplus.com/via-http/2023-06-30/pt-BR/result_style_INFORMAL_2SG_BR%5B7%5D.html), all conjugation correction rules (`AUX_VERBO`, `CONDITIONAL_NO_INF`, `GENERAL_VERB_AGREEMENT_ERRORS`, `AUXILIARY_VERB_INFINITIVE`) should be prioritized over this one. (I'd argue perhaps all grammar rules should be prioritized over style ones 🤔)

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/pt-BR/style.xml
p-goulart: idgi... `lhe` there is not 'not ideal', it's prescriptively incorrect. My original idea was to have `-o`/`-a` there, since that'd be the 3SG equivalent, but there are issues with some verb endings (`-n[ao]`, `-l[ao]`).

afaics the bad diff results come from ditransitive verbs, so perhaps it might be enough to expand that list (I created an entity for that). Or even just having a separate rule for ditransitive verbs? Either way, I don't think it's a good idea to introduce a proscribed structure.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/pt-BR/style.xml
p-goulart: That's very true, rule ordering in this kind of case is crucial.

PR #8905, opened 2023-07-07
[fr] Rule linking OS CW22

GillouLT:
To address: https://github.com/languagetooler-gmbh/languagetool-premium/issues/5043

4 rules have been linked to blog posts, one of the rules already had an blog article url, but the new blog article seems to fit better.

Typo in rule ID fixed

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
@@ -23266,7 +23266,7 @@ USA
                 <example correction="afin de|pour|de façon à|ainsi">Il le dit <marker>de manière à</marker> réussir son projet.</example>
             </rule>
         </rulegroup>
-        <rulegroup id="REP_CONCEQUENCE" name="conséquence (style)" min_prev_matches="1">
+        <rulegroup id="REP_CONSEQUENCE" name="conséquence (style)" min_prev_matches="1">
             <antipattern>
                 <token regexp="yes">voies?</token>
                 <token>de</token>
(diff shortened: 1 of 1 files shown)


GillouLT:
Not sure about fixing the typo @LucieSteib do you think I should just leave it as is?

LucieSteib:
Yeah, we might as well, since changing the name would make the rule "reappear" to people that disabled it (aaand there's quite a few...)
image
Maybe that would be worth it if the status of the rule was chnaged too (tone_tag? Adding picky if not present?)

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
```diff
@@ -23265,7 +23266,7 @@ USA
                 <example correction="afin de|pour|de façon à|ainsi">Il le dit <marker>de manière à</marker> réussir son projet.</example>
             </rule>
         </rulegroup>
-        <rulegroup id="REP_CONCEQUENCE" name="conséquence (style)" min_prev_matches="1">
+        <rulegroup id="REP_CONSEQUENCE" name="conséquence (style)" min_prev_matches="1">
```
LucieSteib, line 23269: Nitpick: maybe that shouldn't be changed (or, if we change the name, we should also maybe change the rule itself). The rule is often disabled, and changing the name will make it pop up again for the users...
image

GillouLT: Ok, so I've reverted the typo correction and the rule is now picky, we will focus on picky, we will see what to do with this kind of repetition rules wrt picky tag later this quarter I think

PR #8914, opened 2023-07-10
[pt] Extract entities into .ent files

p-goulart:
- less repetition;
 - single point of failure;
 - cleaner and shorter grammar/style XMLs;
 - already replaced some common repeating messages with entities.


maphjo:
@p-goulart I like the idea!! :peanuts: :peanuts: :peanuts: If it works properly, please share with the rest of LINGU in the daily and maybe more in-depth on Friday. Might also be worth adding the .ent files to https://languagetooler.atlassian.net/wiki/spaces/DEV/pages/2163933198/Overview+JAVA+TXT+files+etc.+How+they+work+conventions+WIP

p-goulart:
@maphjo: @jaumeortola is checking whether we can change the relative paths to the `.ent` files so that the same entities are available on both OS **and** premium repos. Once we've got that nailed down we'll share the solution with the team.

susanaboatto (approved):

jaumeortola:
I didn't find a better solution for premium. We could keep a (manual) synchronized copy of the files with entities in premium. 

There is one drawback. The paths in this PR are right for compiling, but they will give errors in an XML editor. I guess we can live with that. The relative paths are the ones expected by an editor, but they don't work when compiling.

jaumeortola:
There is [XInclude](https://en.wikipedia.org/wiki/XInclude). Have you tried it? @p-goulart 
It doesn't look easy to make it work, and it will probably have the same problems with paths.

jaumeortola (approved):

jaumeortola:
There is a thing called EntityResolver in the Java XML parsers: https://stackoverflow.com/questions/30352951/edit-the-link-of-a-xml-entity-with-java
We could use relative paths in the XML files, and then adjust these paths with the EntityResolver. That will need some more research.

p-goulart:
> There is one drawback. The paths in this PR are right for compiling, but they will give errors in an XML editor. I guess we can live with that.
Yes, that's annoying, but not a dealbreaker, at least not in my case.

> There is [XInclude](https://en.wikipedia.org/wiki/XInclude). Have you tried it?
I did try. We'd probably need to modify the rule XML parser to allow for `xi:include` – when I attempt to use it, the XML verification test told me something about unexpected elements. Either way, I thought working with entities would be closer to what we have now, so it'd be less of a departure from what people are already used to.

> There is a thing called EntityResolver in the Java XML parsers
I'll try to look into it, sounds promising.

PR #8944, opened 2023-07-13
Fix entity file relative path mess

p-goulart:
**Problem:** relative paths to entity files from within rule XMLs were not being resolved correctly.

**Reason:** paths are resolved from the root of the project being built, so the relative paths were not pointing to files inside the resources directories next to the rules XML.

**Solution:** introduce [EntityResolvers](https://docs.oracle.com/javase/8/docs/api/org/xml/sax/EntityResolver.html), which intercept the IDs of the external entities inside the XML files and resolve their paths according to arbitrary rules. (In our case, just resolve the path relative to the source of the XML.) In order to do that, we had to introduce a couple of new classes (one for SAX and one for DOM) and make sure that they were instantiated with the correct params (namely, the path to the source XML with rules).

I've tested it locally – not only have run all core and PT tests (in addition to some other languages for good measure), I have also compiled it and double-checked that requests to the server are well handled. It seems to work! 🤞 

Huge thanks @jaumeortola and @danielnaber for helping out with this! 🙏 🙇 

**EDIT:** spoke to @jaumeortola, two new things:
- [x] change the actual location of the entity files to `resource`, not `rules`;
- [x] extend this logic to the disambiguation grammars.


p-goulart:
@jaumeortola if we want to keep `.ent` files in the `resource` dir, we have **two** options:

1. keep the relative path, e.g.:
```xml
<!DOCTYPE rules [
        <!ENTITY % messages  SYSTEM "file:../../resource/pt/entities/messages.ent" >
        %messages;
]>
```
This way we keep the logic in the code exactly as it is, but we require longer relative paths in our XMLs.

2. **assume** (as standard practice, that we'll need to document somewhere) that all `.ent` files **must** be within `resources`, e.g.:
```xml
<!DOCTYPE rules [
        <!ENTITY % messages  SYSTEM "file:entities/messages.ent" >
        %messages;
]>
```
This would require changes to how we resolve paths in the core code.

p-goulart:
@jaumeortola files migrated to `resource` and disambiguator grammars now also in scope, lmk if we need anything else here.

@susanaboatto I suppose I just need you to check the `pt*` files to see if anything too crazy has happened. I already had to disamble an AP in `pt-PT/grammar.xml` because the disambiguator entities had diverged from the ones in `rules`.

PR #8947, opened 2023-07-14
[pt] Moved rule from wordiness-pt.txt to style.xml (verb "sendo")

marcoagpinto:
@p-goulart @susanaboatto 

Ahhhh… it committed the two files I changed at the same time… cool.

I thought I had to manually do file by file.

So, here is the thing: The “considerado como” worked with all sentences I could think of, but in my thesis I had the verb "ser” (“sendo”) and it gave a false positive.

So, I turned it into a rule.

I tried the inflected=yes to verb “ser” but it only had two hits, and they were “sendo”, so I changed the rule I created to “sendo” (min=0, max=1).

Here is the before and the after:
[0before.txt](https://github.com/languagetool-org/languagetool/files/12046256/0before.txt)

[1.txt](https://github.com/languagetool-org/languagetool/files/12046258/1.txt)

As you will notice, it is 5am, this is usually the time I work on rules.

I do things at strange time, at all hours of the day and night.

😄 😋 😛


languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
```diff
@@ -1424,6 +1424,21 @@ USA
             <!-- Created by Tiago F. Santos, Portuguese rule, 2017-06-11 -->
             <url>https://pt.wikipedia.org/wiki/V%C3%ADcio_de_linguagem#Prolixidade_ou_preciosismo</url>
             <short>Expressão prolixa</short>
+
+            <rule>
```
p-goulart, line 1428:
```suggestion
            <rule default="temp_off">
```

p-goulart (changes requested):

p-goulart (approved):

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
marcoagpinto: Ahhhh… right! Sorry! I have committed it again!

p-goulart: FYI when I add a suggestion, you don't have to commit it yourself, you can just click the 'Commit suggestion' button on GitHub ;)

Issue #8948, opened 2023-07-14
[pt] Idea for rule: “dá” → “confere” - 2023-07-14


marcoagpinto:
Heya, @ricardojosehlima

Yes, it is 6am, and I came up with an idea for a scientific/academic/formal rule:
`E isto tudo DÁ aleatoriedade ao jogo.`
to:
`E isto tudo CONFERE aleatoriedade ao jogo.`

I was meditating how to detect the right noun after the “dá” and I came up with words ending in:
“de” → “velocidade”, “integridade”, etc.
“ez” →”rapidez”

And after it, it should work with:
`a, as, à, às, ao, aos`

Is it a good idea?

Academic? Scientific? Formal?

Thanks!

ricardojosehlima:
Now that seems a challenge, but it might work!

marcoagpinto:
Ricardo!

❤️ 😄 😋

marcoagpinto:
Ricardo, I will work on it at 5am.

But I am very happy to see it is a good idea and that it may work.

🤗 🤗 🤗 🤗 🤗

marcoagpinto:
@ricardojosehlima 

Hello!

I couldn't wait for 5am, I was so eager.

It took me 10 minutes to write the rule.

```
Portuguese (Portugal): 52 total matches
Portuguese (Portugal): 599999 total sentences considered
Portuguese (Portugal): ø0.00 rule matches per sentence
```

[0.txt](https://github.com/languagetool-org/languagetool/files/12054067/0.txt)

Is this it? 20 lines of code? Do you have any suggestions to improve it, or is it good?

Thanks!

```
        <rule id='CIENTÍFICO_DAR_CONFERIR' name="[Universitário][Científico] dar → conferir" > <!-- type="style" tags="picky" tone_tags="academic"-->
            <!--
            It uses filters:
              -de → velocidade, integridade, etc.
              -ez → rapidez, etc.
            -->
            <pattern>
                <marker>
                    <token inflected='yes'>dar</token>
                </marker>
                <and>
                    <token regexp='yes'>.*de|.*ez</token>
                    <token postag='NC.+|AQ.+' postag_regexp='yes'/>
                </and>
                <token regexp='yes'>[aà]s?|aos?</token>
            </pattern>
            <message>Num contexto formal/científico, é preferível escrever 'conferir'.</message>
            <suggestion><match no='1' postag='V.+' postag_regexp='yes'>conferir</match></suggestion>
[...]

PR #8954, opened 2023-07-15
[pt] Added rule ID:CIENTÍFICO_DAR_CONFERIR (thanks to Ricardo Joseh Lima)

marcoagpinto:
Heya, Pedro and Susana,

Here is a brand-new academic/scientific rule developed thanks to Ricardo Joseh Lima.

https://github.com/languagetool-org/languagetool/issues/8948

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -10188,6 +10188,30 @@ USA
         </rulegroup>
 
 
+        <rule id='CIENTÍFICO_DAR_CONFERIR' name="[Universitário][Científico] dar → conferir" type="style" tags="picky" tone_tags="academic" default='temp_off'>
+            <!--
+            It uses filters:
+              -de → velocidade, integridade, etc.
+              -ez → rapidez, etc.
+            -->
+            <pattern>
+                <marker>
+                    <token inflected='yes'>dar</token>
+                </marker>
+                <and>
+                    <token regexp='yes'>.*de|.*ez
+                        <exception regexp='yes'>continuidade|oportunidade</exception>
+                    </token>
+                    <token postag='NC.+|AQ.+' postag_regexp='yes'/>
+                </and>
+                <token regexp='yes'>[aà]s?|aos?</token>
+            </pattern>
+            <message>Num contexto formal/científico, é preferível escrever 'conferir'.</message>
+            <suggestion><match no='1' postag='V.+' postag_regexp='yes'>conferir</match></suggestion>
+            <example correction="confere">Esta equação <marker>dá</marker> aleatoriedade ao jogo.</example>
+        </rule>
+
+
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
```diff
@@ -10188,6 +10188,30 @@ USA
         </rulegroup>
 
 
+        <rule id='CIENTÍFICO_DAR_CONFERIR' name="[Universitário][Científico] dar → conferir" type="style" tags="picky" tone_tags="academic" default='temp_off'>
```
p-goulart, line 10191: IMO this rule is solid enough not to be `picky`. Maybe it can be `is_goal_specific`. @susanaboatto what do you think?

p-goulart (approved):

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
```diff
@@ -10188,6 +10188,30 @@ USA
...
+            It uses filters:
+              -de → velocidade, integridade, etc.
+              -ez → rapidez, etc.
+            -->
+            <pattern>
+                <marker>
+                    <token inflected='yes'>dar</token>
+                </marker>
+                <and>
+                    <token regexp='yes'>.*de|.*ez
```
p-goulart, line 10202:
```suggestion
                    <token regexp='yes'>.+de|.+ez
```

p-goulart: Preferring `.+de` (e.g. `dar de comer a` -> *`conferir de comer a`.

p-goulart: Also, is there any particular reason for these two suffixes, and not others?
* -mento;
* -idão;
* -itude.

marcoagpinto: Yes, because I was stressed and couldn't think of all the hypotheses.

Feel free to make any changes necessary to have the best rule possible.

❤️ 😋 😛

p-goulart: Let's leave it like this for now and see how it performs, then. Later on we can add some suffix entities to the XMLs and experiment. FWIW this is a really good rule.

marcoagpinto:
@p-goulart 

Peter, the truth is that I break under pressure, which was one of the reasons I retired with disability.

😋 

The things I am telling in public.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
susanaboatto: Yes, I think we can remove `picky` and see how the rule performs

PR #8955, opened 2023-07-15
Fix typos found by codespell in languagetool-core

DimitriPapadopoulos:
I hope modifying user-visible strings directly in `languagetool-core/src/main/resources/org/languagetool/*.properties` is OK.


danielnaber:
Thanks! 

> I hope modifying user-visible strings directly in languagetool-core/src/main/resources/org/languagetool/*.properties is OK.

For English it's actually okay.

PR #8960, opened 2023-07-15
Improve DOPPELTES_VERB again

Discostu36:
I split up the auxiliary verb rule and also fixed the other false positives. 

Rules 2-4 are now not "temp_off" anymore, they seem solid.

 A few months ago, I just wanted to write a simple rule, now it's 250 lines 😂

@danielnaber @Luisa-LT


Luisa-LT:
Hi Michael, yes that's how it often happens to us too, first you think it's a quick fix, and then you're hit again how complicated the German language is actually is :D but thanks for that! 
As far as I could see now, the rule group looks good and super that you have set `temp_off` the individual rules now or some already activated some. I have merged the PR. Tomorrow we will see new results in the [diff](https://internal1.languagetool.org/regression-tests/via-http/2023-07-18/) (link available from tomorrow)  :-)

Discostu36:
Only a small to-do list remaining:

**All rules**
* Exception if `;` `!`, `?`, or `…` is between the two verbs

**Rule 5**
* Exception for stylistic uses like `denn ein Shirt ist ein Shirt ist ein Shirt `
* Fix overshooting antipattern that allows `Das System ist inzwischen 70 Jahre alt ist.`

Discostu36:
Shouldn't the rules without `temp_off` already work on the website? Or isn't it updated nightly anymore?

danielnaber:
> Shouldn't the rules without temp_off already work on the website?

The rules for the website are updated daily, unless you're a premium user.

Luisa-LT:
@Discostu36 I just checked the example sentences from rule 2, and it's working fine in the editor: 

image

> Wenn wir kaufen Wurst kaufen, müssen wir sie auch bezahlen.
> Wie konnte er wissen, dass Ursula isst Spaghetti am liebsten mit Tomatensoße isst? Er war sich sicher, dass er wird Feuerwehrmann wird.
> Er stellte fest, dass er hat gute Laune hat.

If you're a premium user you should probably see the changes tomorrow (until then you could just log out to check)

Discostu36:
Ah, so the premium users are on a more stable release. Didn't think about that, thanks.

PR #8989, opened 2023-07-19
[all] Change wording for test message if default="…" is set on rulegroup and rule level

maphjo:
The message is made more explicit about what is causing the error and a suggestion on how to eliminate the cause of the error has been added in bold.


maphjo:
@danielnaber This concerns the message of the test triggered today when @GillouLT was moving the fr rules back.
I think the new wording reflects better what is happening.
I am unsure about the `\033[1m` to make part of the message appear in bold. It worked locally on my machine and did not create any errors. But this is why I wanted you to have a quick look at it.

PR #8991, opened 2023-07-19
[fr] new French variants: CH, BE, CA

jaumeortola:
The new variants (fr-CH, fr-BE, fr-CA) extend the default variant (fr, fr-FR) with disabled and enabled rules by default.


languagetool-language-modules/fr/src/main/java/org/languagetool/language/BelgianFrench.java
```diff
@@ -0,0 +1,24 @@
+package org.languagetool.language;
```
danielnaber, line 1: Could you also add the standard copyright header here?

PR #8994, opened 2023-07-20
[pt] Removed "temp_off" from rules, they seem solid enough in the results

marcoagpinto:
Heya,

I was looking at the results:
https://internal1.languagetool.org/regression-tests/via-http

And they seem solid enough to remove the temp_off.

I also removed the picky from the academic rule.

There may a false positive or two, but they can be easily removed in future builds.

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -1224,7 +1224,7 @@ USA
             <url>https://pt.wikipedia.org/wiki/V%C3%ADcio_de_linguagem#Prolixidade_ou_preciosismo</url>
             <short>Expressão prolixa</short>
 
-            <rule default='temp_off'>
+            <rule>
                 <pattern>
                     <marker>
                         <token regexp='yes'>considerad[ao]s?</token>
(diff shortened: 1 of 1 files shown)


p-goulart:
I'd personally add some more exceptions to `CIENTÍFICO_DAR_CONFERIR` before enabling for all users. Still a few collocations that don't quite work:
- dar visibilidade -> *conferir visibilidade;
- dar prioridade -> *conferir prioridade;
- dar saúde -> *conferir saúde (not wrong per se, but I don't think 'saúde' should be here).

marcoagpinto:
> I'd personally add some more exceptions to `CIENTÍFICO_DAR_CONFERIR` before enabling for all users. Still a few collocations that don't quite work:
> 
>     * dar visibilidade -> *conferir visibilidade;
> 
>     * dar prioridade -> *conferir prioridade;
> 
>     * dar saúde -> *conferir saúde (not wrong per se, but I don't think 'saúde' should be here).



Done!

I have just committed the exceptions you mentioned.

p-goulart (approved):

marcoagpinto:
@p-goulart @susanaboatto @maphjo 

Heya,

Remember when I was terribly scared of moving from TortoiseSVN to TortoiseGit?

I was wrong, it is very easy to use.

I made a drama out of a simple thing.

It was just a matter of using it for a few days.

😄 😋 😛 🤗 ❤️

Issue #9007, opened 2023-07-20
[pt?] TESTRULES PT doesn't work with nightly build - 2023-07-20


marcoagpinto:
Heya,

I am not sure if anyone else has noticed it, but the new entities' implementation makes the nightly snapshot where we test the rules “`TESTRULES PT`” unusable.

```
C:\Users\BLAH BLAH\LanguageTool-20230720-snapshot\LanguageTool-6.3-SNAPSHOT>testrules pt
Running XML pattern tests...
LanguageTool version 6.3-SNAPSHOT (2023-07-20 16:45:17 +0000, a3e185c)
Known languages: [Arabic, Asturian, Belarusian, Breton, Catalan, Catalan (Valencian), Catalan (Balearic), Danish, German, German (Germany), German (Austria), German (Swiss), Simple German, Greek, English, English (US), English (GB), English (Australian), English (Canadian), English (New Zealand), English (South African), Esperanto, Spanish, Spanish (voseo), Persian, French, Irish, Galician, Italian, Japanese, Khmer, Dutch, Dutch (Belgium), Polish, Portuguese, Portuguese (Portugal), Portuguese (Brazil), Portuguese (Angola preAO), Portuguese (Moçambique preAO), Romanian, Russian, Slovak, Slovenian, Swedish, Tamil, Tagalog, Ukrainian, Chinese, Testlanguage]
Running XML validation for pt/grammar.xml...
Exception in thread "main" java.io.IOException: Cannot load or parse '/org/languagetool/rules/pt/grammar.xml'
        at org.languagetool.XMLValidator.validateWithXmlSchema(XMLValidator.java:117)
        at org.languagetool.rules.patterns.PatternRuleTest.validatePatternFile(PatternRuleTest.java:242)
        at org.languagetool.rules.patterns.PatternRuleTest.validatePatternFile(PatternRuleTest.java:218)
        at org.languagetool.rules.patterns.PatternRuleTest.runTestForLanguage(PatternRuleTest.java:189)
        at org.languagetool.rules.patterns.PatternRuleTest.runGrammarRulesFromXmlTestIgnoringLanguages(PatternRuleTest.java:184)
        at org.languagetool.rules.patterns.PatternRuleTest.main(PatternRuleTest.java:927)
[...]

PR #9011, opened 2023-07-21
[pt] Fix entity resolution on Windows (bug #9007)

p-goulart:
- there might be an issue (#9007) when resolving external entity paths when running local source code tests on Windows machines, because of how Windows' Java treats the format of Java URLs;

 - added a workaround from https://stackoverflow.com/questions/43972777/java-nio-file-invalidpathexception-illegal-char-at-index-2;

 - also improved the exception messages a little bit.

⚠️ **Pending tests on Windows** ⚠️ 
1. fetch latest for `languagetool` and checkout to branch `fix/entities/windows_paths`;
2. build with `mvn clean install -DskipTests`;
3. go to `./languagetool-standalone/target/LanguageTool-6.3-SNAPSHOT/LanguageTool-6.3-SNAPSHOT`;
4. run the `testrules` script – in this case, `testrules.bat`.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/XMLValidator.java
+++ b/languagetool-core/src/main/java/org/languagetool/XMLValidator.java
@@ -35,6 +35,7 @@ import javax.xml.transform.stream.StreamSource;
 import javax.xml.validation.*;
 import java.io.*;
 import java.net.MalformedURLException;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


p-goulart:
Hm, spent all day trying to get someone to help me test this, which was much more challenging than I thought. It's not a huge change and – at the very least – it doesn't break the Unix tests. So it should be safe enough... If someone wants to review (and merge) at some point before Monday, that's great, if not, then I'll go back to debugging this with someone else on Monday. 🖖

p-goulart:
Probably stale; cf. #9023

PR #9012, opened 2023-07-21
[sv] Changing character encoding of fsa dict files from ISO-8859-1 to UTF-8

ljo:
- Changing character encoding of fsa dict files from ISO-8859-1 to UTF-8. 
- Adding more tests for synthesizer, tagger and disambiguation.


ljo:
The failing checks are not due to the PR code but rather due to blocked dependencies for language-jp at circleci: `[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.6.0:go-offline (default-cli) on project language-ja: org.eclipse.aether.resolution.DependencyResolutionException: Failed to collect dependencies at com.github.lucene-gosen:lucene-gosen:jar:ipadic:6.2.1 -> org.apache.solr:solr-core:jar:6.2.1 -> org.restlet.jee:org.restlet:jar:2.3.0:...`

languagetool-language-modules/sv/src/main/java/org/languagetool/tokenizers/sv/SwedishWordTokenizer.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * Tokenizes just like WordTokenizer with the exception for words such as
+   * "oma's" that contain an apostrophe in their middle.
+   * @param text Text to tokenize
+   * @return List of tokens
+   */
+  @Override
+  public List<String> tokenize(final String text) {
```
danielnaber, line 51: This is basically a copy of the Dutch code, isn't it? We should avoid code duplication. There should be a common super class, with both the Dutch and the Swedish code extend.

danielnaber (changes requested):

languagetool-language-modules/sv/src/main/java/org/languagetool/tokenizers/sv/SwedishWordTokenizer.java
ljo: This extension of WordTokenizer is mainly based on the Dutch, yes. So what is the preferred location and naming conventions for the extension strategies you mention? I just followed the strategies I could see from earlier implementations. But It does differ from the Dutch here as well Also from the Polish, French and German. But neither of the ones having extended the base class WordTokenizer, differ very much.

danielnaber: The French, German and Polish word tokenizer don't seem to duplicate code from the Dutch tokenizer, the code looks very different. To avoid duplication, you could have e.g. a `GenericWordTokenizer` with common code in `languagtool-core` and have the Dutch and Swedish one extend it.

ljo: I removed it and its tests for now. I don't have the needed overview of all the 25 `WordTokenizer` implementations to make a generic one yet, even though I reviewed quite a lot of them. I saw only a few strands, though.

PR #9023, opened 2023-07-23
entities: adjust paths

jaumeortola:
* use relative paths without protocol (`file:`) for entities files. 

It is an open discussion whether using "protocol + relative paths" is a standard practice. It is not clear. In some editors, it doesn't work. It is better to avoid it. 
Paradoxically, the current implementation in LT (with "protocol + relative paths") works because the detection of the protocol doesn't work in one of the internal steps. 

Comment: With this PR, the path inside the disambiguation.xml file has to be: `<!ENTITY % entities SYSTEM "../../resource/es/entities.ent" >` (including always `/resource/es/`). It is a bit strange, but we can live with it. It could be fully computed in Java if we wanted.


marcoagpinto:
@jaumeortola 

Does this mean that TESTRULES PT will work tomorrow, Monday, in the nightly?

jaumeortola:
> Does this mean that TESTRULES PT will work tomorrow, Monday, in the nightly?

Nothing should break at any moment.

p-goulart:
Pushed the change in the filepath also for `pt-BR` files.

p-goulart:
@jaumeortola I assume this means #9011 is stale.

Does it address #9007 or will that still need to be done separately?

languagetool-core/src/main/java/org/languagetool/RuleEntityResolver.java
```diff
@@ -26,22 +23,19 @@ public class RuleEntityResolver implements EntityResolver {
...
   }
 
-  /**
-   * @param xmlUrl path to the source XML where the external entity relative path is found.
-   */
-  public RuleEntityResolver(URL xmlUrl) {
-    this.xmlUrl = xmlUrl;
+  public String getPathFromLTResourceFolder(String input) {
+    // we assume that the entities file is in the resource folder of each language
+    return input.replaceAll(".*/resource/", "").replaceAll("\\.\\./", "");
```
p-goulart, line 34: What if a grammar file is inside a further directory (e.g. `pt/pt-BR`). There, the relative path will be written `../../../`. Will the `.replaceAll()`s here be enough?

languagetool-core/src/main/java/org/languagetool/RuleEntityResolver.java
```diff
@@ -26,22 +23,19 @@ public class RuleEntityResolver implements EntityResolver {
...
     return null;
   }
 
-  /**
-   * @param xmlUrl path to the source XML where the external entity relative path is found.
-   */
-  public RuleEntityResolver(URL xmlUrl) {
-    this.xmlUrl = xmlUrl;
+  public String getPathFromLTResourceFolder(String input) {
+    // we assume that the entities file is in the resource folder of each language
```
p-goulart, line 33: This is certainly an assumption, but what we're doing here isn't really resolving any paths relative to anything else, we're just hardcoding their location, and keeping the relative paths there only so XML editors can resolve them.

This is *fine*, but we're moving from a solution that actually resolved the paths as *relative* paths to one that *assumes* stuff about their location, so we should probably document this well.

languagetool-core/src/main/java/org/languagetool/RuleEntityResolver.java
jaumeortola: It is enough. In most cases, `.replaceAll(".*/resource/", "")` will be enough. In fact, it is a requirement to have this string in the relative path.

jaumeortola:
I am not usually on Windows, but I will make a test of this branch on Windows.

p-goulart:
Your force push overwrote changes I had made to some `pt` files, namely `pt-BR/*.xml` and `pt/style.xml`. Can you re-apply?

jaumeortola:
> Your force push overwrote changes I had made to some `pt` files, namely `pt-BR/*.xml` and `pt/style.xml`. Can you re-apply?

Yes, sorry. I had to resolve a merge conflict, and my skills are limited. 

The tests passed without those changes. Does it mean that we are not validating those files?

p-goulart:
It might mean that, which would be bad :(

PR #9024, opened 2023-07-24
[pt] Added words to wordiness-pt.txt

marcoagpinto:
Heya,

Some words related to my thesis reviewal.


p-goulart:
For consistency, wouldn't it better to have a proper rule that replaces `fazer uso de` with `usar`?

marcoagpinto:
> For consistency, wouldn't it better to have a proper rule that replaces `fazer uso de` with `usar`?

Sure, I will create the rule tonight.

@p-goulart 
Could you tell me the rule tone so that I don't mess up?

Thanks!

😋

Issue #9000, opened 2023-07-20
[pt] Idea for rule: “dar valor” → “valorizar” - 2023-07-20


marcoagpinto:
@ricardojosehlima

Heya, again,

I was exercising and this popped-up on my mind:
`“Ele não dá valor à mulher.”`
to:
`“Ele não valoriza a mulher.“`

Verb “dar” + "valor” + “a(s)/o(s)/à(s)/ao(s)”

Is it a good idea?

Thanks!

ricardojosehlima:
Sounds good!

marcoagpinto:
Ahhhhhhhhhhh…

Dear brother, Ricardo,

Thank you!

A mission for tonight!

😄 😋 ❤️

p-goulart:
IMO it doesn't fit any of the [tone tags](https://dev.languagetool.org/style_tone_tags#description-of-tonetags) other than the (as yet unimplemented) `shorten_it`. It doesn't make it particularly clearer, more professional-sounding, formal, more persuasive, etc.

I don't think it's a particularly useful rule, but I suppose if you want to add it, it should be set to `picky` until we have support for `shorten_it`. @susanaboatto ?

marcoagpinto:
@ricardojosehlima 

We are stuck on this: https://github.com/languagetool-org/languagetool/issues/9007

susanaboatto:
I agree @p-goulart. If you add the rule, @marcoagpinto, make sure to add `<!--IDEA shorten_it-->` to it so we can tag it once `shorten_it` is active.

marcoagpinto:
@susanaboatto 

I have created the rule but have been unable to test it because TESTRULES PT doesn't work.

I will wait until the command is working so that I can test it.

```
        <rule id='DAR_VALOR_VALORIZAR' name="Simplificar: Verbo 'dar' + valor → valorizar">
            <!-- IDEA shorten_it -->
            <pattern>
                <marker>
                    <token regexp="yes" inflected='yes'>dar</token>
                    <token>valor</token>
                    <token regexp='yes'>às?|aos?</token>
                </marker>
            </pattern>
            <message>&simplify_msg;</message>
            <suggestion><match no='1' postag='V.+' postag_regexp='yes'>valorizar</match> <match no='3' postag='SPS00:DA0(..)0' postag_replace='DA0$10'>o</match></suggestion>
            <example correction="valoriza">Ele não <marker>dá valor à</marker> mulher.</example>
        </rule>
```

[...]

PR #9031, opened 2023-07-24
[pt] Added rule ID:DAR_VALOR_VALORIZAR

marcoagpinto:
Heya,

This is the rule I couldn't create last week because TESTRULES PT wasn't working.

Depends on #9030


marcoagpinto:
https://github.com/languagetool-org/languagetool/issues/9000

Only 5 hits in 600 000 sentences, so I didn't use “picky” since it is a very low-impact rule.

p-goulart:
Just a minor comment: it looks like you didn't branch off `master` here, but off `lt_marcoagpinto_20230724_2329`, so we see changes here also from that branch.

p-goulart (approved):

susanaboatto (approved):

PR #9042, opened 2023-07-25
[sv] Adding initial ngram support for Swedish

ljo:
- Adding initial ngram support for Swedish. 
- Adding some initial infra and some rules with tests.


danielnaber:
The build failure is a problem with the build system. It should disappear if you sync your changes with the latest head.

languagetool-language-modules/sv/src/main/java/org/languagetool/rules/sv/UpperCaseNgramRule.java
```diff
@@ -0,0 +1,103 @@
...
+    return "SV_UPPER_CASE_NGRAM";
+  }
+
+  @Override
+  public String getDescription() {
+    return "Granskar ord som ofta skrivs med liten eller stor begynnelsebokstav när det ska vara tvärsom";
+  }
+
+  @Override
+  public RuleMatch[] match(AnalyzedSentence sentence) {
```
danielnaber, line 68: This is a copy of `UpperCaseNgramRule` for German, isn't it? We should avoid that kind of code duplication.

danielnaber (changes requested):

languagetool-language-modules/sv/src/main/java/org/languagetool/rules/sv/UpperCaseNgramRule.java
ljo: Yes, I have looked at what would be the best foundation for each of the features and started from there. But I have definitely not just copied it. It is all carefully selected to make initial testing possible for each feature. But if you have a suggestion on how to add the ngram feature without duplicating some extending methods which happens to be extended in a few other languages as well?

danielnaber: These methods seem to be identical except the `String msg` part, aren't they? Then all the identical parts should be in a super class in `languagetool-core` and only the messages should be defined in the sub classes.

ljo: Yes, as an initial dummy for creating the tests. Ok, I will start with removing it and then add it according to your suggestion. There are way fewer implementations, three only, to keep track of here. :)

PR #9052, opened 2023-07-27
[fr] Adding antipattern on FAIRE_VPPA

LucieSteib:
To fix https://github.com/languagetooler-gmbh/languagetool-premium/issues/6008

Adding an antipattern to avoid past participle to be consider wrong after verb _souhaiter._
The parser considers "chargé" as an adjective in this context, that's why no AP was triggered.
image


GillouLT:
That's great, we should keep improving those rules :)

PR #9066, opened 2023-08-01
[fr] A_A_ACCENT[1] adding AP

LucieSteib:
To fix https://github.com/languagetooler-gmbh/languagetool-premium/issues/6065

Adding antipattern for A-tu structure (technically false, but should and will be handled by ACCORD_V_QUESTION[5], not A_A_ACCENT[1] that suggest a wrong correction.


GillouLT:
Nice workaround! LGTM!

PR #9085, opened 2023-08-04
Revert "[nl] postags"

p-goulart:
This reverts commit 5e75ebf6b8a74d8b7a45c26fa22f45ec7f10583e.

@danielnaber this commit breaks the premium build; in @mark-baas 's absence, I suppose the only thing we can do is revert it.


danielnaber:
I've already commented out the Premium rule that's affected and notified Mark, everything should be okay again for now.

PR #9086, opened 2023-08-04
[pt] Disabled sentences in wordiness-pt.txt (too many FPs, need to convert to rule)

marcoagpinto:
@p-goulart @susanaboatto 

Hello!

I was looking at the results, and it produces too many FPs, so I have disabled the two lines.

During the weekend, I will create a rule for it.

At least we have an idea about the focus of the rule.


p-goulart (approved):
Fair enough. I did think this seemed a little too radical, but didn't see the harm in giving it a go for a few days. Thanks for checking!

PR #9087, opened 2023-08-05
[pt] Added rule ID:SER_NECESSARIO_NECESSARIO_V2

marcoagpinto:
Heya, Susana and Pedro,

Like I wrote before, the wordiness-pt.txt rules were giving too many FPs in the two lines I added there.

So, I have converted the lines into a XML rule.

Here are the results:
[_1.txt](https://github.com/languagetool-org/languagetool/files/12266696/_1.txt)

❤️ ❤️ ❤️ 

Thanks!

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -11944,6 +11944,27 @@ USA
     <category id='SHORTEN_IT' name="Linguagem concisa" type="style">
 
 
+        <rule id='SER_NECESSARIO_NECESSARIO_V2' name="Simplificar: Ser necessário → necessário" default="temp_off">
+            <!-- IDEA shorten_it -->
+            <pattern>
+                <token postag='V.+|NC.+|AQ.+' postag_regexp='yes'>
+                    <exception scope='previous' postag_regexp='yes' postag='SPS00:[DP][ADIPRT].+'/>
+                </token>
+                <token min='0' max='1' postag='_QUOT|_PUNCT|RG' postag_regexp='yes'/>
+                <marker>
+                    <token>que</token>
+                    <token regexp='yes'>é|são</token>
+                    <token regexp='yes'>necessários?</token>
+                </marker>
+            </pattern>
+            <message>&simplify_msg;</message>
+            <suggestion>\5</suggestion>
+            <example correction="necessário">Acho <marker>que é necessário</marker> você ir lá.</example>
+            <example correction="necessários">Ele considera sempre <marker>que são necessários</marker> 2/3 de energia.</example>
+            <example>O médico disse à polícia que é necessário fazer uma necropsia para saber a causa da morte da mulher.</example>
+        </rule>
+
+
         <rule id='DAR_VALOR_VALORIZAR' name="Simplificar: Verbo 'dar' + valor → valorizar" tags="picky">
             <!-- IDEA shorten_it -->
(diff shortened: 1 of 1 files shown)


p-goulart (changes requested):
Based on the results, I do think the AP shouldn't just rely on the preposition (which, I suppose, is there to block verbs like 'dizer'). I also see:
- _Todos nós sabemos ~que é~ necessário acelerar o Programa_: 'sabemos necessário' is very awkward
- _entendo ~que é~ necessário_: also awkward IMD.

The same will happen with a few other verbs that wouldn't accept an adjective as a complement: *'acredito necessário', *'creio necessário'.

marcoagpinto:
@p-goulart 

Heya, Pedro, I will work on it tonight.

Thanks for the feedback.

marcoagpinto:
I did have some doubts with certain results, but it always happens when one is analysing line by line and the brain gets “melted” and one can no longer reason.

p-goulart:
Yes, that happens 🫠

For one, this rule seems like a good candidate for us to expand with more adjectives, as this pattern isn't restricted to 'necessário'... Though if we do that, we need to work on in several iterations, checking the nightly results and adapting over several days/weeks.

marcoagpinto:
> Yes, that happens 🫠
> 
> For one, this rule seems like a good candidate for us to expand with more adjectives, as this pattern isn't restricted to 'necessário'... Though if we do that, we need to work on in several iterations, checking the nightly results and adapting over several days/weeks.
> 
> We can keep it restricted to 'necessário' for now, of course, but I do think we ought to expand. Maybe change the rule ID to reflect this intention?

@p-goulart 
There are always improvements that can be made as we create a rule and come up with more ideas on how to enhance it.

It is what the Americans call: “refinement and betterment”.

❤️ ❤️ ❤️ ❤️ ❤️

marcoagpinto:
@p-goulart 

Heya,

I have created an AP to fix the FPs.

BEFORE:
```
Portuguese (Portugal): 15 total matches
Portuguese (Portugal): 599999 total sentences considered
```
[0_before.txt](https://github.com/languagetool-org/languagetool/files/12269522/0_before.txt)


AFTER:
```
Portuguese (Portugal): 11 total matches
Portuguese (Portugal): 599999 total sentences considered
```
[5_after.txt](https://github.com/languagetool-org/languagetool/files/12269525/5_after.txt)

Antipattern removed the sentences in the examples:
```
            <antipattern>
                <token postag='VMIP.+|VMSF.+' postag_regexp='yes'/>
                <token>que</token>
                <token regexp='yes'>é|são</token>
                <token regexp='yes'>necessários?</token>
                <token postag='VMN03.+|RG|CS' postag_regexp='yes'/>
                <token postag='(SPS00:)?D[AI].+' postag_regexp='yes'/>
                <example>Sinceramente, não recomendo esse tipo de pratica pois entendo que é necessário sempre um ambiente profissional para garantir a qualidade e a produtividade.</example>
                <example>Todos nós sabemos que é necessário acelerar o Programa de Aceleração do Crescimento", emendou ele, usando um trocadilho para se referi...</example>
                <example>Com a incorporação da Adequação Sociotécnica à Teoria, passa-se a entender que é necessário que um processo seja concebido nas condições em que ele acontecerá.</example>
                <example>Para atingir esta missão fundamental da Igreja, o documento declara que é necessário dar aos católicos "uma 'consciência de Igreja' mais coerente, para que também se possam valorizar a...</example>
            </antipattern>
```

marcoagpinto:
Thank you, dear Pedro!

😋 🤗

PR #9104, opened 2023-08-08
[pt] Fix PT/BR confusion pair corrections

p-goulart:
Closes languagetooler-gmbh/languagetool-premium#5563 .

---

We load both the `PT` and `BR` list of words and read from them depending on the specific dialect from the `RuleMatch`, as explained by @jaumeortola. We also have a new test, `PortugueseDiacriticsRuleTest`, which makes sure this works as expected.

The `pt-PT` and `pt-BR` files have been split to the best of my knowledge, but it is entirely possible that I have missed some words from the `pt` file.


jaumeortola:
What are the PT/BR differences? Is it just `ô` <> `ó`? Anything else?

After some changes I did some weeks ago, we can know what variant is being used inside a filter. 

```
  private String getLanguageVariantCode(RuleMatch match) {
    PatternRule pr = (PatternRule) match.getRule();
    return pr.getLanguage().getShortCodeWithCountryAndVariant();
  }
```

Using this, there is no need to duplicate the Java filter or the XML rules. We only need some way to translate the words from one variant to the other. We could add a new column in the `confusion_pairs.txt`, if necessary.

p-goulart:
> What are the PT/BR differences? Is it just ô <> ó? Anything else?

It's not biunivocal correspondence, sometimes both languages use the circumflex, and sometimes `pt-PT` prefers the acute.

> After some changes I did some weeks ago, we can know what variant is being used inside a filter.

I think this is enough to keep the same filter and the same rules, and just dynamically choose which `confusion_pairs.txt` file to load.

---

@susanaboatto do you think you could unsplit the rules and Java filters and just load the TXT file dynamically based on this:
```java
private String getLanguageVariantCode(RuleMatch match) {
    PatternRule pr = (PatternRule) match.getRule();
    return pr.getLanguage().getShortCodeWithCountryAndVariant();
  }
```

We'd still load **two** confusion pair TXT files, look at this method:
```java
 private List<String> getFilepaths() {
    List<String> paths = new ArrayList<>();
    paths.add("/pt/confusion_pairs.txt");
    // we'd just need to declare shortCodeWithCountry somewhere, or resolve it here
    // if we use @jaumeortola 's code from above we'll also need to add a `RuleMatch`-typed param to this method
    paths.add("/pt/" + shortCodeWithCountry + "/confusion_pairs.txt");
    return paths;
  }
```

jaumeortola:
We have to preload both files in the filter: 
```
private final Map<String, AnalyzedTokenReadings> relevantWordsPT = new ConfusionPairsDataLoader().loadWords(getFilepathPT());
private final Map<String, AnalyzedTokenReadings> relevantWordsBR = new ConfusionPairsDataLoader().loadWords(getFilepathBR());
```

And use one or the other according to the language variant from the RuleMatch.

p-goulart:
@susanaboatto , @jaumeortola this is ready for review now, I've applied the changes suggested by @jaumeortola and it works as expected. Might want to add a test for this, though...

jaumeortola (approved):

PR #9110, opened 2023-08-09
[fr] Changes in messages, simple-double quote and final punctuation

LTLucieOld:
In the context of https://github.com/languagetooler-gmbh/task-force-french/issues/52
**Changes in messages:**
- simple quotes surrounding words changed to double quotes
- changes in messages not ending properly
- changing also a few formulations

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
@@ -10087,7 +10087,7 @@ Pour en savoir plus sur les tone tags, suivez le lien vers notre documentation (
                     <token>e</token>
                     <token>.</token>
                 </pattern>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
```diff
@@ -10087,7 +10087,7 @@ Pour en savoir plus sur les tone tags, suivez le lien vers notre documentation (
                     <token>e</token>
                     <token>.</token>
                 </pattern>
-                <message>L'abréviation 'i.e.' peut être considéré comme un anglicisme.</message>
+                <message>L'abréviation "i.e.' peut être considéré comme un anglicisme.</message>
```
GillouLT, line 10090: Only "i.e.'
The rest is great! :)

PR #9114, opened 2023-08-10
[en] An XML rule to detect 'to' after modal verbs

LeahKorol:


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -83988,6 +83988,29 @@ USA
             <example correction="must|have to">I <marker>must to</marker> go shopping.</example>
             <example>It is a must to be aware of that.</example>
         </rule>
+        <rule id="TO_AFTER_MODAL_VERBS" name="to after modal verbs" default="temp off">
```
evan-defran-lt, line 83991: Hi @LeahKorol, the correct value here of the attribute `default` is `"temp_off"`, with the underscore.

evan-defran-lt:
Hi @LeahKorol, thanks for contributing to LanguageTool! 

Before merging, you'll need to make a tiny correction in line 83991: the value of the attribute in this case should be `"temp_off"`, with an underscore. 

Note also that your rule will also flag sentences like _You should do all that you can to help your friends_, which is grammatical, so you might want to think about more narrowly writing the pattern or adding additional antipatterns. 

Thanks again!

LeahKorol:
Hi @evan-defran-lt ,
The rule won't flag sentences like _You should do all **that** you can to help your friends_
thanks to this line:  ` <token regexp="yes">that|(what|which|who)(ever)?</token>`

I ran the command: mvn --projects languagetool-language-modules/en --also-make clean test,
but it didn't detect the missing underscore in temp_off, nor did it identify any other compilation errors.
What did I do wrong?

Thank you for your time and assistance.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -83988,6 +83988,29 @@ USA
...
+            <token chunk_re="E-NP.*" />
+            <token postag="RB" min="0" />
+            <token regexp="yes">can|could|may|might|should|shall</token>
+            <example>I will do whatever I can to help.</example>
+         </antipattern>
+         <pattern>
+             <token regexp='yes' postag='MD'>can|could|may|might|should|shall</token>
+             <token>to</token>
+         </pattern>
+         <message>Modal verbs are followed directly by the infinitive of another verb, without 'to'. Probably you should use <suggestion>base form</suggestion></message>
```
evan-defran-lt, line 84009: This suggestion will suggest the literal phrase _base form_. 
There are a couple of ways to suggest what you want. What I would probably do: 
Extend the pattern:
```
    <pattern>
        <token regexp='yes' postag='MD'>can|could|may|might|should|shall</token>
        <token>to</token>
        <token postag="VB"/>
    </pattern>
```
And for the suggestion: `<suggestion>\1 \3</suggestion>`

Don't forget, then to also modify the `example correction` and the `marker` in it. 

But there are various approaches here.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -83988,6 +83988,29 @@ USA
...
             <example>It is a must to be aware of that.</example>
         </rule>
+        <rule id="TO_AFTER_MODAL_VERBS" name="to after modal verbs" default="temp_off">
+         <antipattern>
+            <token regexp="yes">an?</token>
+            <token regexp="yes" min="0">absolute|total</token>
+            <token regexp="yes">can|could|may|might|should|shall</token>
+         </antipattern>
+         <antipattern>
+            <token regexp="yes">that|(what|which|who)(ever)?</token>
```
evan-defran-lt, line 83998: You can extend this token to include `all|everything|anything`
This will remove a lot of FPs like _We will do everything we can to assist_

evan-defran-lt:
Hi @LeahKorol, 

### Maven test
I'm not sure why your maven test didn't fail, as it definitely should have with `default="temp off`. 

You should have gotten an error message like:
>[ERROR]   EnglishPatternRuleTest.testRules:37->PatternRuleTest.runGrammarRulesFromXmlTest:152-PatternRuleTest.runGrammarRuleForLanguage:174->PatternRuleTest.runTestForLanguage:189-PatternRuleTest.validatePatternFile:218->PatternRuleTest.validatePatternFile:242 » IO Cannot load or parse '/org/languagetool/rules/en/grammar.xml'`

This means there's something wrong with the syntax of the XML code. If you can't spot the syntax error, you could open up the surefire-report for `EnglishPatternRuleTest` to find out where the error occurs. You'll see something like:

>     <system-err><![CDATA[cvc-datatype-valid.1.2.1: 'temp off' is not a valid value for 'NMTOKEN'. Problem found at line 25720, column 88. ]]></system-err>


### Antipatterns
> The rule won't flag sentences like You should do all that you can to help your friends
thanks to this line:  <token regexp="yes">that|(what|which|who)(ever)?</token>

Yes, you're right. I didn't catch that. Your pattern still catches sentences like _You should do all you can to help your friends_, though. You can remove those FPs easily by adding to one of your antipatterns. I've offered a solution in-line. 

### Suggestions
You'll want to amend your suggestion as well. Right now, it will suggest the literal phrase _base form_. I've offered one possible solution in-line, but there are several! 

I hope this is helpful. Thanks again for contributing to LanguageTool!

evan-defran-lt:
By the way, I'm not sure if you already know this, but you can also test your rules in LT's "expert" RuleEditor here: https://community.languagetool.org/ruleEditor/expert?devMode=true

It will run your rule over a large corpus of sentences, allowing you to see how your rule performs, which means you can easily identify TPs and FPs. It will also flag some errors that would break the build. It's not perfect – you cannot test rule groups with it, for instance – but it is a helpful resource to start!

LeahKorol:
Hi @evan-defran-lt,
I wanted to let you know that I followed your instructions, and they were incredibly helpful.
Thank you!
There are a few new antipatterns according to the sentences the LT's "expert" RuleEditor provided.

By the way, the project was too large for any IDE I've attempted to work with (the compiler wasn't available). Is there a way to address this issue?

evan-defran-lt:
Congratulations on you first LT rule! 

> By the way, the project was too large for any IDE I've attempted to work with (the compiler wasn't available). Is there a way to address this issue?

It's a known issue, and it's something we're trying to address!

PR #9135, opened 2023-08-17
[fr] LEVEE_DE_FONDS Removed

GillouLT:
Bad results of the rule, also related to: https://github.com/languagetooler-gmbh/languagetool-premium-modules/pull/876


LucieSteib (approved):
That all makes sense 🥜 🥜 🥜

PR #9164, opened 2023-08-23
allow multiple ab-tests & add ab-test for ggec

SteVio89:


fabrichter (changes requested):
Can we introduce a general `abtest` option that is set in remote rule configurations (under `options` instead of hard-coding this?

fabrichter (changes requested):
Just one small change, then LGTM

fabrichter (approved):

PR #9169, opened 2023-08-24
[pt] Add entries to post-reform-compounds.txt

p-goulart:
- we were pretty much only covering words that require hyphens, but not those that specifically MUST NOT be hyphenated;

 - we still don't have a good solution for "ultrassom" (from ultra-som);

 - also add a new Java test to make sure this works as expected.

(This is kind of related to the `SUPER_ADJ_ADV` rule.)


susanaboatto:
> we still don't have a good solution for "ultrassom" (from ultra-som);

There is a fix for this in one of the subrules for AUTO in the grammar.xml file:

```
            <rule>
                <pattern>
                    <token>auto</token>
                    <token min='0' spacebefore='no'>-</token>
                    <token regexp='yes'>s\p{Ll}+
                        <exception postag='SPS.+|C.+|RG|D.+|P.+' postag_regexp='yes'/>
                    </token>
                </pattern>
                <message>Esta é uma palavra só.</message>
                <suggestion>\1s\3</suggestion>
                <example correction="autossabotagem">Isso é muita <marker>auto sabotagem</marker>.</example>
            </rule>
```

Perhaps we could broaden it to include most/all prefixes?

susanaboatto:
Hm, true, I suppose it doesn't work with augmentatives. But we could still create a rule for _ultrassom_ specifically, just to have it detected in _some_ way

susanaboatto (approved):

PR #9172, opened 2023-08-24
[pt] Add ABREVIATURAS_FEMININAS_COM_SOBRESCRITO

p-goulart:
I've finally got fed up with all the `profa` and `dra` in the logs. This should address it. The list of abbreviations is relatively small and unambiguous, to keep things simple. A good candidate for a picky typography rule.

We can, in the future, add sister rules for stuff like `Engo.`, but I feel those will be much rarer anyway.

---

There are other abbreviations that might require the superscript in some situations, but there is very limited support for other superscript characters in most fonts (despite some [Unicode support](https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts), though do note the caveat with regards to their size and position). I'd steer clear of adding support for those more obscure cases for now.

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -18972,6 +18972,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
             <example correction="et al.">SANTOS, T.F. <marker>et al</marker> (2005),…</example>
         </rule>
 
+        
         <rulegroup id='ABREVIATIONS_PUNCTUATION' name="Abreviaturas: Pontuação">
             <!--       Based on the Catalan version, Tiago F. Santos,  2017-02-18      -->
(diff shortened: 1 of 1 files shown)


susanaboatto:
Do you think we could make the `.` optional? This wouldn't detect `Gostaria de ir com a Profa Goldberg`, for example. And then we remove `profa|dra|exa|enga|sra|reva|srta` from `ABREVIATIONS_PUNCTUATION`, which could also help with the apply rates for that rule.

Also, I suspect people might write `a prof` out of laziness as well, but that would be another issue.

susanaboatto (approved):

p-goulart:
> Do you think we could make the . optional? This wouldn't detect Gostaria de ir com a Profa Goldberg, for example. 

That was actually a problem I had while writing this, I was hoping you'd pick up on that. We _can_ do that, but then we'd be tying together **two** different rules. The one about adding a dot to those abbreviations (`dra` -> `dra.`) and the one about adding the superscript.

My thinking here was we only want to correct the superscript in `picky` mode, and only if the period is already there, making it 100% clear that it's an abbreviation. It's strictly a typographical thing... Right? I don't know any more lol

PR #9176, opened 2023-08-25
[pt] suggestion improvements in HIPHEN_SPACE_RULES

susanaboatto:


p-goulart (approved):
I can't say I fully *see* 👁️ what the issue is, but this seems small enough :p

susanaboatto:
There is the em dash (—) and the en dash (–). I noticed that when the en dash was used, this rule did not keep it as a suggestion, and it should be one.

PR #9202, opened 2023-08-28
[pt] Add CONCORDANCIA_SOBRESCRITO rule

p-goulart:
Simple rule that replaces older wikipedia rule. Has better coverage.


susanaboatto:
This is already covered by our gender agreement rule, isn't it? 

image

p-goulart:
As discussed [here](https://languagetooler.slack.com/archives/C058BK14VND/p1693314465224159?thread_ts=1693311656.099239&cid=C058BK14VND), we'll keep this rule separately so as not to `<pt_br>`poke the jaguar with too short a stick`</pt_br>`.

Also changes to disambiguation.

p-goulart:

Issue #9203, opened 2023-08-28
Remove trove4j dependency


donnerpeter:
LanguageTool is currently using Trove4J for more memory-efficient storage. Unfortunately, this library is quite old, not maintained, and there are better alternatives. IntelliJ is using LanguageTool as a bundled grammar checker. The IDE is trying to remove Trove from classloading completely, so we'd prefer to switch to [fastutil](https://github.com/JetBrains/intellij-deps-fastutil). It has a somewhat larger jar size (6 MB), but is actively maintained and more performant.

We can prepare a pull request for this migration if there are no objections.

SteVio89:
Hi @donnerpeter 

you can prepare a pull request for this migration. Thank you!

donnerpeter:
@SteVio89 any updates? :)

SteVio89:
Sorry again for the delay, we also need to change the code in our premium repository for this change. I have it on my list, but did not make it yet. I will solve this as soon as possible.

donnerpeter:
@SteVio89  Any progress?

PR #9210, opened 2023-08-29
[#9203] Replace trove4j with fastutils

FirstTimeInForever:
Replace the obsolete `trove4j` library with the `fastutil` as per #9203.


languagetool-core/pom.xml
```diff
@@ -208,8 +208,8 @@
             <artifactId>annotations</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.jetbrains.intellij.deps</groupId>
-            <artifactId>trove4j</artifactId>
+            <groupId>it.unimi.dsi</groupId>
+            <artifactId>fastutil</artifactId>
```
donnerpeter, line 212: Maybe use a stripped version from IJ?

FirstTimeInForever: Oh, yes, I kind of forgot about it. But we decided to just use the official `fastutil-core`, as the IJ version does not worth the trouble to use outside of IJ.

donnerpeter: I hope it's OK to add a 22 MB dependency

FirstTimeInForever: `fastutil-core` is only `6.1MB`. See the commit below.

PR #9211, opened 2023-08-29
[pt] Add DUAS_MILHOES and AS_MILHARES_DE_PESSOAS rules

p-goulart:
An offshoot from the recent agreement issues. These appear to be frequently disabled by users, and indeed using the masculine there actually _sounds_ off to me, so I think we should make it `picky` and let users decide.


p-goulart:

susanaboatto (approved):

PR #9213, opened 2023-08-29
[pt] Improve compounding rules

p-goulart:
- move a few common disables from the compounding file to the grammar XML so we can have more granular logic;
- add new rule for colour name hyphenation, since that's something that people commonly disable – that waay, users will have more control over LT's behaviour when it comes to hyphenating colour names, without us losing any coverage (and the performance of the compounding rule should improve, to boot).

Blocked by #9216 ⚠️


p-goulart:

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -38791,6 +38791,71 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
     <category id='COMPOUNDING' name="Palavras Compostas" type="misspelling">
+        <rule id="SEM_ABRIGO" name="Composição de sem-abrigo">
+            <pattern>
+                <token postag_regexp="yes" postag="[ADP].+M[SP].+|Z.+|SENT_START"></token>
```
susanaboatto, line 38796: I'm a bit weary of including adjectives here, I feel like some FPs will happen. Hopefully few enough to deal with with antipatterns. For example: `Eu com casa e o bonito sem abrigo`

Also with `DP*`:

`O meu em casa e o seu sem abrigo`

Maybe adjectives and possessives should be dealt with separately.

susanaboatto (approved):

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: Hm, you're not wrong, I'll remove adjectives from here and see what happens.

PR #9216, opened 2023-08-30
Add normalisation logic to toId in StringTools

p-goulart:
- German has some specific logic with 'ä' -> 'ae', but this is isn't necessary in the other languages;

 - pass language code as a param so the rule ID conversion method knows what to do with non-ASCII characters;

 - since a new param is passed, all classes that need sub-rule-specific IDs needed to have a new Language-type property.

All relevant tests pass locally for both core and all locales (though the maven install does fail on the HTTP server tests, as usual).

---

I'll make parallel changes to the premium repository in a moment.


languagetool-core/src/main/java/org/languagetool/rules/AbstractCheckCaseRule.java
```diff
@@ -104,7 +106,7 @@ public RuleMatch[] match(AnalyzedSentence sentence) {
           }
           ruleMatch = new RuleMatch(this, sentence, startPos, endPos, msg, getShort());
           if (subRuleSpecificIds) {
-            ruleMatch.setSpecificRuleId(StringTools.toId(getId() + "_" + correctPhrase));
+            ruleMatch.setSpecificRuleId(StringTools.toId(getId() + "_" + correctPhrase, language.getShortCode()));
```
danielnaber, line 109: Could you use `language` instead of its code? This leads to better type-safety.

languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/ContractionSpellingRule.java
```diff
@@ -45,8 +46,8 @@ public Map<String, List<String>> getWrongWords() {
     return wrongWords;
   }
 
-  public ContractionSpellingRule(ResourceBundle messages) {
-    super(messages);
+  public ContractionSpellingRule(ResourceBundle messages, final Language language) {
```
danielnaber, line 49: You can remove the `final` here. It's correct to have it and in theory it makes the code more clean, but it's also very verbose, so we usually don't use it for parameters.

danielnaber (changes requested):

jaumeortola (approved):

languagetool-core/src/main/java/org/languagetool/rules/AbstractCheckCaseRule.java
p-goulart: Done!

languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/ContractionSpellingRule.java
p-goulart: Done!

danielnaber (approved):

PR #9219, opened 2023-08-30
[pt-PT] Added "ENTITY % messages SYSTEM" to style.xml

marcoagpinto:
Heya, @p-goulart 

I noticed that you had added it to the pt-PT grammar.xml already, thank you.

I have now added it to the pt-PT style.xml as well.

My goal for 2024 will be to revise all pt-PT rules and improve them the maximum possible, adding these messages.

The year is ending, which means that in this year it isn't doable… too much going on… I can't focus 100% on specific tasks.

❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️


p-goulart:
Thanks a tonne for this!

Our focus is obviously pt-BR but to the extent it's possible those improvements should also help you make pt-PT better.

I think at some point we may want to actually _split_ the `message.ent` files completely to keep the messages as compatible as possible with users' specific dialect choices.

p-goulart (approved):

PR #9243, opened 2023-09-04
[de] extended list

Luisa-LT:
@danielnaber in contect of [this issue](https://github.com/languagetooler-gmbh/languagetool-premium/issues/5876#issuecomment-1702862925) I added only _Wahrung_ to the list since _Jung_ was already added. Is it correct like this, or did I miss something?


danielnaber:
merged manually

PR #9268, opened 2023-09-06
[pt] Review GENERAL_VERB_AGREEMENT_ERRORS: basic cases

### LINKED ISSUE #8945: [pt] LanguageTool  fails to detect very basic subject-verb agreement errors
See the two wrongly written sentences just below:

> Os meninos gosta das meninas

> O menino gostam das meninas 

The correct sentence should be:

> Os meninos gostam das meninas

-------------

**EXTRA**

These ones are more complicated. LT doesn't detect that the two sentences just below are wrong:

> Poderá ainda acontecer mais incentivos como esses.

> Deu dez da noite e ele ainda não chegou.

The correct forms are:

> Poderão ainda acontecer mais incentivos como esses.

> Deram dez da noite e ele ainda não chegou.

------------

I took these sentences from [here](https://blog.imaginie.com.br/erro-de-concordancia/).

p-goulart:
# Agreement review

This first PR takes care of the **basic** cases from #8945. The 'extra' ones will probably take extra rules, so that'll be a separate PR.

## Sub-rule 12

This takes care of the 'plural subject with singular verb' case. We're disabling some APs and updating a few of them, as well as adding and moving examples around for more robust testing.

## Sub-rule 13

This takes care of the 'singular subject with plural verb' case. Here I couldn't nuke too many APs without causing a bloodbath in the current examples, so instead I created a few extra rules (lettered, so `13A`, `13B`, and `13C`) to cover some really unambiguous cases.

---

In addition to solving issues, we also:

 - remove needless comments, update examples, add some basic errors to check sanity;
 - update the disambiguator;
- add some in-line comments (not making the file longer than it has to be) just to help me navigate the different sub-rules and APs of the agreement rules.

I'm really sorry that the changes don't line up well enough for the diff to be displayed in a more convenient way 😩

The nightly diffs will also be a complete catastrophe with IDs moving back and forth – I have no way how to counter that.

<sup>The parent issue in the sprint board is languagetooler-gmbh/languagetool-premium#6261</sup>


susanaboatto:
All right, this was a bit hard to read. My takeaway is that so many of these antipatterns need to be more specific, particularly when it comes down to the singular/plural form of postags. We have so many general postags that could be otherwise hindering accurate corrections, but it'd be impossible to go through all of them in a PR review. I think one way we could go about this is to first target the antipatterns that actually _have_ examples and see how we can improve there. Then we move to the tougher, exemple-less ones. 

For now, I think it's safe to give it a green light. I'm sure the diff results will be somewhat enlightening.

Issue #9264, opened 2023-09-06
[de] Improve suggestions for typo "schammiger"


Discostu36:
![grafik](https://github.com/languagetool-org/languagetool/assets/30999326/a2e8885c-aa57-4105-9113-75c4624fb417)

I think "gammliger" shoud not be in this list at all, "schlammiger" and "schwammiger" should be the first suggestions.

Luisa-LT:
Thank you, @Discostu36, for pointing this out. Unfortunately, the order of the suggestions is not pre-definable. If there were only 1 right suggestion, we could offer just this 1 suggestion. In this case, as there is more than just 1 correct suggestion (schlammiger/schaumiger/schwammiger...) the user can choose the most proper correction.

Discostu36:
after some trial and error (😊 ) I have created a pull request for a partial fix to at least get the "schwammiger" suggestion

PR #9271, opened 2023-09-07
add suggestion

Discostu36:
partial fix of  #9264

### DIFF
--- a/languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
+++ b/languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
@@ -1548,6 +1548,7 @@ public class GermanSpellerRule extends CompoundAwareHunspellRule {
     put("löchen", w -> Arrays.asList("löschen", "löchern", "Köchen"));
     put("wergen",  w -> Arrays.asList("werfen", "werben", "werten"));
     put("Wasn",  w -> Arrays.asList("Was denn", "Was ein", "Was"));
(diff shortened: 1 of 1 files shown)


Luisa-LT:
Thanks for trying this out! What about the case a user actually wanted to write "schlammig" or "schaumig"?
Bsp:
> Wie war die Party? Sie war sehr schwammig, schlammig und schaumig. :-D

Discostu36:
"schlammig" and "schaumig" would still be in the suggestions, won't they? This is only adding to the suggestions, not replacing them? For example:
https://github.com/languagetool-org/languagetool/blob/bdd98c2bfe1908b0e88901863304e907209242a8/languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java#L97 
"leider" and "Lieder" are addded to the suggestions, but LanguageTool ist still also suggesting "wieder", "nieder" and "Dieter".
grafik

Luisa-LT:
That's true, seems like you're absolutely right about that, thanks for the example and let's try it then :)

Discostu36:
grafik

It only kind of worked. "schlammiger" is now gone from the suggestions. Maybe I should overwrite suggestions entirely by adding everything that should be there as an array ("schwammiger", "schlammiger", "schaumiger")?

Luisa-LT:
Thanks for checking so quickly! Yes, perhaps you can try this approach:   ` put("Wasn",  w -> Arrays.asList("Was denn", "Was ein", "Was"));`

Discostu36:
Yes, but with RegEx for the different word forms.

PR #9283, opened 2023-09-08
[de] add "Typst" to spelling ignorelist

NicoWeio:
https://github.com/typst/typst/ is sufficiently popular, with >20k stars on GitHub.


NicoWeio:
Actually, shouldn't this be included into the ignorelist of each language? Or is there a better way?

jaumeortola:
> Actually, shouldn't this be included into the ignorelist of each language? Or is there a better way?

Added to the global spelling file (for all languages): https://github.com/languagetool-org/languagetool/commit/ac6333a699a5c634a682cb2c08224e722f83df01

NicoWeio:
Thanks, I didn't know that file existed! 
Maybe the [documentation](https://dev.languagetool.org/hunspell-support) can be improved in that regard.
I saw you requested a review, @jaumeortola, so I'll keep this PR open for the moment.

Luisa-LT:
Hi, I think we should not add it to the German ignore.txt if it's already added to global spelling, right @jaumeortola?

PR #9315, opened 2023-09-14
[fr] COULEURS_SPE off rule, deletion

LucieSteib:
- COULEURS_SPE rule was off
- It would have been a style rule anyway
- the tags!"picky creative" wasn't right 

See also https://github.com/languagetooler-gmbh/languagetool-premium/issues/6327


GillouLT (approved):
LGTM! Thanks for fixing the tags issue :)

Issue #9299, opened 2023-09-12
[pt] Idea for antipattern - 2023-09-12


marcoagpinto:
Heya, @ricardojosehlima

Sorry for falling asleep… 😋 

I wanted to create an antipattern for:
`"Eu sou letras, prazer, poesia, pecado, redenção."`

![Screenshot 2023-09-12 at 16-53-29 Editor](https://github.com/languagetool-org/languagetool/assets/5192600/4166b143-a044-4df7-aa75-a49c8cf9f1a0)

![predicativo](https://github.com/languagetool-org/languagetool/assets/5192600/09e34b4f-371a-4365-9a61-18107c1219e7)

`Verb Ser + NCPlural + "," + NC`

Is it right only for pt-PT or to all variants of PT?

Thank you, my dear brother!

❤️ ❤️ ❤️ ❤️ ❤️

marcoagpinto:
Maybe it will work with all verbs with the form of “eu” and “nós”?

“(eu) quero letras”, “(nós) queremos letras”?

marcoagpinto:
`"Ele É letras, poesia (...)"`

"é" should also be correct? Maybe the whole verb ser?

ricardojosehlima:
This seems a rare construction to me, but keeping the verb in the singular seems fine.
Worth looking and check if there are false positives

marcoagpinto:
@ricardojosehlima 

Thank you!

I will work on it tonight.

❤️ ❤️ ❤️ ❤️ ❤️

marcoagpinto:
It removed zero hits in 900 000 sentences.

PR #9323, opened 2023-09-14
[pt] Added antipattern to rule ID:LINKING_VERB_PREDICATE_AGREEMENT

marcoagpinto:
Heya, Susana and Pedro,

Here is an antipattern.

In a 900 000-sentence check, it didn't remove a single hit.

This raises the question:
`<token inflected='yes' regexp='no'>ser</token> <!-- Add more verbs here when/if found. -->`

Should it accept all verbs, or just verb “ser” and others we may find?

https://github.com/languagetool-org/languagetool/issues/9299

Thanks!


p-goulart (approved):
> Should it accept all verbs, or just verb “ser” and others we may find?

The only other linking verb that has this kind of agreement pattern (potentially agreeing with plural predicates), is 'estar', and I don't see how it'd be used in this kind of context. I think just 'ser' is enough for this.

PR #9332, opened 2023-09-15
[fr] AIT_AIE_AIES_EST_and_ACCORD_V_PPA_fixed

GillouLT:
AIT_AIE_AIES_EST[14]
- FP found and fixed while working on https://github.com/languagetooler-gmbh/recurring-issues/issues/3
Explanation: "À qui ai-je l'honneur ?" when writen without hyphen triggers AIT_AIE_AIES_EST[14] when it should actually trigger the hyphen rule (which has lower prio), with this change, the hyphen rule will directly trigger.

ACCORD_V_PPA[1]
- AP improved to fix: https://github.com/languagetooler-gmbh/languagetool-premium/issues/6342


LucieSteib (approved):
🥜 LGTM! 🥜

PR #9347, opened 2023-09-18
[nl] improve SpaceInCompoundRule, move compound words

mark-baas:


danielnaber (approved):
Looks okay, but please run it locally on at least 20,000 sentences before and after this change, just to be sure (contact me if you're not sure how to do that).

PR #9361, opened 2023-09-20
[pt] clean up/add exceptions to CIENTÍFICO_FAZER_EFETUAR_REALIZAR_CONDUZIR_CONCRETIZAR_ELABORAR; make it goal-specific

susanaboatto:
- Casually noticed false positives in this rule; this fixes some of them;
- removed comments from rule;
- made this rule goal-specific due to its generality. It [has _many_ opens and not enough applies (plus high disable rates)](https://internal1.languagetool.org/grafana/d/BY_CNDHGz/rule-events-analysis?orgId=1&from=now-30d&to=now&var-rule_id=CIENT%C3%8DFICO_FAZER_EFETUAR_REALIZAR_CONDUZIR_CONCRETIZAR_ELABORAR&var-language=pt)


p-goulart (approved):
let's add some more words to the rule ID

susanaboatto:
Right? I figured it's too late to change it now. But the length bothers me as well 🫠

Issue #9355, opened 2023-09-19
[pt] "Aileron" and "estabilizador" (stabilizer) are different things


brunnopleffken:
When I type "aileron", it suggests "stabilizer". When referring to aircraft controls, ailerons and stabilizers are NOT the same thing. They're completely different parts and mislead the writer.

Ailerons are small moving surfaces at the tips of the wings, and they control the roll of the aircraft. Stabilizers are those horizontal and vertical wings in the back.

"Aileron" is a word recognized by the Brazilian national civil aviation agency (source: https://www2.anac.gov.br/anacpedia/por_por/tr112.htm)

<img width="331" alt="Screenshot 2023-09-19 at 08 28 25" src="https://github.com/languagetool-org/languagetool/assets/3427344/67ee07d3-9fb5-4ad4-a596-d2960ca80384">

susanaboatto:
Thanks for the feedback. I have now fixed this. It could be a while until changes are effective.

PR #9362, opened 2023-09-20
[pt] remove non-sensical barbarisms (#9355)

susanaboatto:
Barbarism fix based on feedback #9355


p-goulart (approved):
ahahaha [spoiler](https://en.wikipedia.org/wiki/Spoiler_(aeronautics)) that's amazing

PR #9376, opened 2023-09-23
[pt-PT] Improved antipattern in rule ID:POSSESSIVE_WITHOUT_ARTICLE

marcoagpinto:
Heya, Susana and Pedro,

This enhancement fixes false positives that have been around for a very long time, and I should have done it before.

But it is better late, than never.

❤️ ❤️ ❤️ ❤️ ❤️

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -1399,6 +1399,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
         <rulegroup id='GENERAL_GENDER_AGREEMENT_ERRORS' name="Concordância: Geral">
             <url>https://pt.wikibooks.org/wiki/Portugu%C3%AAs/Concord%C3%A2ncia/Concord%C3%A2ncia_nominal</url>
             <antipattern>
+                <token inflected='yes'>impor</token>
                 <token postag='SPS00:DA0MP0' postag_regexp='no'/>
                 <token>outros</token>
                 <token postag='AQ..P.+|NC.P.+' postag_regexp='yes'/>
@@ -8430,15 +8431,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
             <!-- actual disambig issue, should be fixed now -->
             <example>os espaços não formais são</example>
             <example>as bagagens não convencionais agora desaparecerão</example>
-        </antipattern>
-
-        <antipattern>
-            <token postag='(SPS00:)?[DP][ADIPRT]..P.+' postag_regexp='yes'/>
-            <token postag='AQ..P.+|NC.P.+' postag_regexp='yes'/>
-            <token min='1' max='2' postag='RG' postag_regexp='no'/>
-            <token postag='VMIS.+' postag_regexp='yes'/>
-            <token postag='VMN0000' postag_regexp='no'/>
-            <example>As unhas nunca mais pude arranjar.</example>
         </antipattern>
 
         <rule> <!-- #1: eu + verb -->
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -1398,6 +1398,12 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 
         <rulegroup id='GENERAL_GENDER_AGREEMENT_ERRORS' name="Concordância: Geral">
             <url>https://pt.wikibooks.org/wiki/Portugu%C3%AAs/Concord%C3%A2ncia/Concord%C3%A2ncia_nominal</url>
+            <antipattern>
+                <token postag='SPS00:DA0MP0' postag_regexp='no'/>
+                <token>outros</token>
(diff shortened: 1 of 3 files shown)


marcoagpinto:
And another antipattern for other rule (id:PARA-POR_TER_PARTICIPIO-PASSADO)

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -1398,6 +1398,12 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 
         <rulegroup id='GENERAL_GENDER_AGREEMENT_ERRORS' name="Concordância: Geral">
             <url>https://pt.wikibooks.org/wiki/Portugu%C3%AAs/Concord%C3%A2ncia/Concord%C3%A2ncia_nominal</url>
+            <antipattern>
```
p-goulart, line 1401: This antipattern technically allows for:
> o que você acha dos outros regras?
> quero dar mais consistência aos outros regras

Both of which are incorrect, so I don't think we should be introducing this.

---

The key word here is the **verb**. If you have a verb that governs a certain preposition (such as 'impor', in your example), then we should allow these constructions. I'd suggest adding an extra token accounting for this verb.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -8426,6 +8432,15 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
             <example>as bagagens não convencionais agora desaparecerão</example>
         </antipattern>
 
+        <antipattern>
```
p-goulart, line 8435: The key blocker of agreement here is the verb in the first-person singular, not the whole construction. I'd rethink this.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/pt-PT/grammar.xml
```diff
@@ -1423,16 +1423,27 @@
...
 
-        <!-- MARCOAGPINTO 2022-11-25 (Checked/Enhanced) (25-JUL-2022+) *START* -->
         <antipattern>
-          <token postag='SENT_START|_PUNCT' postag_regexp='yes'/>
+          <token postag='SENT_START|_PUNCT|I|SPS00:DA.+' postag_regexp='yes'>
+              <exception regexp='yes'>como|já|mas|também</exception> <!-- Used specific words because POSes remove valid entries -->
+          </token>
           <token postag='DP1.+' postag_regexp='yes'/>
-          <token regexp='yes'>car[ao]s?|estimad[ao]s?|excelentes?|excelentíssim[ao]s?|querid[ao]s?</token>
+          <token regexp='yes'>adoráv(el|eis)|amantes?|amig[ao]s?|car[ao]s?|clientes?|colaborador(es)?|colegas?|companheir[ao]s?|empregad[ao]s?|estimad[ao]s?|excelentes?|excelentíssim[ao]s?|filh[ao]s?|funcionári[ao]s?|inimig[ao]s?|irmã(s|o|os)?|mães?|médic[ao]s?|pais?|prim[ao]s?|professor(res)?|querid[ao]s?</token>
```
p-goulart, line 1431: **[nitpick 🥜]** to avoid this type of thing, why not just a list of words with a POS tag filter for nouns/adjectives, and `inflected="yes"`?

p-goulart (changes requested):
This PR includes disparate changes to unrelated rules, IMO it should be split into multiple PRs.

---

We've finished our (Git documentation)[https://dev.languagetool.org/pt/git] for Portuguese-language contributors, do have a look, it might help guide you.

marcoagpinto:
@p-goulart 

Sure, thanks for the feedback.

I will fix everything at night, and tomorrow morning you will see the new code.

p-goulart:
> I will fix everything at night, and tomorrow morning you will see the new code.

Sure, there's no rush. Just try to keep unrelated changes on different branches (and therefore PRs), because we could already have merged a couple of your changes! 😢

marcoagpinto:
> Sure, there's no rush. Just try to keep unrelated changes on different branches (and therefore PRs), because we could already have merged a couple of your changes! 😢

Ahhhh… sorry… I thought I should only do that for rules and for APs add all in the same branch.

We are always learning, I will do better.

p-goulart:
Aha, I understand your confusion. No, it's generally a better idea to keep everything separate. No worries, though, these APs should be easy enough to fix.

marcoagpinto:
@p-goulart 

Heya,

I believe I have resolved the issues.

Please have a look when you have the time.

Regarding the `“as unhas não pude arranjar”` I simply deleted the AP as I am too stressed to rethink the logic of it.

I simply made the changes you requested and did a `TESTRULES pt` and `TESTRULES pt-PT` to check if they broke anything.

From now on, I will create a branch for each rule and for each AP, this will make reviewing easier.

Thanks!

p-goulart:
> Regarding the `“as unhas não pude arranjar”` I simply deleted the AP as I am too stressed to rethink the logic of it.

My larger point here is that it is the 1SG inflection of the verb that makes that construction allowable. The fronting of the direct object is only incidental to that construction being incorrect. You can delete that rule, but we plan to do some larger work with agreement rules in the near future and I am pretty sure that we will greatly reduce the number of instances of 1SG verbs tagged.

---

> From now on, I will create a branch for each rule and for each AP, this will make reviewing easier.

I think you can keep antipatterns of the same rule within the same PR, no need to split them. Unless, of course, they are each massive and there's a high probability one of them will need more work.

p-goulart (approved):

PR #9379, opened 2023-09-25
Spelling 25.9 

Luisa-LT:


languagetool-language-modules/de/src/main/java/org/languagetool/tokenizers/de/GermanCompoundTokenizer.java
```diff
@@ -73,8 +73,32 @@ static Set<String> extendedList() {
       // Note: adding words, especially short ones, can also cause incorrect splits. E.g. if "sport"
       // is in the list and you add "tran", without "transport" being in the list, it would split "transport".
       words.add("synonym");
-      words.add("aufbringung"); // Aufbringungsfläche
+      words.add("aufbringung");
       words.add("robustheit");
+      words.add("nachuntersuchung");
+      words.add("diskussion");
+      words.add("digitalisierung");
```
danielnaber, line 80: At least these two work for me without being added here, so there seems to be an issue with either your or my setup. Let's discuss tomorrow.

PR #9392, opened 2023-09-26
[pt] Added rule ID:PARA_SER_MELHORADA_A_MELHORAR

marcoagpinto:
Heya, Pedro and Susana,

Here is a simple rule which I came up with while revising my thesis.

I tried to replace:
`<token regexp='yes'>a|para</token>`
with:
`<token/>`
just to see the difference, but it only had two hits and only one was a real positive.

Here are the results:
```
Portuguese (Portugal): 1 total matches
```

[_2.txt](https://github.com/languagetool-org/languagetool/files/12732193/_2.txt)

Thanks!


p-goulart:
The rule seems fine, but can you move it to `pt-PT`? These `a + INF` constructions tend to sound a little awkward in `pt-BR`.

marcoagpinto:
@p-goulart 

Done!

😛😛😛😛😛😛😛

PR #9398, opened 2023-09-27
[de] add surnames to de_DE.dic

affemitkaraffe:
- added surnames to de_DE.dic
- moved street names, region names and last names to ignore.txt
- small fixes (missing hyphen, missing spaces, underscores instead of hyphens...)


tiff:
Is it possible to exclude them from being suggested for spelling mistakes? Similar to what ignore.txt does?

SteVio89 (approved):

Issue #3303, opened 2020-07-20
[jdk11] problem with dependencies


firm1:
Hello,

I have an dependency problem with your packages.

My gradle script look like : 

```
dependencies {
...
compile group: 'org.languagetool', name: 'language-fr', version: '5.0'
}
```

When i run my code with this, i have some problem like

```
error: the unnamed module reads package org.languagetool.language from both language.fr and languagetool.core
```

Can you have an example of java project wich use `jdk >= 9` with maven or gradle ?

chbriem:
The code seems not to be compatible with Java 9+, because of split packages.

I'd suggest to at least add a hint to the documentation. The sentence

> LanguageTool requires Java 8 or later.

is clearly wrong and should be "LanguageTool works with Java 8 only. Later versions are not yet fully supported."

miurahr:
For SEO, I'd like to leave a precise comment to the issue.

More clearly said, LanguageTool works with Java 8 and later WITHOUT Java Platform Module System (JPMS) support.

All the project with configured with JPMS will get `java.lang.module.ResolutionException` error when booting your application, then detect package duplication  and  show a cause that LT languages are share the same package that is bad habit in modern Java. 

It may be a possible solution for the application with JPMS  that  you may want to "unsplit" all the LanguageTool's language jars into a single custom module.

Because LanguageTool project don't take any action to solve the issue as in Sept. 2023, It is not easy to use LanguageTool in a project with JPMS enabled.

miurahr:
Similar issue is reported in languagetool forum. https://forum.languagetool.org/t/java-11-dependencies-requirements/6155

LavishKothari:
Do we have any timelines around when this issue will be fixed?

Given https://github.com/languagetool-org/languagetool/pull/9406 is a breaking change, and if we don't want to go ahead with this, can Languagetool export a fat jar (languagetool-core and language-en) so that there is a workaround for users for the interim time.

miurahr:
> Do we have any timelines around when this issue will be fixed?
> 
> Given #9406 is a breaking change, and if we don't want to go ahead with this, can Languagetool export a fat jar (languagetool-core and language-en) so that there is a workaround for users for the interim time.

#9406 provide a compatibility module that keep  users who implement JAVA8 base client to used as same as before.
It is for user who don't want to keep a legacy code.
[...]

PR #9406, opened 2023-09-29
WIP: Support Java9 Platform Module System(JPMS)

miurahr:
# Motivation

`languagetool` does not support Java Platform Module System (JPMS).
The project only not support it,  also break downstream applications/libraries to migrate Java 11, 17 and 21 and adopt JPMS .

It is because `languagetool` share same package among artifacts jar files, such as `o.l.language` package.
It is  well known as `split package` problem.

# What issue is fixed

[jdk11] problem with dependencies
https://github.com/languagetool-org/languagetool/issues/3303

# What changes

- escalate Contributor(s), TokenExpressionFactory public
- move modules classes
   - o.l.launguage to o.l.l.(lang)
   - o.l.chunking to o.l.c.(lang)
   - o.l.synthesis to o.l.s.(lang)
- Add Automatic-Module-Name manifest entry
- Update imports to adjust for the above changes

# Reference

- https://msayag.github.io/SplitPackage/
- https://github.com/javaee/jaxb-v2/issues/1168
- https://www.reddit.com/r/java/comments/9yexog/jdk_11_migration_how_to_deal_with_split_packages/
- 

# Status

work in progress


miurahr:
Does LanguageTool project want to provide a Java8 compatibility option for users who stick in java8 world,  and don't want to go forward to future?

miurahr:
> Does LanguageTool project want to provide a Java8 compatibility option for users who stick in java8 world, and don't want to go forward to future?

I'd like to raise it as another PR because of resolving a dependency issue with #9406 changes. After release it as SNAPSHOT version, the java8 compat module can be introduced and tested.

miurahr:
We need to change visibility of these FOUR `protected` methods of `Languages` abstract class to be `public` in order to implement compatibility classes of `org.languagetool.language.(LANGUAGE)` using "PROXY PATTERN" of GoF design pattern.

```java
  public LanguageModel initLanguageModel(File indexDir, LanguageModel languageModel);
  public SpellingCheckRule createDefaultSpellingRule(ResourceBundle messages) throws IOException;
  public int getPriorityForId(String id);
  public synchronized List<AbstractPatternRule> getPatternRules() throws IOException;
 ```

When you don't want to change `o.l.Language` class visibility, I 'd like to introduce intermediate class `o.l.compat.LanguageCompat` class that extends `Language` and change visibility and then all `language-*` modules extends new `LanguageCompat` class.

When the change for `Language` class methods visibility, I'd like just to do it? 

Any preference?

danielnaber:
> When the change for Language class methods visibility, I'd like just to do it?

Yes, I see no issue with that change.

miurahr:
Partial provide of compatible classes, now work in progress.

miurahr:
I've updated `LanguageSpecificTest` class to respect `onlyRunCode` on Disambiguation Rule Tests.

see https://github.com/languagetool-org/languagetool/pull/9406/commits/b50d10dad0ca628d4f99d51b55d42f9830fc5c89

It is why `java8compat` module test environment has all known language modules, and current test run all the languages loaded.  It lead test executions too long duration, because of redundant cases.

miurahr:
## Status

Ready for review and merge.

## Explanation of compatibility module

I've introduced project in `languagetool-language-modules/language-java8compat` that provide `org.languagetool.language.(LANGUAGE)` classes.
When user specified the module `language-java8comat` that become `split package` status which is incompatible with JPMS. It is why I give a name `java8compat` means as `java9 incompatible`.

### Core class of the compatibility module

`org.languagetool.language.AbstractLanguageProxy` class implements "Proxy pattern" in GoF design pattern.
All the individual language classes are inherited from it, which only have a default constructor with new language class path as argument.

### Tests

Most of compatibility language classes are tested with simple test case.

miurahr:
The change here moves all the language file package, so easily conflict other changes.
It is not pragmatic to track all the changes in master branch.

miurahr:
rebased on current master.

miurahr:
Now I have an idea that I want to remove the compatibility module and integrate all the compatibility class into core package. With this change, all the current users will be working without any changes.

LavishKothari:
Are there any updates on this PR? I see this change will be super useful to me.

VISTALL:
Same - I have project with java modulesm and I need the fork, because without it libraries throw error

miurahr:
I have proposed a way to move forward.
I won’t proceed unless the maintainer decides to go ahead with the move.

PR #9408, opened 2023-09-29
[pt] antipattern added to DNA[2] (#6083)

susanaboatto:
- Trying to address [this error](https://github.com/languagetooler-gmbh/languagetool-premium/issues/6083).
- It appears the disambiguator is generating non-existent postags (`D.+`, `R.+`) due to an overlapping of postags in the rule pattern:

```
<pattern>
        <marker>
          <unify>
            <feature id="gender"/>
            <feature id="number"/>
          <token postag="D.+" postag_regexp="yes"/>
        <unify-ignore>
          <token postag="R.+" postag_regexp="yes"/>
        </unify-ignore>
          <token postag="A.+" postag_regexp="yes"><exception>bem</exception></token>
          </unify>
        </marker>
      </pattern>
```

image

- As we can see, `mesmo` and `conforme` appear twice in the disambiguator because they include more than one postag from the pattern; the antipattern added tries to fix this.

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
@@ -1368,6 +1368,11 @@
       <example type="untouched">Um <marker>mal trabalhador</marker> reclama de suas ferramentas.</example><!-- triggers MAL_MAU_CONFUSION -->
     </rule>
     <rule>
+      <antipattern>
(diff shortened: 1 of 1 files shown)


p-goulart (approved):
Looks well-formed, but idk what we're trying to accomplish with this :p

susanaboatto:
From what I could gather, the error comes from the wrong postags coming from the disambiguator. My hope is that the antipattern will remove `R.+` and `D.+` from the postags, but I'm not sure it will work, tbh. Shall we deploy this and see what happens? 🤔

PR #9412, opened 2023-09-29
[de] add infix_s/no_infix_s to GERMAN_SPELLER

AgnesKleinhans:
- tokens added to infix_s showed a consistently high frequency of requiring an infix-s (tested on Jan Schreiber's word list)


languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2614,7 +2644,7 @@ private boolean isNeedingFugenS(String word) {
   private String getWordAfterEnumerationOrNull(List<String> words, int idx) {
     for (int i = idx; i < words.size(); i++) {
       String word = words.get(i);
-      if (!(word.endsWith("-") || StringUtils.equalsAny(word, ",", "/", "&", "und", "oder", "bzw.", "beziehungsweise", "sowie", "statt") || word.trim().isEmpty())) {
+      if (!(word.endsWith("-") || StringUtils.equalsAny(word, ",", "/", "&", "und", "oder", "bzw.", "beziehungsweise", "sowie") || word.trim().isEmpty())) {
```
danielnaber, line 2647: Could you sync your branch? I think this (and maybe the changes to the `s.matches` statement) shouldn't be part of the diff. I'm not sure why the CI fails, BTW.

danielnaber: I see why it fails, the paths need to look like this (i.e. without the org/languagetool etc, it will be added automatically):

```
    loadFile("de/words_infix_s.txt", wordsNeedingInfixS);
    loadFile("de/words_no_infix_s.txt", wordsWithoutInfixS);
```

AgnesKleinhans: Fixed it! Thanks! :-)

PR #9413, opened 2023-09-30
[pt] Added AP to rule ID:GENERAL_NUMBER_AGREEMENT_ERRORS

marcoagpinto:
An antipattern to fix false positives.

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -3990,7 +3990,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 
             <antipattern>
                 <token postag='RM' postag_regexp='no'/>
-                <token postag='DA.+' postag_regexp='yes'/>
+                <token>a</token> <!-- Disambiguator needs to be fixed. It has DA.+ instead of SPS00 -->
                 <token postag='AQ..P.+|NC.P.+' postag_regexp='yes'/>
                 <token postag='SPS00|SPS00:DA..[SN].+' postag_regexp='yes'/>
                 <token postag='AQ..S.+|NC.S.+' postag_regexp='yes'/>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -3988,6 +3988,15 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
            MARCOAGPINTO inserted global antipatterns for all subrules on top using the logic (2-MAR-2023+)
       -->
 
+            <antipattern>
+                <token postag='RM' postag_regexp='no'/>
+                <token postag='DA.+' postag_regexp='yes'/>
```
p-goulart, line 3993: In your example, the `a` is a preposition, not an article.

marcoagpinto: @p-goulart 

This is what I got:
Screenshot 2023-09-30 at 13-31-22 Análise de Texto - LanguageTool

p-goulart: I'm not sayingn it's tagged as such, I am saying that's what it is. Using the `DA` tag to refer to a prepositional 'a' is dangerous:
- if the disambiguator ever gets fixed, this AP will stop working;
- it accepts 'as', which is legal.

This structure works **because** 'a' is a preposition. If it's not tagged correctly, I suggest adding a literal `a` to the antipattern. Or fixing the disambiguator :p

marcoagpinto: Pedro, sorry, I was sleeping… 😋 I will fix it… just give me a few minutes.

ahhhh… I can't modify the disambiguator because I never learned how to, so I will just replace the `DA.+` with an `a`.

marcoagpinto: @p-goulart 

It is fixed!

😋 😛 ❤️ 🤗

p-goulart (approved):

PR #9418, opened 2023-10-02
[fr] PAS_DE_VIRGULE rework

GillouLT:
To fix: https://github.com/languagetooler-gmbh/languagetool-premium/issues/6321

- Some subrules from PAS_DE_VIRGULE have been moved to style as they were recommendations.
- Deleted "off" rules from the rulegroup 
- Deleted subrules 14 and 47 because they had a big gap of disables compared to the other subrules

Feedback board issue:
- Subrule 47 deleted
- Subrule 53 will now be a style recommendation (even though this won't fix the loop issue)

### DIFF
--- a/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
+++ b/languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
@@ -21187,4 +21187,552 @@ Pour en savoir plus sur les tone tags, suivez le lien vers notre documentation (
             <example correction="ainsi|alors|en conséquence|de ce fait">Le vin blanc est <marker>donc</marker> idéal pour accompagner les viandes blanches.</example>
         </rule>
     </category>
+    <category id="PONCTUATION_STYLE_OS" name="Ponctuation style OS">
+        <rulegroup id="PAS_DE_VIRGULE_STYLE" name="Pas de virgule style">
+            <rule tags="picky">
+                <antipattern>
+                    <token postag="[NJVA].*" postag_regexp="yes"/>
+                    <token>,</token>
+                    <token regexp="yes">et|ou|&amp;</token>
+                    <token inflected="yes" min="0" max="1">ne</token>
+                    <token postag="R pers obj.*" postag_regexp="yes" min="0" max="3"/>
+                    <token postag="V.*" postag_regexp="yes"/>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
```diff
@@ -21187,4 +21187,552 @@ Pour en savoir plus sur les tone tags, suivez le lien vers notre documentation (
             <example correction="ainsi|alors|en conséquence|de ce fait">Le vin blanc est <marker>donc</marker> idéal pour accompagner les viandes blanches.</example>
         </rule>
     </category>
+    <category id="PONCTUATION_STYLE_OS" name="Ponctuation style OS">
```
LucieSteib, line 21190: praise: that's a good moove 🥜

LucieSteib (approved):
LGTM! 🥜 
Any idea how to fix the loop? 🤔

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/style.xml
GillouLT: We can get rid of the loop by deleting the second suggestion (which doesn't really make sense anyway, or I'm just not seeing something):

```
            <rule tags="picky">
                <pattern>
                    <token postag="V.* (ind|con|sub).*" postag_regexp="yes">
                        <exception postag="[JN].*" postag_regexp="yes"/>
                    </token>
                    <marker>
                        <token>,</token>
                        <token postag="A">
                            <exception regexp="yes">en|par|non|outre|hier|sinon|oui|à|etc|ex|même|d'|sans|là|ici|ne|n'|voire|bien|soit</exception></token>
                        <token>,</token>
                    </marker>
                </pattern>
                <message>La virgule peut être omise.</message>
                <suggestion> \3</suggestion>
                <suggestion> \3,</suggestion>
                <example correction=" notamment| notamment,">Il ne vient<marker>, notamment,</marker> le mardi.</example>
            </rule>
```

The idea would be to delete: <suggestion> \3,</suggestion>

Can you see any issue with this deletion @LucieSteib?

LucieSteib: OOoh, yeah, that will do the trick, yes! 
Worst case we have to do something after tonight's diff.

PR #9419, opened 2023-10-02
[fr] PAS_DE_VIRGULE rework 2

GillouLT:
Fxing the loop from: https://github.com/languagetooler-gmbh/languagetool-premium/issues/6321

The second suggestion of the rule causing the loop has been deleted, therefore fixing the loop issue.


LucieSteib (approved):
LGTM! 🥜 🥜

PR #9421, opened 2023-10-02
[en] test for misspelled hyphenated words

evan-defran-lt:
TODO:
- need to check locally generated diff with `SentenceSourceChecker`


languagetool-core/src/main/java/org/languagetool/rules/spelling/morfologik/MorfologikSpellerRule.java
```diff
@@ -586,6 +590,35 @@ private List<SuggestedReplacement> calcSpellerSuggestions(String word, boolean f
 
     return Lists.newArrayList(Iterables.concat(userSuggestions, defaultSuggestions));
   }
+    private void addHyphenSuggestions(String[] parts, List<SuggestedReplacement> topSuggestions) throws IOException {
+      int i = 0;
+      for (String part : parts) {
+          if (isMisspelled(part)) {
+              List<String> partSuggestions = speller1.getSuggestions(part);
```
danielnaber, line 597: Please re-indent with 2 spaces here for coherency.

danielnaber:
@evan-defran-lt Could you revert this for now and see if you can get it to the English speller only? (But revert first)

PR #9428, opened 2023-10-04
[en] add rule SALT_TO_INJURY

St-ac-y:
please check :-) @evan-defran-lt @AzadehSafakish


evan-defran-lt:
Nice rule! 
Also could be a case of mixed metaphors: "rub salt in the wound" is fairly close in meaning.

PR #9430, opened 2023-10-04
[pt] add tags to "abraçadinho" and its variants

susanaboatto:
The diminutive forms of "abraçado" were not tagged.  Plus: the only way to mark them as diminutives is to tag them as nouns. I wonder if we should add D to adjective attributes.


p-goulart (approved):
We should prob. just have a way of handling the productive diminutive and augmentative derivations of all nouns and adjectives, yes.

PR #9436, opened 2023-10-04
[de] improve GermanSpellerRule java + add GermanSpellerRuleTest Cases

AgnesKleinhans:
@danielnaber GermanSpellerRuleTests pass! 💪

Edit: CircleCi fails on testGetAdditionalTopSuggestions


danielnaber:
> assertFirstSuggestion("Rechtsschreibungsfehlern", "Rechtschreibfehlern", rule, lt);

I think this could be fixed by adding `Rechtschreibung` to `ExtendedGermanWordSplitter.extendedList()` as it will then not be split anymore. Not sure if that's a good fix, though.

PR #9437, opened 2023-10-05
[de] multitoken spelling

jaumeortola:


AgnesKleinhans:
I have tested the implementation with a small corpus, I added a few test expressions to multitoken-expressions.txt in order to check if everything works as expected, and compared the results with the output in main.
✅ functions perfectly!

PR #9447, opened 2023-10-07
feat: allow loading language-* module as plugins in application

miurahr:
There are many language modules LanguageTool project provided. It is benefitical that langaugetool allow language-* module to be loaded dynimically before calling languagetool core functions.

Currently language list is built in static **initialization** context. 

This is a proposal that the creation of languages list is happened **when requested.** This allows applications to load language-* modules after a launch of the application, aka. classLoader load them after static initialization.

Here is a client example which loading language-en in dynamic way, and implements `brokers`. 
https://github.com/miurahr/languagetool-dynamic-loading-client-example

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/Languages.java
+++ b/languagetool-core/src/main/java/org/languagetool/Languages.java
@@ -116,9 +116,8 @@ public final class Languages {
     List<Language> staticLanguages = new ArrayList<>();
     Set<String> languageClassNames = new HashSet<>();
     try {
-      Enumeration<URL> propertyFiles = Language.class.getClassLoader().getResources(PROPERTIES_PATH);
-      while (propertyFiles.hasMoreElements()) {
-        URL url = propertyFiles.nextElement();
+      List<URL> propertyFiles = JLanguageTool.getDataBroker().getAsURLs(PROPERTIES_PATH);
+      for (URL url: propertyFiles) {
         try (InputStream inputStream = url.openStream()) {
           // We want to be able to read properties file with duplicate key, as produced by
           // Maven when merging files:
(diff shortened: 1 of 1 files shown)


miurahr:
An enhancement is requested in application,  OmegaT  project https://sourceforge.net/p/omegat/feature-requests/1714/

languagetool-core/src/main/java/org/languagetool/Languages.java
```diff
@@ -283,7 +285,7 @@ public static Language getLanguageForLocale(Locale locale) {
         return firstFallbackLanguage;
       }
     }
-    for (Language aLanguage : languages) {
```
miurahr, line 286: Why `getLanguageForLocale(Locale locale) ` exclude dynamic language module?

languagetool-core/src/main/java/org/languagetool/Languages.java
```diff
@@ -42,11 +42,8 @@ public final class Languages {
   private static final String PROPERTIES_KEY = "languageClasses";
   private static final Language NOOP_LANGUAGE = new NoopLanguage();
 
-  private static final List<Language> languages = getAllLanguages();
-  private static final List<Language> dynLanguages = new ArrayList<>();
```
miurahr, line 46: `dynLanguages` updated but never read.

languagetool-core/src/main/java/org/languagetool/Languages.java
miurahr: LT assumes en-US is static language module. Now the change allow dynamic loading of even en-US for fallback.

miurahr:
Updates in July 2024

1. Rebased on master
2. Add a commit to fix the problem Languages does not find a language module when custom class loader is used.
3. Share a client example code how to implement class broker and data broker.

miurahr:
I found a way to loading languages in dynamic way without patching languagetools-core.
see example project https://github.com/miurahr/languagetool-dynamic-loading-client-example

PR #9450, opened 2023-10-09
[en] enabled rules

evan-defran-lt:
Simply enabling rules, not sure if this warrants a PR.


AzadehSafakish:
>Simply enabling rules, not sure if this warrants a PR.

As the PiP for Git and GitHub practices, I'm going to create issues for all teams to document their language-specific PR creation/review standards.
I think we should discuss what we want for English at our standup, and then the documentation can be an example for other teams :)

AzadehSafakish (approved):

PR #9451, opened 2023-10-09
[fr] ETRE_VPPA_OU_ADJ_fix

GillouLT:
To address: https://github.com/languagetooler-gmbh/languagetool-premium/issues/6459


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -82270,7 +82292,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
                             <exception postag="V.* ppa [fe] .*|J [fe] .*|V ppa m sp" postag_regexp="yes"/></token>
                     </marker>
                 </pattern>
-                <message>Le mot masculin « \7 » n’est pas accordé en genre avec le mot féminin « \2 ».</message>
+                <message>Vérifiez l'accord en genre et en nombre de « \7 », le féminin semble plus probable.</message>
```
GillouLT, line 82295: « \2 » is in min="0", so in some sentences we had an empty « »

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -82250,6 +82262,16 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
                     <token min="0" max="4" postag="A"/>
                     <token postag="(V ppa|J) m .*" postag_regexp="yes"/>
                 </antipattern>
+                <antipattern>
+                    <token postag="UNKNOWN"/>
```
GillouLT, line 82266: The only difference from the pattern is the UNKNOWN tag, with this change, we will prevent the rule from triggering when the gender of the token is also unknown.

LucieSteib (approved):
LGTM! 🥜

PR #9452, opened 2023-10-09
[fr] LEUR_LES_V removal

LucieSteib:
Fix https://github.com/languagetooler-gmbh/languagetool-premium/issues/6486
- LEUR_LES_V demonstrate largely more FPs than TPs (see diff or Matomo disable).
- The Grafana scores aren't good.
- These situations will be handle in other ways, like targeted data for the FR gGEC model.


GillouLT:
LGTM! :)

PR #9455, opened 2023-10-10
[de] improve GermanSpellerRule.java

AgnesKleinhans:
- Compounds that can be split into three parts trigger an spelling error if the second part can't be alone, e.g. `Weinkühlschrank` as according to the current logic we check if `Weinkühl` (false) and `Kühlschrank` (true) are valid words.
- `germanPrefix.txt` is a list that comes from the jwordsplitter repo.


languagetool-language-modules/de/src/test/java/org/languagetool/rules/de/GermanSpellerRuleTest.java
```diff
@@ -540,7 +541,7 @@ public void testGetAdditionalTopSuggestions() throws Exception {
     assertFirstSuggestion("Überstreitung", "Überschreitung", rule, lt);
     assertFirstSuggestion("werkzeug.", "Werkzeug", rule, lt);
     assertFirstSuggestion("Wärkzeug.", "Werkzeug", rule, lt);
-    assertFirstSuggestion("Fußgängerunterweg", "Fußgängerunterführung", rule, lt);
+    //assertFirstSuggestion("Fußgängerunterweg", "Fußgängerunterführung", rule, lt);
```
danielnaber, line 544: This is very far away, why did this work before? Anyway, it has only 9 matches with Google and we can probably live with it not working well anymore,

languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/german_prefix.txt
```diff
@@ -0,0 +1,927 @@
+sach
```
danielnaber, line 1: As jwordsplitter is a dependency, this should be available in class path, so we can probably just load it from there without duplicating the file.

languagetool-language-modules/de/src/test/java/org/languagetool/rules/de/GermanSpellerRuleTest.java
```diff
@@ -60,6 +60,7 @@ public class GermanSpellerRuleTest {
   @Test
   public void testIgnoreMisspelledWord() throws IOException {
     GermanSpellerRule rule = new GermanSpellerRule(TestTools.getMessages("de"), GERMAN_DE);
+    assertTrue(rule.ignorePotentiallyMisspelledWord("Weinkühlschrank"));
```
danielnaber, line 63: Could you also add negative examples like "Weinskühlschrank", "Weinkühlsschrank" and maybe more to make sure these don't get accepted now?

danielnaber (changes requested):

languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/german_prefix.txt
AgnesKleinhans: Could you share a hint on how to do that? I noticed that the `germanPrefixes.txt` has been excluded in the jwordsplitter's `pom.xml`, see [here](https://github.com/danielnaber/jwordsplitter/blob/77d8a87f0ed4f07e0d58e70943b9a39b84bf047d/pom.xml#L125).

danielnaber: > I noticed that the germanPrefixes.txt has been excluded in the jwordsplitter's pom.xml, see [here](https://github.com/danielnaber/jwordsplitter/blob/77d8a87f0ed4f07e0d58e70943b9a39b84bf047d/pom.xml#L125).

Oh, I didn't remember that. In that case, my idea won't work. Could you add a comment (`# ...`) at the top of the file that explains where it comes from? Please let me know if the PR is ready.

AgnesKleinhans: I added a comment at the the top of germanprefixes.txt! PR is ready for review! Thanks 💪

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -94,6 +94,7 @@ public class GermanSpellerRule extends CompoundAwareHunspellRule {
   private final Set<String> wordEndingsToBeProhibited   = new HashSet<>();
   private final Set<String> wordsNeedingInfixS          = new HashSet<>();
   private final Set<String> wordsWithoutInfixS          = new HashSet<>();
+  private final Set<String> germanPrefix         = new HashSet<>();
```
danielnaber, line 97: Please call this germanPrefix**es** and indent it the same way as the line above

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2178,11 +2180,17 @@ protected boolean ignorePotentiallyMisspelledWord(String word) throws IOExceptio
       String compound1 = parts.get(0) + parts.get(1);
       String compound1noS = compound1.replaceFirst("s$", "");
       String compound2 = uppercaseFirstChar(parts.get(1)) + parts.get(2);
-      boolean compound1ok =
-        !isMisspelled(compound1) || ignorePotentiallyMisspelledWord(compound1) ||
-        !isMisspelled(compound1noS) || ignorePotentiallyMisspelledWord(compound1noS);
+      boolean compound1ok = false;
+      if (germanPrefix.contains(part2)){
```
danielnaber, line 2184: There's a space missing before `{`

danielnaber (changes requested):

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
AgnesKleinhans: fixed!

PR #9456, opened 2023-10-10
[fr] IMP_PRON AP added

GillouLT:
In the context of: https://github.com/languagetooler-gmbh/languagetool-premium/issues/6459

IMP_PRON FPs fixed with APs


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -91780,6 +91780,28 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+                    <example>Découvrez en quoi consiste notre méthode.</example>
+                </antipattern>
+                <antipattern>
+                    <token postag="V.* imp .*" postag_regexp="yes"/>
+                    <token>en</token>
+                    <token>quoi</token>
+                    <token postag="D.*" postag_regexp="yes"/>
+                    <token postag="[NZ].*" postag_regexp="yes"/>
+                    <example>Découvrez en quoi ce produit peut vous aider.</example>
+                </antipattern>
```
LucieSteib, line 91804: praise: very good AP! The risk of loosing some good cases is low (let's see that in the diff, I can think of some like "Nous avons beaucoup de glaces sur notre carte, découvrez-en quelques parfums de notre séléction" but still, probably not frequent, the diff for [11] doesn't contain any)

LucieSteib (approved):
LGTM!

PR #9461, opened 2023-10-11
[pt] Added verb "tomar" to disambiguation.xml

marcoagpinto:
Hello Pedro and Susana,

Is this it? Two or three lines of XML?

It was supposed to be hard?

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
@@ -351,6 +351,13 @@
     </pattern>
     <disambig action="remove" postag="NC.*"/>
   </rule>
+  <rule id="TOMAR" name="tomar (not noun)">
+    <pattern>
+      <!-- TODO: exceptions -->
+      <token regexp="yes">tomas?</token>
+    </pattern>
+    <disambig action="remove" postag="NC.*"/>
+  </rule>
   <rule id="QUANTA" name="quanta (not noun)">
     <pattern>
       <!-- TODO: exceptions in physics: um quanta -->
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
@@ -354,7 +354,9 @@
   <rule id="TOMAR" name="tomar (not noun)">
     <pattern>
       <!-- TODO: exceptions -->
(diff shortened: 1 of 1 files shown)


p-goulart:
Hm, well, this *works*, in a sense, but we don't want to remove the noun tag from *every* instance of `tomas?`, do we? I mean, it *is* a noun occasionally, albeit a rare one.

Instead, I think it'd work better if you added an antipattern to the disambiguator rules responsible for mistagging things.

From the original issue:
```
DAN[1]: crime[crime/AQ0CS0,crime/NCMS000] -> crime[crime/AQ0CS0]
DET-NOUN_PRON-VERB[3]: crime[crime/AQ0CS0] -> crime[crime/AQ0CS0]
DAN[1]: toma[toma/NCCS000,tomar/VMIP3S0,tomar/VMM02S0] -> toma[toma/NCCS000]
``` 

This means the disambiguator rule `DAN` (sub-rule 1) is doing weird things, such as removing the noun tag from `crime` and then removing the verb tags from `toma`. This is what we have to stop. We can do that by adding anti-patterns specifically targeting `tomas?`, of course, as a hotfix.

marcoagpinto:
@p-goulart 

Is this it?:

```
  <rule id="TOMAR" name="tomar (not noun)">
    <pattern>
      <!-- TODO: exceptions -->
      <token regexp="yes">tomas?
        <exception scope='previous' postag_regexp='yes' negate_pos='yes' postag='NC.+|AQ.+'/>
      </token>
    </pattern>
    <disambig action="remove" postag="NC.*"/>
  </rule>
```

marcoagpinto:
@p-goulart 

Hello, Pedro!

Is this it?

I have updated the pull request.

Thanks!

p-goulart:
That's not quite what I had in mind earlier, but if this rule applies before `DAN` it should have roughly the same effect. Once it's enabled, does your style rule behave as intended?

marcoagpinto:
> That's not quite what I had in mind earlier, but if this rule applies before `DAN` it should have roughly the same effect. Once it's enabled, does your style rule behave as intended?

Yes, `TESTRULES PT` no longer complains about the correction examples, which means it is converted to a verb.

p-goulart (approved):
In that case, I suppose that's good enough! ✨ See, not so hard was it..?

marcoagpinto:
> In that case, I suppose that's good enough! ✨ See, not so hard was it..?

@p-goulart 

But, years ago, the disambiguator had to follow a certain order (alphabetically? I can't remember), if you placed the code at a different place, it would not work.

Has that behaviour been changed?

p-goulart:
I was not working at LT years ago, so it's hard for me to say ;) As far as I know, this should work.

PR #9464, opened 2023-10-12
[en] added entries to removed.txt

evan-defran-lt:
Requesting a review since I don't often make changes to the dictionary files.


AzadehSafakish (approved):
Yup, I see now that these entries are in the dump file and not `added.txt`.
Which means an LT contributor did not tag them for a specific reason, so removing them should cause minimal issues.

However, as to be noted with all tagging changes:
There is a possibility that this will cause lots of shift in the diff because any pattern/antipattern that relied on these words being tagged as verbs will no longer match.
We'll see if anything unusual happens tomorrow, but it might not affect anything at all.

PR #9491, opened 2023-10-17
[fr] ESPACE_UNITES_27_rule_creation

GillouLT:
Adjusted currency rule for proper FR formatting.

"€ 30" was already corrected to "30 €" with ESPACE_UNITES[7], but the rule didn't work in cases where there is no sapcing between the currency symbol and the number (for example €30).

This new subrule will suggest replacing "€30" with "30 €", the regexp is an adapted version of EN CURRENCY[1] rule.


LucieSteib (approved):
LVGTM 🥜 💰 💶 🥳

PR #9494, opened 2023-10-18
[fr] ESPACE_UNITES_27_temp_off_removed

GillouLT:
Temp_off removed, the rule works as expected
Also fixing an issue with ESPACE_UNITES[20] after checking the diff


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -117186,7 +117186,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
                         <token regexp="yes">\d.*\d\d\d$
                             <exception scope="previous" spacebefore="no" regexp="yes">[\-\/]</exception>
                             <exception regexp="yes">[12]\d\d\d$|\d.* \d.*|\d.*[\/]\d.*|\d.*[\.\,]\d.*|\d.*[a-z]\d.*|00\d.*|0[123456789]\d\d\d\d\d\d\d\d</exception>
-                            <exception regexp="yes">^\d{7,9}$</exception></token>
+                            <exception regexp="yes">^\d{7,30}$</exception></token>
```
LucieSteib, line 117189: question(unblocking):i'm curious, why? :D

LucieSteib (approved):
LGTM! Here's €30 worth of 🥜 😄

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
GillouLT: The rule adds spaces to numbers like for example 10000 >> 10 000

The rule works as expected for numbers up to 999 999, from 1 000 000, the rule will convert 1000000 to 10 000 00 and will mess up the spacing instead of fixing it :( 

I managed to prevent the rule from triggering in numbers with up to 7 units with {7,} (the rest in handled in premium rules), but the day after I realized that it created a warning in the maven test asking me to put a number after 7...

So I changed it to {7,9}, which was a bad decision because it means that for numbers that have at least 9 units, the rule will trigger (and produce a wrong spacing) you can see that in today's diff for ESPACE_UNITES[20], so now with {7,30} the rule will trigger in numbers with up to 7 units and for numbers with at least 30 units (hopefully it won't happen too much)

PR #9497, opened 2023-10-18
[fr] DE_DU_D rule creation

LucieSteib:
In the context of [this round of Grammalecte rules cleanup](https://github.com/languagetooler-gmbh/recurring-issues/issues/10)

- Creation of rule DE_DU_D.
- This rule is a test to replace grammalecte_g2__conf_de_du_d__b1_a1_1
- grammalecte_g2__conf_de_du_d__b1_a1_1 has a bad apply rate, despite being essentially triggered on TPs (in diff and Matomo disable). We thought that this might be du to the quite bad error message from the Grammalecte rule, juste saying "incohérence"
image


LucieSteib:
@GillouLT merging now, so we can check the results in diff tomorrow (anyway, it's temp_off)

PR #9507, opened 2023-10-20
[pt] Added exception to rule ID:ESTA_ESTÁ

marcoagpinto:
Heya,

It removes false positives.


languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -31604,10 +31604,12 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                     </marker>
                     <token postag='NC.+|AQ.+' postag_regexp='yes'>
                         <exception postag_regexp='yes' postag='AO.+'/>
+                        <exception scope='next' postag_regexp='yes' postag='V.+'/>
```
p-goulart, line 31607: A bit of a corner case, but maybe make sure this verb token **doesn't** allow for participle matches, since those are often syntactically adjectival: e.g. `esta situação complicada`.

p-goulart (changes requested):

marcoagpinto:
@p-goulart 

The fix you suggested has been implemented!

p-goulart (approved):

PR #9508, opened 2023-10-20
[de] add gGEC APs

St-ac-y:
please check :-)


languagetool-language-modules/de/src/main/resources/org/languagetool/rules/de/remote-rule-filters.xml
```diff
@@ -1774,6 +1774,28 @@ To ignore a remote rule match, set the <marker> so that it exactly covers the te
             </rule>
         </rulegroup>
 
+        <rulegroup id="AI_DE_GGEC_REPLACEMENT_OTHER" name="">
+            <rule><!--1-->
+                <pattern>
+                    <token regexp="yes">[A-Za-z]*</token>
```
AgnesKleinhans, line 1780: `[a-zäöü]` would be sufficient, as the pattern is not case-sensitive :-)

AgnesKleinhans (approved):

AgnesKleinhans (approved):

PR #9510, opened 2023-10-20
Add extended sentence range

SteVio89:
For multi-language support


languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/HunspellRule.java
```diff
@@ -259,10 +259,10 @@ public RuleMatch[] match(AnalyzedSentence sentence) throws IOException {
...
           ruleMatches.add(ruleMatch);
           if (foreignLanguageChecker != null && !gotResultsFromForeignLanguageChecker) {
-            String langCode = foreignLanguageChecker.check(ruleMatches.size());
-            if (langCode != null) {
-              if (!langCode.equals(ForeignLanguageChecker.NO_FOREIGN_LANG_DETECTED)) {
-                ruleMatches.get(0).setErrorLimitLang(langCode);
+            Map<String, Float> langCodesScoring = foreignLanguageChecker.check(ruleMatches.size());
+            if (!langCodesScoring.isEmpty()) {
+              if (langCodesScoring.get(ForeignLanguageChecker.NO_FOREIGN_LANG_DETECTED) == null) {
+                ruleMatches.get(0).setNewLanguageMatches(langCodesScoring);
```
fabrichter, line 265: Can you add a comment about why only ruleMatches.get(0) gets annotated?

languagetool-core/src/main/java/org/languagetool/rules/RuleMatch.java
```diff
@@ -602,22 +599,21 @@ public int hashCode() {
   /**
    * The language that the text might be in if the error limit has been reached.
    *
-   * @since 5.3
+   * @since 6.4
    */
-  @Nullable
-  public String getErrorLimitLang() {
-    return errorLimitLang;
+  public Map<String, Float> getNewLanguageMatches() {
```
fabrichter, line 604: Should we list this as a breaking change or even keep the old method for a while, implement it based on the map and deprecate it?

languagetool-core/src/main/java/org/languagetool/tools/RuleMatchesAsJsonSerializer.java
```diff
@@ -232,6 +234,29 @@ private void writeSentenceRanges(JsonGenerator g, List<CheckResults> res) throws
     g.writeEndArray();
   }
 
+  private void writeExtendedSentenceRanges(JsonGenerator g, List<CheckResults> res) throws IOException{
+    g.writeArrayFieldStart("extendedSenteceRanges");
```
fabrichter, line 238: typo: Sentence, not Sentece

languagetool-core/src/main/java/org/languagetool/CheckResults.java
```diff
@@ -18,45 +18,49 @@
...
+    this(ruleMatches, ignoredRanges, Collections.emptyList());
   }
 
-  public List<RuleMatch> getRuleMatches() {
-    return ruleMatches;
+  public CheckResults(List<RuleMatch> ruleMatches, List<Range> ignoredRanges, List<ExtendedSentenceRange> extendedSentenceRanges) {
+    this.ruleMatches = Objects.requireNonNull(ruleMatches);
+    this.ignoredRanges = Objects.requireNonNull(ignoredRanges);
+    this.extendedSentenceRanges = Objects.requireNonNull(extendedSentenceRanges);
+    //TODO: use this later
```
fabrichter, line 49: what about this TODO? can you add more information, implement it now or delete it?

languagetool-core/src/main/java/org/languagetool/CheckResults.java
```diff
@@ -18,45 +18,49 @@
...
+
   public void addSentenceRanges(List<SentenceRange> sentenceRanges) {
     this.sentenceRanges.addAll(sentenceRanges);
   }
-  
-  public void setIgnoredRanges(List<Range> ignoredRanges) {
-    this.ignoredRanges = Objects.requireNonNull(ignoredRanges);
+
+  public List<ExtendedSentenceRange> getExtendedSentenceRanges() {
+    return extendedSentenceRanges.stream().sorted().collect(Collectors.toList());
```
fabrichter, line 63: Can we store this in sorted form and then just return the list instead of possibly sorting multiple times?

fabrichter (changes requested):
Looks mostly good, added some minor fixes/questions.
Can you also add a comment somewhere that shortly summarizes the general approach for multi-language handling that this is meant to support and give a brief overview of the steps (both inside the LanguageTool server itself as well as external)?

languagetool-core/src/main/java/org/languagetool/rules/RuleMatch.java
SteVio89: I can't imagine it being used anywhere other than directly in our code.

PR #9532, opened 2023-10-25
[en] world capitals

evan-defran-lt:
requesting review because I'm not sure if some of these should rather go in `multiwords.txt`.


AzadehSafakish (changes requested):

AzadehSafakish (approved):
LGTM! 👍

PR #9540, opened 2023-10-26
[pt] Fixed postags of word "tórax"

marcoagpinto:
Heya,

This should fix the false positive of “`tórax”` in the INIMIGOS_ rule.

The other ones I will fix later.


marcoagpinto:
https://dicionario.priberam.org/t%C3%B3rax

susanaboatto (approved):

PR #9542, opened 2023-10-26
Update Healthcheck and add javadoc

SteVio89:


languagetool-server/src/main/java/org/languagetool/server/LanguageToolHttpHandler.java
```diff
@@ -123,7 +123,7 @@ public void handle(HttpExchange httpExchange) throws IOException {
         // healthcheck should come before other limit checks (requests per time etc.), to be sure it works: 
         String pathWithoutVersion = path.substring("/v2/".length());
         if (pathWithoutVersion.equals("healthcheck")) {
-          if (workQueueFull(httpExchange, parameters, "Healthcheck failed: There are currently too many parallel requests.")) {
+          if (workQueueFull(httpExchange, parameters, "Healthcheck failed: There are currently too many parallel requests.") || textCheckerV2.checkerQueueAlmostFull()) {
```
fabrichter, line 126: workQueueFull actually sends an error, the added case doesn't
refactor that part out and call it in the conditional so that we actually send a response

languagetool-core/src/main/java/org/languagetool/Language.java
```diff
@@ -536,6 +536,9 @@ public void setPostDisambiguationChunker(Chunker chunker) {
   /**
    * Create a shared instance of JLanguageTool to use in rules for further processing
    * Instances are shared by Language
+   *
+   * Do not call this in a static block or to initialize a static field in rules or filters.
```
fabrichter, line 540: Maybe mention the consequences, and also that configuring this with e.g. disableRules is also bad.
Can we recommend alternatives?

fabrichter (changes requested):

fabrichter (approved):

PR #9543, opened 2023-10-26
[nl] try to improve CheckCaseRule

mark-baas:


languagetool-core/src/main/java/org/languagetool/rules/AbstractCheckCaseRule.java
```diff
@@ -131,6 +141,14 @@ public RuleMatch[] match(AnalyzedSentence sentence) {
...
 
+  public boolean ignoreShortUppercaseWords() {
+    return ignoreShortUppercaseWords;
+  }
+
+  public void setIgnoreShortUppercaseWords(boolean value) {
+    ignoreShortUppercaseWords = value;
+  }
+
   private boolean isPunctuationStart(String word) {
```
jaumeortola, line 152: You are overwriting the method `public boolean ignoreShortUppercaseWords()`, but in the code you are using the variable `ignoreShortUppercaseWords`, which is unchanged by your method. 
Remove one of these methods, and use only the other one.

languagetool-core/src/main/java/org/languagetool/rules/AbstractCheckCaseRule.java
```diff
@@ -96,7 +97,15 @@ public RuleMatch[] match(AnalyzedSentence sentence) {
           break;
         }
         if (originalPhrase.equals(originalPhrase.toUpperCase())) {
-          continue;
+          if (ignoreShortUppercaseWords) {
+            continue;
+          } else {
+            if( originalPhrase.length() < 5 ){
```
jaumeortola, line 103: watch the formatting

mark-baas: Has been modified now

jaumeortola (approved):

PR #9546, opened 2023-10-26
feat[core]: skip reading dictionary when cached one exists

fabrichter:
load dictionary content lazily, when we don't have it already cached
also fixed formatting


danielnaber:
Would it be possible to get the whitespace changes as a separate commit, to more easily see the real change?

fabrichter:
You can use the hide whitespace setting on GitHub: https://github.com/languagetool-org/languagetool/pull/9546/files?diff=split&w=1

danielnaber:
> You can use the hide whitespace setting on GitHub

But that only helps with real whitespace changes and not with differences in line breaks, e.g.
whitespace

Anyway, Stefan is the reviewer :-)

SteVio89 (approved):

SteVio89:
Did not see the latest changes before aprove...pls wait

SteVio89 (approved):

PR #9549, opened 2023-10-27
[nl] improve CompoundAcceptor

mark-baas:


languagetool-language-modules/nl/src/main/java/org/languagetool/rules/nl/CompoundAcceptor.java
```diff
@@ -96,7 +95,7 @@ private boolean abbrevOk(String nonCompound) {
 
   private boolean spellingOk(String nonCompound) throws IOException {
     AnalyzedSentence as = new AnalyzedSentence(new AnalyzedTokenReadings[] {
-      new AnalyzedTokenReadings(new AnalyzedToken(nonCompound, "FAKE_POS", "fakeLemma"))
+      new AnalyzedTokenReadings(new AnalyzedToken(nonCompound.toLowerCase(), "FAKE_POS", "fakeLemma"))
```
danielnaber, line 98: After having merged the PR, I wonder whether this change is ok: aren't there words like country names or names in general that are always uppercase and that won't be accepted when lowercased? These are probably less likely to appear as part of a compound, but maybe you can add a test case for them anyway?

mark-baas: Yes, there are some exceptions that will still have uppercases. Like you said, less likely, but I'll think of a better solution for them as I add more words.

PR #9552, opened 2023-10-27
Various optimizations

fabrichter:


languagetool-core/src/main/java/org/languagetool/Language.java
```diff
@@ -93,6 +100,12 @@ public abstract class Language {
   private Chunker postDisambiguationChunker;
   private Synthesizer synthesizer;
 
+  private final String shortCodeWithCountryAndVariant;
+
+  protected Language() {
```
SteVio89, line 105: This is not used by -> shortCodeWithCountryAndVariant is always NULL

languagetool-core/src/main/java/org/languagetool/Language.java
```diff
@@ -631,6 +644,10 @@ public final String getTranslatedName(ResourceBundle messages) {
    * @since 3.6
    */
   public final String getShortCodeWithCountryAndVariant() {
+    return shortCodeWithCountryAndVariant;
```
SteVio89, line 647: Returns NULL

languagetool-core/src/main/java/org/languagetool/Language.java
```diff
@@ -926,7 +947,7 @@ public boolean equals(Object o) {
 
   @Override
   public int hashCode() {
-    return getShortCodeWithCountryAndVariant().hashCode();
+    return shortCodeWithCountryAndVariant.hashCode();
```
SteVio89, line 950: NullPointerException

languagetool-core/src/main/java/org/languagetool/Language.java
fabrichter: this is impossible, the field is final and not initializing it (by adding another constructor) will cause a compiler error
and in subclasses, superclass constructors are always called, either explicit or implicit

languagetool-core/src/main/java/org/languagetool/Language.java
fabrichter: see comment above

PR #9553, opened 2023-10-27
[pt] Added words to removed.txt

marcoagpinto:
Heya,

“defesas” is not a verb according to the on-line resources.


susanaboatto (approved):
Actually, it is a verb, but it comes from `defesar`, not `defender`. It appears this verb is not in the PT dictionary at all.

susanaboatto (changes requested):
It can also be defender in the [past participle](http://www.portaldalinguaportuguesa.org/index.php?action=lemma&lemma=74784), even though it's hardly used.

susanaboatto:
Yes, please

marcoagpinto:
Closed!

PR #9558, opened 2023-10-29
[pt] Added rule ID:CONFUSÃO_AO_OU

marcoagpinto:
Heya, Susana and Pedro,

Here is a new rule.

It only had one hit in 900 000 sentences, and the hit seems inconclusive but a valid structure for the rule (well triggered).

```
Portuguese (Portugal): 1 total matches
```

[6.txt](https://github.com/languagetool-org/languagetool/files/13198030/6.txt)

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -30540,7 +30540,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
         </rule>
 
 
-        <rule id='CONFUSÃO_AO_OU' name="[Confusão] ao/ou" default="temp_off">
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -30540,6 +30540,29 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
         </rule>
 
 
+        <rule id='CONFUSÃO_AO_OU' name="[Confusão] ao/ou" default="temp_off">
```
susanaboatto, line 30543: Users can see the brackets [] here. Best to remove them: `Confusão de termos: ao/ou`

susanaboatto (changes requested):

marcoagpinto:
It is now fixed too.

😛

susanaboatto (approved):

Issue #9555, opened 2023-10-28
[pt] Idea for rule: “quase que + Pronome” → “quase pronome” - 2023-10-28


marcoagpinto:
Heya, @ricardojosehlima

I have been developing rules quietly.

Now and then I ask for the help of the developers about the category of rules and doubts implementing, but I seldom receive an answer.

So, here is an idea for a rule:
`Quase QUE me arrancou a mão.`
Becomes:
`Quase me arrancou a mão.`

Is this a valid rule? Removing the “que”?

When I write “me” I could also use the other pronouns.

Thanks!

ricardojosehlima:
Seems a good idea, but in pt-br we accept "quase que" in informal texts, so the rule should be restricted here to formal texts.

marcoagpinto:
@ricardojosehlima 

Heya, thanks for letting me know about the formal.

I will work on it tonight.

❤️ ❤️ ❤️ ❤️ ❤️ ❤️

marcoagpinto:
Heya,

Sorry for the delay.

It is coded:
```
Portuguese (Portugal): 6 total matches
Portuguese (Portugal): 811110 total sentences considered
Portuguese (Portugal): ø0.00 rule matches per sentence
```
[1.txt](https://github.com/languagetool-org/languagetool/files/13246041/1.txt)

ricardojosehlima:
Looks great to me!

marcoagpinto:
@ricardojosehlima 

My dear brother,

Thank you for your help.

Tomorrow I will write another rule suggestion.

Susana has already approved some five or six pull requests today, and I don't want to overload people.

😛 😛 😛 😛 😛 😛 😛 😛 😛 😛 😛 😛

PR #9591, opened 2023-11-03
[pt] Added rule ID:QUASE_QUE_PRONOME_QUASE_PRONOME

marcoagpinto:
Heya,

Here is a formal rule.

Follow the discussion here:
https://github.com/languagetool-org/languagetool/issues/9555

Thanks!

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -3748,7 +3748,7 @@ USA
                     <token regexp='yes'>&pronomes_obliquos;</token>
                 </marker>
             </pattern>
-            <message>Linguagem informal. Substitua por:</message>
+            <message>&informal_msg;</message>
             <suggestion>\1 \3</suggestion>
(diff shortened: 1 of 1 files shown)


susanaboatto (approved):

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
```diff
@@ -3739,6 +3739,21 @@ USA
             <example correction="danificada|avariada">A minha UPS nova vinha <marker>estragada</marker>.</example>
         </rule>
 
+
+        <rule id='QUASE_QUE_PRONOME_QUASE_PRONOME' name="'quase que + pronome' → 'quase pronome'" type="style" default="temp_off">
```
susanaboatto, line 3743: Can we shorten the rule name a little and leave it as `QUASE_QUE_PRONOME` or just `QUASE_QUE`? It's not an absolute must, but we've now got some rule writing guidelines for reference, you can check them here if it interests you 😸: https://dev.languagetool.org (scroll to the bottom of the page for PT guidelines)

susanaboatto (approved):

PR #9605, opened 2023-11-06
[de] added gunsten + ungunsten (should be capital)

Luisa-LT:
Ok like this @tiff?

### DIFF
--- a/languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
+++ b/languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
@@ -115,8 +115,6 @@ public class GermanSpellerRule extends CompoundAwareHunspellRule {
     put("Kruks", "Krux");
     put("Filterbubble", "Filterblase");
     put("Filterbubbles", "Filterblasen");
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -115,6 +115,8 @@ public class GermanSpellerRule extends CompoundAwareHunspellRule {
     put("Kruks", "Krux");
     put("Filterbubble", "Filterblase");
     put("Filterbubbles", "Filterblasen");
+    put("gunsten", "Gunsten");
+    put("ungunsten", "Ungunsten");
```
tiff, line 119: This will just make "Gunsten" the first suggestion, not the only one. You will have to use the `topMatch` approach further below

tiff (changes requested):

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
Luisa-LT: Thank you! I also added this to the confluence file overview page.

PR #9628, opened 2023-11-08
[pt] Added antipattern to rule ID:GENERAL_NUMBER_AGREEMENT_ERRORS

marcoagpinto:
Hi!

An antipattern to fix false positives.


p-goulart (approved):
✨

PR #9631, opened 2023-11-08
[en] added words

evan-defran-lt:


languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/hunspell/spelling.txt
```diff
@@ -37107,4 +37107,7 @@ sclerosis
 scleroses
 referenda
 polenta
-touristed
\ No newline at end of file
+touristed
+pseudoscientific
+word2vec
```
AzadehSafakish, line 37112: My instinct would be to put 'word2vec' in ignore, but then again I can't really think of a typo where 'word2vec' would unrightfully assume the top position on the suggestion list, so it's probably okay.

AzadehSafakish (approved):

PR #9641, opened 2023-11-09
fix: regex optimizations

j-bolivar-lt:


languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2113,22 +2126,23 @@ && isMisspelled(word)
...
             !isMisspelled(firstPart + "test")) {  // does hunspell accept this? takes infex-s into account automatically
           //System.out.println("will accept: " + word);
           return true;
         } else if (!isMisspelled(firstPart) &&
-                   !firstPart.matches(".{3,25}(tum|ing|ling|heit|keit|schaft|ung|ion|tät|at|um)")) {
+                  SPECIAL_CASE.matcher(firstPart).matches() &&
+                  !firstPart.matches(".{3,25}(tum|ing|ling|heit|keit|schaft|ung|ion|tät|at|um)")) {
                    //System.out.println("will not accept: " + word);
         } else if (firstPart.endsWith("s") && !isMisspelled(firstPart.replaceFirst("s$", "")) &&
-                   firstPart.matches(".{3,25}(tum|ing|ling|heit|keit|schaft|ung|ion|tät|at|um)s") &&   // "handlungsartig"
```
danielnaber, line 2125: Note the "s" at the end - there will need to be two versions of `SPECIAL_CASE`, one with and one without "s".

PR #9647, opened 2023-11-10
[fr] AI_FR_GGEC_REPLACEMENT_ORTHOGRAPHY prio change in French.java

LucieSteib:
We saw that some matches previously handled by some XML rules are now diverted by gGEC rules.
This example `Cette année, Avril a été chaud.` is from rulegroup id="MOIS" name="mois de l’année" (src/main/resources/org/languagetool/rules/fr/grammar.xml).
It's now AI_FR_GGEC_REPLACEMENT_ORTHOGRAPHY_UPPE handling it. That's not conveniant, as the xml rules have antipatterns already (see second screen shot about Mai 68).
**This PR lowers the prio of AI_FR_GGEC_REPLACEMENT_ORTHOGRAPHY.**
capture_d___cran_2023-11-10_141046_720
image_480


GillouLT (approved):
Let's see! LGTM :)

PR #9667, opened 2023-11-14
[core] global spelling: film makers

jaumeortola:


LucieSteib (approved):
Well, nothing seems wrong to me. 
I'm just unsure all these people are film makers, but that's not important.

p-goulart (approved):
Nothing jumps out at me, other than comments I already made the first time around: Japanese long vowels (`ō` could be `o`); and Hungarian name ordering (`István Szabó` is `Szabó István` in Magyarspeak).

jaumeortola:
Thank you!

PR #9671, opened 2023-11-15
[en] update URLs

AzadehSafakish:
Adding a URL for specific `EN_DIACRITICS_REPLACE` IDs.
Same logic as `ConfusionProbabilityRule.java`, but want to double-check that it's implemented correctly.

I'll have to do the same for `EN_FOREIGN_TERM` in Premium.


jaumeortola:
Did you try it? I think it won't work. We need to add the URL to the specificIdRule:
> specificIdRule.setUrl("...");

I can implement it. But if it is for just two or three words, perhaps it is easier to write a specific XML rule.

AzadehSafakish:
>We need to add the URL to the specificIdRule

Ah, I see it now.

>I can implement it. But if it is for just two or three words, perhaps it is easier to write a specific XML rule.

Would certainly be useful now and in the future, but for these four words I'll just let Marvin know we don't support it at the moment, and I'll add the URLs to XML equivalents.

PR #9677, opened 2023-11-15
[de] added words that shouldn't be accepted by the speller

Luisa-LT:
Interesting info: Almost all are corrected by `gGEC` but since not all users get gGEC matches yet, I added them to `prohibit.txt`


AgnesKleinhans (approved):
cool!! 🚀

danielnaber:
The PR is okay, but we should fix the underlying issue here, as e.g. "Bauerbrot" doesn't seem to be a special case. "Bauerwitwe" is also accepted, for example.

danielnaber:
> Where could I add e.g. Bauer that it is not used as a compound part?

Not sure I understand the question, but currently it's in `words_no_infix_s.txt` and it just doesn't seem to belong there, as the typical compound part is `Bauern...`

Luisa-LT:
Of course, true it needs to have an -n, I'll add it there

Luisa-LT:
https://github.com/languagetool-org/languagetool/pull/9680/commits/468e042123d2d925c970adb19beaa38bbbd747f9

PR #9686, opened 2023-11-16
[de] add test rule for Swiss number format

affemitkaraffe:
@St-ac-y


affemitkaraffe:
see [here](https://github.com/languagetool-org/languagetool/commit/bb3ae02e96f828747880a6d3c4b67cf9b465e695)

PR #9687, opened 2023-11-16
[de] corrected, removed, added words (some moved to infix_s) to avoid FNs

Luisa-LT:
First part of the re-visited words listed in `words without infix s`.

Now that we have more knowledge about compound formation, I have revised the list again and removed some cases due to possible confusion, added, removed or moved to `words infix s`

@danielnaber @AgnesKleinhans could you please check if this procedure is correct/helpful?
If yes, I will continue the same way with the second part of the list.

### DIFF
--- a/languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/words_infix_s.txt
+++ b/languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/words_infix_s.txt
@@ -269,3 +269,9 @@ Imparität
 Kollektion
 Abspaltung
 Benachrichtigung
+Freundschaft
+Frieden
+Schwein
+Entscheidung
+Eintracht
+Mönch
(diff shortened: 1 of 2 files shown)


languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/words_infix_s.txt
```diff
@@ -268,4 +268,10 @@ Hibernation
 Imparität
 Kollektion
 Abspaltung
-Benachrichtigung
\ No newline at end of file
+Benachrichtigung
+Freundschaft
+Frieden
+Schwein
```
danielnaber, line 274: "Schweine..." seems to be more common, according to e.g. Jan Schreiber's list: `grep "Schweine" german.txt` -> 602 matches, vs `grep "Schweins" german.txt` -> 45 matches. With these ambiguous terms, I think we have to leave the word out for now (unless both s and e variants are always okay, which doesn't seem to be the case).

Luisa-LT: Ok I will comment it out in both

Luisa-LT: Actually, I think we should keep only _Schweine_ in `without infix s` (e.g. _Schweineohren_ seems to be correct too...)

AgnesKleinhans (approved):

PR #9700, opened 2023-11-20
[nl] improve DutchTagger

mark-baas:


languagetool-language-modules/nl/src/main/java/org/languagetool/tagging/nl/DutchTagger.java
```diff
@@ -168,14 +188,27 @@ public List<AnalyzedTokenReadings> tag(List<String> sentenceTokens) {
...
+                if (part2Reading.getPOSTag() != null && part2Reading.getPOSTag().startsWith("ZNW")) {
+                  // if word2 is contained in these lists, just give it the tag and exit the loop
+                  if (alwaysNeedsHet.contains(part2)){
+                    l.add(new AnalyzedToken(word, "ZNW:EKV:HET", part1lc + part2Reading.getLemma()));
+                    break;
+                  } else if (alwaysNeedsDe.contains(part2)){
+                    l.add(new AnalyzedToken(word, "ZNW:EKV:DE_", part1lc + part2Reading.getLemma()));
+                    break;
+                  } else if (alwaysNeedsMrv.contains(part2)){
+                    l.add(new AnalyzedToken(word, "ZNW:MRV:DE_", part1lc + part2Reading.getLemma()));
```
danielnaber, line 206: To avoid duplication, please call sth. like `l.add(new AnalyzedToken(word, tag, part1lc + part2Reading.getLemma()));` at the end if the if/elseif and in the if/elseif only set the value of `tag`.

languagetool-language-modules/nl/src/main/java/org/languagetool/tagging/nl/DutchTagger.java
```diff
@@ -67,7 +68,26 @@ public class DutchTagger extends BaseTagger {
   public DutchTagger() {
     super("/nl/dutch.dict", new Locale("nl"));
   }
-
+  private final Set<String> alwaysNeedsHet = ImmutableSet.of(
```
danielnaber, line 71: Please also add `static` here

languagetool-language-modules/nl/src/main/java/org/languagetool/tagging/nl/DutchTagger.java
```diff
@@ -168,14 +188,27 @@ public List<AnalyzedTokenReadings> tag(List<String> sentenceTokens) {
...
               AnalyzedTokenReadings part2Readings = part2ReadingsList.get(0);
               String part1lc = part1.toLowerCase();
               for (AnalyzedToken part2Reading : part2Readings) {
-                if (part2Reading.getPOSTag() != null && part2Reading.getPOSTag().contains("ZNW")) {
-                  //System.out.println("Adding " + word + " with postag " + part2Reading.getPOSTag() + ", part2 has lemma " + part2Reading.getLemma());
-                  l.add(new AnalyzedToken(word, part2Reading.getPOSTag(), part1lc + part2Reading.getLemma()));
+                if (part2Reading.getPOSTag() != null && part2Reading.getPOSTag().startsWith("ZNW")) {
+                  // if word2 is contained in these lists, just give it the tag and exit the loop
+                  if (alwaysNeedsHet.contains(part2)){
+                    l.add(new AnalyzedToken(word, "ZNW:EKV:HET", part1lc + part2Reading.getLemma()));
```
danielnaber, line 200: Please add a space between `){`

danielnaber (changes requested):
I don't yet understand why we need the data like `alwaysNeedsHet` - doesn't LT already know which words need which determiner, based on its internal dict and `added.txt`? So, isn't this a duplication?

mark-baas:
It's mostly to keep it a bit more clean. These part2 nouns have multiple meanings and therefore have different tags assigned ("de weer and het weer" are both valid). When these specific words are used in compounds, only one variation is valid, so it seems better to give them the right tag right away.

danielnaber:
Thanks. In that case, feel free to merge once my other small remarks have been addressed.

PR #9708, opened 2023-11-20
[pt] Removed rule ID:MADRUGADA_MATINAL

marcoagpinto:
Heya, @p-goulart and @susanaboatto 

I have removed this useless rule.

😛 😛 😛 😛 😛 😛

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -12810,6 +12810,25 @@ USA
         </rule>
 
 
+        <rule id='MADRUGADA_MATINAL' name="Simplificar: da madrugada/manhã → matinal" type="style" default='temp_off'>
+            <!--IDEA shorten_it-->
+            <pattern>
+                <token postag='V.+' postag_regexp='yes'/>
+                <token regexp='yes'>[ao]s?</token>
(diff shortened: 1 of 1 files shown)


p-goulart:
Tbh I like this as a style rule... it does confer the text a more polished feel to say 'a neblina matinal' rather than 'a neblina da manhã'. The one exception would be café, at least in Brazil, since there 'café da manhã' is lexicalised as 'breakfast'. Personally I wouldn't remove it.

marcoagpinto:
> Tbh I like this as a style rule... it does confer the text a more polished feel to say 'a neblina matinal' rather than 'a neblina da manhã'. The one exception would be café, at least in Brazil, since there 'café da manhã' is lexicalised as 'breakfast'. Personally I wouldn't remove it.

@p-goulart 
I had the same feeling about the rule, but I wasn't sure.

I sometimes need a guiding hand to achieve things.

😛 😛 😛 😛 😛 😛 

I will readd it tonight with “café” disabled.

Thank you!

marcoagpinto:
@p-goulart 

Thank you!

I have readded the rule with “café” removed.

❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

p-goulart (approved):
LGTM; btw, I think we could use the same rule for `da tarde` -> `vespertino` and `da noite` -> `noturno`.

marcoagpinto:
> LGTM; btw, I think we could use the same rule for `da tarde` -> `vespertino` and `da noite` -> `noturno`.

Sure, I will work on it during the week or on the weekend.

It is an excellent idea!

PR #9712, opened 2023-11-20
RemoteRule: orthography match, but the original is correct

jaumeortola:


danielnaber:
I've tested this locally and on a test server, looks good. I'll merge this myself or close this PR and use a slightly different version of it. Thanks!

PR #9725, opened 2023-11-22
[fr] VOUS_NOUS_NON_VERBE AP added

LucieSteib:
xml part of https://github.com/languagetooler-gmbh/languagetool-premium/issues/6741

Adding AP to prevent match on 

To be noted ❗ 
LT rule ID seems to be [7]
image_720
but main says [16] 
image_720

The current AP is set on [16] and Main test doesn't find any match anymore. 
To be checked tomorow!


GillouLT (commented):
LGTM! Great AP 🥜 🥜

PR #9744, opened 2023-11-24
[fr] GEMA_23D_FP_fix_batch4

GillouLT:
To address: https://github.com/languagetooler-gmbh/languagetool-premium/issues/6755

ACCORD_V_QUESTION[2] New suggestion added as the user intention could be both singular or plural
ADJ_ADJ_SENT_END[1] AP added (this time actually fixing the FP we encountered during GEMA 23D)
IMP_PRON[12] AP added, no example as the example would need to contain a wrong “ou/où” confusion pair. It’s just to prevent the rule from suggesting gibberish.


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -92061,6 +92068,11 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
                     </and>
                     <token regexp="yes">lui|y|[nv]ous|leur|[mt]oi</token>
                 </antipattern>
+                <antipattern>
```
LucieSteib, line 92071: discussion: so here (IMP_PRON[12]) the example would have been "On vous recommande de vous adresser vers le magasin ou vous l'avez acheté." (indeed it's generating really weird things Oo)

LucieSteib (commented):
Just some questions here, but otherwise LGTM.

PR #9754, opened 2023-11-24
[pt] Removed "temp_off" from rule ID:CONFUSÃO_AO_OU

marcoagpinto:
Zero hits in the nightly results.


susanaboatto (approved):

marcoagpinto:
> Perhaps a "sister rule" to this would be to suggest the infinitive to the verb next to “ao” when it appears to be conjugated in the third-person singular. This is a common mistake:
> 
> `errou o alvo ao mira no objeto` → `errou o alvo ao mirar no objeto`

Sure, I will try to code it at 5am or tonight. 😛😛😛😛😛

marcoagpinto:
@susanaboatto 

I was making the rule, but “mira” appears as a noun and not as a verb, so it isn't possible to create the rule, unless the disambiguator accepts it as a verb.

Screenshot 2023-11-27 at 22-53-03 Análise de Texto - LanguageTool

PR #9765, opened 2023-11-27
[fr] BARBARISME_TU[16]

LucieSteib:
BARBARISME_TU[16] really has bad scores in Grafana (to me, it's clearly a style suggestion).

image
Plus, FR gGEC seems to suggest a change (wich by the way could be considered Style, and definitely premium).
image

Will also fix https://github.com/languagetooler-gmbh/languagetool-premium/issues/5988


GillouLT (approved):
Yes, it's clearly style, maybe removing is the right move :) LGTM!

PR #9770, opened 2023-11-28
[fr] PREP_PAYS-and-TRES-modified

GillouLT:
PREP_PAYS to partially address: https://github.com/languagetooler-gmbh/languagetool-premium/issues/6320
Subrules 7 8 and 9 OFF, these rules work with the tag Z, which is probably the main cause of FPs. Let's give a second chance to the rule and see in a few weeks how it performs after this change.

TRES[8] to address and close: https://github.com/languagetooler-gmbh/languagetool-premium/issues/5423
APs added, even though adding a simple question mark to the example sentence also works.


LucieSteib (approved):
LGTM!

PR #9778, opened 2023-11-29
[fr] TIRET_14-fix

GillouLT:
To fix: https://github.com/languagetooler-gmbh/languagetool-premium/issues/6125


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -123210,6 +123210,11 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
               <token spacebefore="no">-</token>
               <token>pas</token>
             </antipattern>
+            <antipattern>
+              <token regexp="yes">&mois_annee;</token>
+              <token spacebefore="no">-</token>
+              <token regexp="yes">&mois_annee;</token>
+            </antipattern>
```
LucieSteib, line 123217: nitpick(unblocking): I've tried to do the review in codespace, because I thought I could do it myself there, but I don't know how :D 
Maybe adding an example like "j'irai en février-mars" in the AP would be a good idea (?)

LucieSteib (approved):
LGTM!

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
GillouLT: True :) Example added!

PR #9786, opened 2023-11-29
[pt] Rewrote completely rule ID:SIMPLIFICAR_QUE_E_TEM_TÊM

marcoagpinto:

I have rewritten this rule.

Now it is stricter, it has around half of the hits of the previous version, but they are more accurate.

The antipatterns were very complex to create, and it might have this or that false positive, but it is impossible right now for me to look at each one of the 200+ hits and fix one by one.

This was the best I could do in two or three days, just almost entirely dedicated to it.

😛 😋 ❤️ 

```
Portuguese (Portugal): 262 total matches
```

[53.txt](https://github.com/languagetool-org/languagetool/files/13502054/53.txt)

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -17837,96 +17837,139 @@ USA
         </rule>
 
 
-        <!-- QUE/E TEM/TÊM com -->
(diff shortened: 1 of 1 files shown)


p-goulart (approved):
In and of itself, this looks okay, but I'm not convinced this adds that much value. Maybe it should be `picky`?

marcoagpinto:
Done!

I will wait for the tests to finish before committing.

PR #9808, opened 2023-12-01
[fr] SOUHAITER-AI_AP

GillouLT:
To address: https://github.com/languagetooler-gmbh/languagetool-premium/issues/6700

Simple APs, maybe too wide scope, but we want to make sure the FP won't appear again in another form.


LucieSteib (approved):
LGTM!

PR #9822, opened 2023-12-04
[pt] Add minor fix to UNIDADES_METRICAS

p-goulart:
@susanaboatto I had to change the default spellings of the units to `pt-BR`, but also made some changes to the messages so they're less awkward.

---

@danielnaber there is an issue with this rule in that it fails to parse composite expressions like `2.3cm x 2.3cm`. In `pt-BR`, this appears as a frequent `temp_disable` context, and I believe that's also a reason why its apply rate is so low. The same seems to be true in other locales that have a unit conversion rule implemented.

It's prob. not a huge issue right now, so I'm just adding a simple hotfix to get rid of same-unit conversions, which takes care of many of the squaring/cubing FPs. But we might want to rethink the unit conversion rule, esp. for complex dimensions (2.5 x 2.3 x 6.9 cm).

Screenshot 2023-12-04 at 5 07 24 PM

Screenshot 2023-12-04 at 5 12 50 PM

Screenshot 2023-12-04 at 5 13 18 PM

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/AbstractUnitConversionRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/AbstractUnitConversionRule.java
@@ -419,7 +419,8 @@ public abstract class AbstractUnitConversionRule extends Rule {
     return numberRangePart.matcher(textBefore).find();
   }
 
-  private void tryConversion(AnalyzedSentence sentence, List<RuleMatch> matches, Pattern unitPattern, Double customValue, Unit customUnit, Matcher unitMatcher, List<Map.Entry<Integer, Integer>> ignoreRanges) {
+  private void tryConversion(AnalyzedSentence sentence, List<RuleMatch> matches, Pattern unitPattern, Double customValue,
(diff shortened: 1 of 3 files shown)


p-goulart:
There's another minor issue for English, apparently related to the `'` symbol for the Imperial/US Customary foot.

It reads `1'23` as '1 foot 23 inches' (which is sort of bizarre anyway since there are 12 inches to a foot) and converts that (correctly) to `.89 m`. The percent sign should **not** be mistaken for a unit, but it is, and then it says it can't recognise it.

Screenshot 2023-12-04 at 5 15 29 PM

---

(This is actually incorrect and, IMO, picky mode should correct stuff that uses typewriter quotes (`5'10"`) to use the [prime symbol](https://en.wikipedia.org/wiki/Prime_(symbol)) (`5′10″`).

@AzadehSafakish @evan-defran-lt just pinging you here since this is more of an English-language matter. Do let me know if you need to me to describe this in a separate GH issue, though.

susanaboatto (approved):

jaumeortola (approved):

PR #9835, opened 2023-12-06
[fr] couleur_invariable-AI-AP

GillouLT:
To address the looping issue: https://github.com/languagetooler-gmbh/task-force-french/issues/90
But also GEMA 23D gGEC FPs, so: https://github.com/languagetooler-gmbh/task-force-french/issues/85

- AI_FR_GGEC_REPLACEMENT_NOUN_FORM.*: AP added in remote-rule-filter to avoid FPs and loops with grammalecte_g3__gn_couleurs_invariables__b1_a1_1 and ACCORD_COULEUR[1] on, for example, "Elle avait les yeux marron"

AP tested and working :) it's actually the pattern from XML ACCORD_COULEUR[1], I only changed the token from the marker from plural to singular (`[JN].* p` to `[JN].* s`)


LucieSteib (approved):
LGTM!

PR #9848, opened 2023-12-07
[pt-PT] Indent grammar and style files

p-goulart:
Here are the `pt-PT` files indented with 4-space-wide soft tabs, same as the other files.


marcoagpinto (approved):
Thank you, Pedro!

PR #9855, opened 2023-12-07
[pt] remove duplicate scientific names from multiwords.txt

jaumeortola:
Nouns moved to spelling_global.txt


p-goulart:
Will they continue to be PoS-tagged the same way, btw?

jaumeortola:
> Will they continue to be PoS-tagged the same way, btw?

No. But we could add some default tag for all names in spelling_global.txt, if necessary. 
(This tag was necessary in the early versions of the multitoken suggestions, but it is not essential anymore. )

PR #9858, opened 2023-12-07
[fr] D_N[1] AP added

LucieSteib:
To fix https://github.com/languagetooler-gmbh/languagetool-premium/issues/6802.
AP to avoid FPs on "estime" - probably never gonna be used as a noun that way anymay:
image


GillouLT (approved):
LGTM!

PR #9859, opened 2023-12-07
Update RemoteRuleFilters.java

nihadamirov:
If you are using Java 11 or later, I used Files.readString to read the file more simply.I modified the exceptions by catching more specific exceptions and providing meaningful error messages


danielnaber:
> If you are using Java 11 or later, I used Files.readString to read the file more simply

Could you explain why this is an improvement?

PR #9860, opened 2023-12-07
[pt] Added words to wordiness.txt

marcoagpinto:
Heya,

Just one expression enhancement.

❤️ ❤️ ❤️ ❤️ ❤️ ❤️

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/wordiness.txt
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/wordiness.txt
@@ -151,6 +151,7 @@ devido ao fato de=por|pelo|pela
 devido aos efeitos de=porque
 diligência devida=diligência
 diligências devidas=diligências
+durante o decorrer=no decorrer
 é algo que é=é
 é apenas uma questão de tempo antes=eventualmente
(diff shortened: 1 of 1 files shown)


p-goulart (approved):
This looks okay, but I do think we need to have a serious think about `wordiness.txt`...

There are many expressions that are far too greedy and that should be turned into proper XML rules, e.g.:
Screenshot 2023-12-08 at 10 29 37 AM

Others have the opposite problem, and should be turned into XML rules precisely so we can match in all contexts:
Screenshot 2023-12-08 at 10 32 14 AM


And, well, some (most?) should only be triggered when `picky` mode is active, IMO. This rule has a super high temp disable rate (4.7%), which kind of goes to show that we need a more targeted approach.

marcoagpinto:
Yes, several `wordiness.txt` expressions need to be turned into XML rules.

I have already done so with some expressions from there, but it would be great if someone could write a list of them for me to slowly convert them to XML, or perhaps I should look at the nightly diffs.

Anyway, this weekend I have a rule to improve and an academic one to create.

The academic one is turning “básic[ao]s?” to “elementar(es)”, and I was thinking about a rule group for it with two subrules, but I still need to think better to see the best approach.

😋 😛 ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

PR #9861, opened 2023-12-08
[fr] AIT_AIE_AIES_EST_33_fix2

GillouLT:
Fixed suggestion token position in message for QUE_AVOIR[7] and QUE_AVOIR[10] after seeing the diff (there was more to fix). I've checked the rest of the suggestions, and they seem to be fine.

The rulegroup still has FPs...


LucieSteib (approved):
LGTM!
Yeah, that's a complex one... Still, less and less FPs (and better messages and suggestions) thanks to you!

PR #9864, opened 2023-12-08
[pt] Improve DET_FEM_MAS_A_O grammar rule

p-goulart:
- split it, so we also have SUBT_SOBRECOMUM_MASCULINO, which takes care of nouns used in the 'wrong' gender but that make reference to animate entities;

 - improve the disambiguator slightly to never lose the determiner tag of "a" before a feminine singular noun;

 - add entities for commonly misused nouns wrt gender (can be expanded);

 - comment out one example in pt-PT/style that's been affected by the recent change of the disambiguator (@marcoagpinto, you may need to have a look at this).

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
@@ -1711,7 +1711,7 @@
     <!-- Verbo + Preposição: p.ex. 'a' como preposição depois do verbo	-->
     <!-- Created by Tiago F. Santos, Portuguese rule, 2016-11-15  	-->
     <!-- https://pt.wikipedia.org/wiki/A_(preposi%C3%A7%C3%A3o)		-->
-    <rule>
+    <rule> <!-- #1 -->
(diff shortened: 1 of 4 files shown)


marcoagpinto:
> * comment out one example in pt-PT/style that's been affected by the recent change of the disambiguator (@marcoagpinto, you may need to have a look at this).

Sure, I will look at it tonight.

My plan is the following:
Due to the recent changes like Morfologik, I have been waiting for Friday night to download the latest stand-alone and Wikipedia tool so that I can work with all the most recent files, assuring that on the weekend both Susana and you are not working on LanguageTool, so I will have the most recent files tonight.

This is why I have been waiting for the weekend.

marcoagpinto:
@p-goulart 

Heya,

I was looking at it.

The line can be simply commented out.

Anyway, my plan was to revise all rules and antipatterns in 2024+.

If you noticed, I placed dates in the antipatterns, which means that some are very old and require review because of disambiguator plus postags changes.

p-goulart:
@susanaboatto I see you're reviewing some PRs, this one from last week is still pending...

susanaboatto (approved):

PR #9876, opened 2023-12-10
Add a grammar rule about verbs followed by a name in french

Sharcoux:
This rule adds support for the grammar rule detailed here: `https://fr.wiktionary.org/w/index.php?title=Cat%C3%A9gorie:Verbes_en_fran%C3%A7ais_suivis_d%E2%80%99un_nom_commun_sans_d%C3%A9terminant&pageuntil=faire+pression#mw-pages`. This rule should detect sentences like "J'ai fée des pâtes" and the likes, where a noun follows a verb.

This rule could probably be improved a bit with the help of someone with more knowledge. We could provide as suggestions the list of the nouns available after each verb.

I'm still trying to figure out how to test my changes.


GillouLT:
We don't really have a straightforward solution for it, you can look into the [WordWithDeterminerFilter](languagetool-language-modules/fr/src/main/java/org/languagetool/rules/fr/WordWithDeterminerFilter.java), but a determiner would be required in the pattern. Although, WordWithDeterminerFilter can be changed or a new rule filter can also be created for determiner additions.

The following line would also work for some cases:
`<match no="2" postag="N ([mfe]) (s|p|sp)" postag_regexp="yes" postag_replace="D $1 $2">le</match>`

Sharcoux:
I went to generating suggestions like "un camion|une camion|des camion", knowing that 2 of them would be wrong, but another rule would solve that afterwards. Tell me if another approach would have been better.

Sharcoux:
The rule seems ready. We should test it on a corpus. There might be cases I forgot, but if we keep adding them, this rule should detect a lot of mistakes that are currently being ignored.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -115936,6 +115936,1297 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+                <suggestion>des \2</suggestion>
+                <example correction="un rhume|une rhume|des rhume">J'attrape <marker>rhume</marker>.</example>
+                <example>J'attrape <marker>froid</marker>.</example>
+            </rule>
+            <rule id="AVOIR_PLUS_NOM_INCORRECT" name="avoir suivi d'un nom incorrect" default="temp_off">
+                <pattern>
+                    <token postag="V.*" postag_regexp="yes" inflected="yes">avoir</token>
+                    <marker>
+                        <token postag="N.*" postag_regexp="yes">
+                            <exception>accès</exception>
```
LucieSteib, line 115979: **suggestion(format):** here and everywhere in the PR, change the format of the list to a shorter, more readable and manageable list, like this line (there are many others in the file):
`<exception scope="next" inflected="yes" regexp="yes">humain|et|ou|de</exception></token>`

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -115936,6 +115936,1297 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
...
+                            <exception>vent</exception>
+                            <exception>voix</exception>
+                        </token>
+                    </marker>
+                </pattern>
+                <message>Après le verbe 'avoir', ce nom n'est généralement pas utilisé.</message>
+                <suggestion>un \2</suggestion>
+                <suggestion>une \2</suggestion>
+                <suggestion>des \2</suggestion>
+                <example correction="un voiture|une voiture|des voiture">J'ai <marker>voiture</marker>.</example>
```
LucieSteib, line 116030: **question(blocking):** What's intended here? On the sentence "j'ai fée des pâtes", for example, the rule would produce suggestions leading to "j'ai un|une|des fée des pâtes". How is that useful for the users, especially compared to the very high rate of false positives I can already see (j'ai piscine, poney, clarinette...).

LucieSteib (changes requested):
Hi Sharcoux, thanks for your PR! Please find here a review (in code conventional comments).

**thought:** All the rules here aim at changing the noun after the verb. But your original idea was to change the verb when confused with a homophone or near homophone, correct (fait/fée)? Maybe a less broad approach in that direction would be better than creating rules that will probably result in more false positives than true positives.

`I'm still trying to figure out how to test my changes.`
Please use the RuleEditor in ExpertMode to have at least a first check on the rules (you'll have to do that one by one since the tool can't handle rule groups). That would also maybe help with having a first idea of the corpus-based real situations we are dealing with.

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Sharcoux: You're right that the suggestion is not so relevant in that context, however, it usefully drags the attention of the user to the problem that "fée" is a noun. Also, "j'ai une fée des pâtes" is grammatically correct, provided that such a fairy exist (or that you are married to a skilled pasta cooker ^^).
Anyway, generating suggestions for those errors will always be tricky, but detecting the error is already very useful.

Your point about "j'ai piscine" and such is a very strong point. We might have to give up the rule about "avoir", just like we did with "parler", because too many words could go there. We could, still detect some nouns that should not go there, like "fée", for people with dyslexia.

Tell me if that strategy would seem ok to you.

LucieSteib: Yes, that could work.
Please ensure to do the other changes requested 💪

languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
Sharcoux: I don't think that "inflected" can combine with "regexp". At least, chatGPT says it doesn't. I can replace this in most of the rules except the last one.

LucieSteib: Testing any rule from the grammar.xml file that contains the sequence can prove ChatGPT isn't always right 😉 
See for example the matches found for this rule (randomly selected amongst the 678 matches for "inflected="yes" regexp="yes"" in the xml file)

```
            <rule>
                <pattern>
                    <token inflected="yes" regexp="yes">accroupir|assoir|étendre|cacher|chier|coucher|cracher|dormir|jeter|poser|rouler|tomber|traîner|transport|vomir</token>
                    <marker>
                        <token>parterre</token>
                    </marker>
                </pattern>
                <message>Confusion entre le nom « parterre » et la locution « par terre » (sur le sol).</message>
                <suggestion>par terre</suggestion>
                <example correction="par terre">Ils dorment <marker>parterre</marker>.</example>
            </rule>
```

will have matches like `Les dames assises parterre lisent ou brodent en compatissant, Les chorégraphies sur cette chansons sont telles qu'avant le refrain, toute la foule s'accroupit parterre.`
It was, though, just an example. You might not even want to inflect the tokens in the list.

LucieSteib (changes requested):

LucieSteib (approved):
Alright, LGTM.
@Sharcoux is there a reason why you force-push your commits?  It’s unlikely you need to change the history from one commit to the very next? That might compromise the merging into master. 
@GillouLT you can also have a look and request changes or give your approval.
Anyway we'll see some matches in the next day diff (some more exceptions might really need to be added or the rules changed to narrow the scope).

Sharcoux:
> @Sharcoux is there a reason why you force-push your commits? It’s unlikely you need to change the history from one commit to the very next? That might compromise the merging into master.

I force push on my own repo. This should not have any impact on your end. The reason is just because, to be honest, I still haven't set up a dev environment for Java, and I'm using the CI to do the validity check. When I realize that I made a typo, I just amend the commit.

GillouLT (approved):
Yes, I also think we're good to go :)

Sharcoux:
I don't know the rest of the process, but if some new problems are met, don't hesitate to list them here, I'll fix the MR to add the required exception.

Sharcoux:
Discussion about this rule is also here: https://github.com/languagetool-org/languagetool/pull/9875

Some fixes are needed.

PR #9883, opened 2023-12-11
[pt] Added rule ID:BÁSICO_ELEMENTAR

marcoagpinto:

Here is one of the rules I was going to work on the weekend.

```
Portuguese (Portugal): 113 total matches
Portuguese (Portugal): 811129 total sentences considered
```

[_REGRAS_1e2_20.txt](https://github.com/languagetool-org/languagetool/files/13635040/_REGRAS_1e2_20.txt)

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -8402,6 +8402,75 @@ USA
     </category>
     <category id='ACADEMIC' name='Linguagem acadêmica' type='style' tone_tags="academic">
 
+
+        <rulegroup id='BÁSICO_ELEMENTAR' name="[Universitário][Científico] Básico → Elementar" tone_tags="academic" is_goal_specific="true" default="temp_off">
+
+            <!-- Subrule 1: "V.+ substantivos básic[ao]s?" -->
+            <rule>
+                <antipattern>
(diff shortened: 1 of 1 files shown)


p-goulart:
Looking at the APs and exceptions here, I don't see much to suggest we won't trigger this rule when chemists write about alkali.

> A teoria de Arrhenius limita-se a substâncias que reagem em meio aquoso e não explica o **comportamento básico** de algumas substâncias

> Para tal, use uma **solução básica**.

> atuando como ácido de Lewis em **meio básico**

marcoagpinto:
@p-goulart 

Pedro,

I have added the exceptions you suggested.

❤️ ❤️ ❤️ ❤️  😋 😋 😋 😋 😛 😛 😛 😛

p-goulart:
Okay, let's see how it performs live.

But I'm not convinced that we can reliably tell apart domain-specific uses of 'básico' from the ones that can be replaced with 'elementar' with just the word list you've provided. (Or any word list.)

Those were just a couple of examples from a cursory look at chemistry articles. When I look at common collocates for 'alcalino', I see some things that should definitely be here, like `pH`, `sangue`, `proteína`, `solo`, etc. Definitely some more research is needed.

marcoagpinto:
Thank you, Pedro.

While developing the rule, I had to limitate myself to the results provided by `pt-BR.txt`.

More words can be added to the rule as they come by.

PR #9886, opened 2023-12-11
[pt] Add NUMERO_NEGATIVO_SINAL_DE_MENOS

p-goulart:
This is very restrictive, which I think for now is okay. It shouldn't trigger **unless** the hyphen/dash is preceded by a space.


p-goulart:
@susanaboatto can you have a look at this before I merge #9892 ? They're kind of related.

susanaboatto (approved):

PR #9893, opened 2023-12-12
[pt] Add city/state separator rules

p-goulart:
At some point, this should be an advanced user option. For now, we're enforcing en-dash-no-spaces for Brazilian states and commas for US states.


languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -41769,6 +41769,110 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
...
+            actually check the *city name*, but this is a bigger thing.
+        -->
+        <!--State abbreviations common to both Brazil and the US:
+            AL: Alagoas/Alabama
+            MA: Maranhão/Maine
+            MS: Mato Grosso do Sul/Mississippi
+            MT: Mato Grosso/Montana
+            PA: Pará/Pennsylvania
+            SC: Santa Catarina/South Carolina
+        -->
```
susanaboatto, line 41827: I have compiled a list of cities in those Brazilian states, since those are the more urgent ones: [Municípios Brasil.zip](https://github.com/languagetool-org/languagetool/files/13661608/Municipios.Brasil.zip), maybe we could turn them into an entity?

susanaboatto (approved):

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
p-goulart: An entity would prob. be super slow. If we have a list, though, we could just add it to `resource/pt` and have a simple Java filter to perform a lookup, maybe?

PR #9896, opened 2023-12-12
[pt] add exceptions to ELISAO_VERBAL_SEM_ENCLITICO and INFINITIVE_NOUN

susanaboatto:
False alarms removed based on [diff results](https://internal1.languagetool.org/regression-tests/via-http/2023-12-06/pt-BR/result_grammar_ELISAO_VERBAL_SEM_ENCLITICO%5B1%5D.html).

I will create an issue to fix other problems found, namely:

- Add a suggestion to include the simple future instead of the infinitive for `estrá` and `responsabilizá` as a typo: `tudo estrá nos conformes. → tudo estará nos conformes` // `O comprecar.com.br não se responsabilizá pelos anúncios constantes de seu site`
- Replace `=` with hyphen: `executá=lo → executá-lo`
- Add confusion pair rule for viuvá vs. viuva: `viuvá de Vasco Lourenço de Almada`
- Create confusion pair rule for atê vs. até
- Create rule to fix wrong use of en dash for verbs
- Fix missing hyphen before pronouns

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
@@ -1469,7 +1469,7 @@
     <!-- exception to DET_VERBONOME -->
     <antipattern>
       <token>de</token>
-      <token regexp="yes">podere?s?</token>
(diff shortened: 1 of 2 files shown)


p-goulart:
Most of these will disappear once we fix the inflection and tokenisation.

p-goulart (approved):

PR #9908, opened 2023-12-13
[pt] Re-do ordinal abbreviation rules

p-goulart:
- there was a lot of weirdness with ordinal indicators and degree signs, esp. in the disambiguator, so that was ironed out;

- created entities to make *working* with these characters easier, as it is currently impossible for a human to reliably tell all the tiny little circles apart;

- reworked the tagger to tag certain numerical expressions (viz. degrees, ordinals, and percentages) with *logic*, instead of hardcoding thousands of entries for numbers in the tagger file...

- ... so, yes, I've removed a bunch of lines from the tagger file;

- the rules themselves for the abbreviation required extensive reworking, both to use the new entities and to untangle the mess of degree signs and ordinal indicators;

- we are also going to be using the superscript letters (those traditionally used for other abbreviations) for the plural forms of ordinals, so that's also changed;

- degree rules had to be extensively redone to account for cases that had to arrive *clean* to the ordinals rule.

---

@GillouLT:
For the ordinal abbreviation rules themselves, we'll need to have `ORDINAL_ABBREVIATION_DOT` and `ORDINAL_ABBREVIATION_NO_DOT` alternating depending on the value of the user options. (I.e., one will be `on` and the other `off`.) Is this okay? Also, this rule is not `picky`, that doesn't quite match the UI concept...

@jaumeortola can you just have a cursory look at the Java? Mostly just the tagger.

@susanaboatto ... well, this ended up being much more than we'd hoped, but I promise there wasn't much we could do to get the ordinal abbreviation rule working without doing this. The most important thing here is to read the new rules carefully to see if there are any blind spots.

Also, we may want to move some of these rules to `picky`, but that's not done in this PR yet. Maybe you can propose something?


jaumeortola:
The changes in the tagger are okay. Alternatively, the same could be accomplished with a few rules in disambiguation.xml.

jaumeortola (approved):

susanaboatto:
I think the minus sign rule should probably be picky. Replacing ordinals with the degree symbol should be a standard, so I wouldn't make those picky. But we should check the results and see if it'd be necessary. I have a feeling that the other way around (turning the degree symbol into ordinal) won't be as well-accepted, so perhaps these rules could become picky. These decisions could be taken after we monitor how the rules behave.

PR #9916, opened 2023-12-14
[pt] add rule to CRASE_CONFUSION_2

susanaboatto:
Quick fix to detect missing crase in `Casa a venda`, etc.


p-goulart (approved):
✨ let's help them sell that house ✨

PR #9917, opened 2023-12-14
[pt] add confusion pair rule ENXARCA_ENCHARCA

susanaboatto:
Quick fix for another confusion pair feedback:

Screenshot_2023-12-13-19-21-49-777_com android chrome-edit

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -31900,6 +31900,15 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
             </rule>
         </rulegroup>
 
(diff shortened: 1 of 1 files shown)


p-goulart (approved):
ngl I had to look this up like multiple times; do we even really need `enxarca` in the dictionary? Looks like a rare variant of `enxerca`, we might as well delete it.

susanaboatto:
I was worried about the super picky users, but chances are nobody is going to use `enxarca` in the correct context. I would be okay with removing it from the dictionary.

p-goulart:
> I would be okay with removing it from the dictionary.

Maybe we could have a live session at some point where we release a new dictionary from your machine? We could trojan-horse 'Moóca->Mooca' along with this change :p

I mean, this change is okay, and we should merge it anyway, but it might be a Good First Issue for dictionary deployment for you?

susanaboatto:
> Maybe we could have a live session at some point where we release a new dictionary from your machine? We could trojan-horse 'Moóca->Mooca' along with this change :p

Yes, let's arrange that

PR #9929, opened 2023-12-15
[fr] BALLADE adding APs

LucieSteib:
To fix https://github.com/languagetooler-gmbh/languagetool-premium/issues/6820 and a FP found when looking at the rule diffs.


GillouLT (approved):
LGTM!

PR #9931, opened 2023-12-16
[pt] Improved rule ID:BÁSICO_ELEMENTAR

marcoagpinto:
Heya, Susana and Pedro,

A few false positives removed, and came up with a strategy:
I will look at the next night diff and will only use valid words in the next change instead of using all with exceptions.
(Fewer results but more accurate)
(WIP).

Thanks!


p-goulart (approved):
✨

PR #9932, opened 2023-12-16
[pt-PT] Added "languages" entity file into grammar.xml and style.xml

marcoagpinto:
Heya,

I added the file of the language entity into pt-PT, I will slowly be adding the entities.


p-goulart (approved):
@marcoagpinto approved; for these kinds of PRs that aren't introducing any new logic or changing matches, I think it's enough that you wait for the tests to pass before merging, no need to require a review for it.

marcoagpinto:
> @marcoagpinto approved; for these kinds of PRs that aren't introducing any new logic or changing matches, I think it's enough that you wait for the tests to pass before merging, no need to require a review for it.

Sure, thanks!

PR #9934, opened 2023-12-16
[pt] Changed name and improved rule ID:ARREAR_ARRIAR

marcoagpinto:
Heya,

I have made the rule more useful.


p-goulart (approved):
Rule logic is okay, but I personally feel like for such terms it might be worth it to have the '(Possível) confusão de termos' message as the short message and a longer message that actually makes reference to the meaning of the verbs?

marcoagpinto:
> Rule logic is okay, but I personally feel like for such terms it might be worth it to have the '(Possível) confusão de termos' message as the short message and a longer message that actually makes reference to the meaning of the verbs?

I am brain toasted 😋 I can't think of good messages.

PR #9935, opened 2023-12-17
[pt] Improved rule ID:CONFUSÃO_APROXIMA_A_PRÓXIMA

marcoagpinto:
The rule is now more useful.


p-goulart (approved):
I see the value, but it'd be nice, if we're doing this, to also suggest 'aproximar', so we don't lose matches like this. (Not sure about the relative priority here.)

Screenshot 2023-12-18 at 11 12 52 AM

marcoagpinto:
> I see the value, but it'd be nice, if we're doing this, to also suggest 'aproximar', so we don't lose matches like this. (Not sure about the relative priority here.)
> 
> Screenshot 2023-12-18 at 11 12 52 AM

Sure, I will do it tonight. 😛

PR #9936, opened 2023-12-17
[pt] Improved rule ID:CONFUSÃO_ASSERTO_ACERTO_DE_CONTAS

marcoagpinto:
The rule is now more useful.


p-goulart (approved):
Like with most confusion pair rules where at least one of them is a rare term, I think it'd be a good idea to add a little more to the message.

PR #9938, opened 2023-12-17
[pt-PT] Added foreign words to entity

marcoagpinto:
Added just one word.


p-goulart (approved):
Looks okay. I'd say if you're just adding a single rule to a regexp like this, it's prob. enough to wait for the tests to pass and then merge. (Keep the PR so we're aware, but there might be no need to wait for an approving review. Esp. if it's a `pt-PT` rule.)

PR #9945, opened 2023-12-18
[fr] D_AVANTAGE new subrule

LucieSteib:
In the context of https://github.com/languagetooler-gmbh/languagetool-premium/issues/5124.

As pointed out by Anne Sophie, in some cases, `d'avantage` et `davantage` can be both correct.
The new subrule proposed here could maybe improve the user experience when encountering this potential FPs.
The message is designed to provide the elements necessary for an informed choice for the user.


languagetool-language-modules/fr/src/main/resources/org/languagetool/rules/fr/grammar.xml
```diff
@@ -22373,22 +22373,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
                 <example correction="vôtre">Je deviens <marker>votre</marker> et vous êtes mien.</example>
             </rule>
         </rulegroup>
-        <rule id="DAVANTAGE" name="Davantage au lieu de d’avantage">
```
LucieSteib, line 22376: Cheh, no idea why this section looks “removed”, I swear it's still there :D

GillouLT (approved):
LGTM!!

PR #9946, opened 2023-12-18
[pt] Add logic for Brazilian municipality/state mapping

p-goulart:
The core of the rule is now a regexp, and we just call the rule filter to make sure we're dealing with an actual city.

This rule doesn't check if the city actually belongs in that state – that'll be a different rule. The idea here is that only things that could be conceivably thought of as Brazilian cities should trigger the state abbreviation separation rule.

We still need to improve (or apply a similar logic?) to the US states rule.

### DIFF
--- a/languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/BrazilianToponymFilter.java
+++ b/languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/BrazilianToponymFilter.java
@@ -1,0 +1,49 @@
+/* LanguageTool, a natural language style checker
+ * Copyright (C) 2012 Jaume Ortolà i Font
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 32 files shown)


languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/BrazilianToponymFilter.java
```diff
@@ -0,0 +1,49 @@
...
+    String toponym = matcher.group(1);
+    String underlined = matcher.group(2);
+    String state = matcher.group(3);
+
+    // TODO: read this from user options or something...
+    String suggestion = "–" + state;
+    if (suggestion.equals(underlined)) {
+      return null;
+    }
+    BrazilianToponymMap map = new BrazilianToponymMap();
```
danielnaber, line 41: Would it make sense to make this a static constant? We maybe don't want to re-initialize the object every time there's a potential match.

p-goulart: Yes, we should definitely do that. I was certain I had, but I think I moved it around a bit and forgot to put it back.

susanaboatto (approved):

languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/BrazilianToponymFilter.java
p-goulart: done

PR #9953, opened 2023-12-19
[de] deactivated rule + typo fixes + added words

Luisa-LT:


languagetool-language-modules/de/src/main/java/org/languagetool/language/German.java
```diff
@@ -522,6 +522,7 @@ protected int getPriorityForId(String id) {
       case "DOPPELUNG_GLEICHES_VERB": return -55; // prefer comma rules
       case "FEHLENDES_NOMEN": return -60; // lower prio than most rules
       case "REPETITIONS_STYLE": return -60;
+      case "WORD_REPEAT_BEGINNING_RULE": return -61;
```
jaumeortola, line 525: The actual ID is GERMAN_WORD_REPEAT_BEGINNING_RULE. 
There are similar IDs for other languages.

PR #9959, opened 2023-12-19
[pt] Fix token and enable CRASE_CONFUSION_2

susanaboatto:
Fixed exception to remove the [remaining false positives](https://regression.languagetoolplus.com/via-http/2023-12-19/pt-BR_full/result_grammar_CRASE_CONFUSION_2%5B6%5D.html). I think now we can enable it.


p-goulart (approved):
✨

PR #9981, opened 2023-12-22
[pt] Add logic for state info and toponym check

p-goulart:
The rule itself is yet to come, I've tested it, but I want to have these classes here first, because I may want to do something also in the premium repo and this needs to be in place first.

---

@susanaboatto as per our Slack convo, if you can think of a clear and concise way to phrase this (that doesn't violate our own style rules :p), just let me know...

### DIFF
--- a/languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/BrazilianToponymStateCheckFilter.java
+++ b/languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/BrazilianToponymStateCheckFilter.java
@@ -22,29 +22,48 @@ import org.languagetool.AnalyzedSentence;
 import org.languagetool.rules.RuleMatch;
 import org.languagetool.rules.patterns.RegexRuleFilter;
 
-import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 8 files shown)


languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/BrazilianToponymStateCheckFilter.java
```diff
@@ -0,0 +1,70 @@
...
+      .map(stateInfo -> stateInfo.abbreviation)
+      .collect(Collectors.toList())
+    );
+  }
+
+  private void setMessage(RuleMatch match, BrazilianToponymStateCheckResult checkResult, BrazilianStateInfo wrongState) {
+    String inTheStateOf = String.format("no estado %s %s",
+      new PortuguesePreposition("de").contractWith(wrongState.articles[0]),
+      wrongState.name);
+    String message = String.format("Se estiver se referindo a um município, não parece haver cidades com o nome de %s %s.",
```
susanaboatto, line 66: I think `Se estiver se referindo a um município, não parece haver cidades com o nome %s %s neste estado.` works. At least it's more accurate than no mention of a state at all.

susanaboatto (approved):

p-goulart:
@jaumeortola can you have a look at this PR 🙏🥺

PR #9983, opened 2023-12-22
[pt] Fix mispelled names

susanaboatto:
- Some names apparently contain the wrong suggestion.

What I found in the PT diff:

_Iuri Gagarine_ (wrong) suggested for **Iuri Gagarin** (correct);
_Ulysses Guimaraes_ (wrong) suggested for **Ulysses Guimarães** (correct);
_Waldir Maranhao_ (wrong) suggested for **Waldir Maranhão** (correct);
_Jorge Vercilo_ (wrong) suggested for **Jorge Vercillo** (correct).

- Probably a different issue, but I have also added an extra S to the personality names I have found in the neighboring tags, i.e., replaced the 0 in NPMS**0**00 with NPMS**S**00. It should be used to distinguish people from places and organizations, but our words are barely tagged as such. 

- More names added to multiwords.txt, spelling.txt, and spelling_global.txt based on the PT diff findings.

- There are some wrong suggestions in the PT_MULTITOKEN rules that will need more attention. For example, `Pára-quedistas` is corrected to  `para-quedistas`, when it should be `paraquedistas` (post-1990). For now, I am adding these to `spelling.txt`, but a more in-depth fix for this will be needed. 

@p-goulart feel free to revert, edit, or comment this branch with your insights.


jaumeortola:
I wouldn't add multitoken expressions to `spelling.txt`. They should go, preferably, to `multiwords.txt`.

Some person names could also be added to global_spelling.txt (but not Russian names like Gagarin).

jaumeortola (approved):

p-goulart:
I'm doing several things here:
- moving a bunch of multi-word entries added to `spelling.txt` to `multiwords.txt`;
- removing all new prefixed words that are being added to `spelling.txt` and adding tests to the Morfologik rule to show that those words are already correctly recognised;
- moving hyphenated words to the compounds dictionaries in the source files (new binaries to be deployed today);
- the correction `pára` -> `para` is *expected behaviour* (and once applied Morfologik suggests the correct form); either way, the correct word is already in the dictionary and it makes no sense to duplicate it in `spelling.txt`. I'm adding `pára- para` and `para- para` as pairs to the Morfologik info file.


Be that as it may, we need to make sure we check what the speller can already handle before adding entries blindly to `spelling.txt`. And we definitely need to add and run Java tests.

p-goulart:
I've moved a bunch of stuff off the `spelling.txt` and `multiwords.txt` lists and recompiled the dictionary. It was much more work than I hoped, but I think the overall result is pretty good. I mean, tests pass, and I don't think we're losing too much. I've had to quite a bit of research on the terms there, and ended up making some choices, esp. with regards to hyphenation, juxtaposition, or using a space for English compound loanwords (e.g. `taskforce`, `task-force`, `task force`). We're losing very little here in terms of actual coverage, as these are already pretty rare words to begin with.

Those files are very close to being clean, though there are still a couple of hundred entries in `spelling.txt` that we prob. need to check against some kind of foreign terms rule, since those are mostly very rarely used (or used only in `pt-PT`) equivalents of loanwords that have already had their orthography adapted. Like `coupé` instead of `cupê`, `vinaigrette` instead of `vinagrete`, etc. It's not that many (just a few hundred) and sounds like it should be a separate task, since it touches upon the 'barbarism' rules.

languagetool-core/src/main/resources/org/languagetool/resource/spelling_global.txt
```diff
@@ -19997,4 +19997,268 @@ DB Schenker
...
+Jennifer Walton
+Marguerite Durand
+María Moliner
+Renee Walker
+Alexis Rose
+Josie Maran
+Leonora Fani
+Marnie Simpson
+María Gabriela
+Adrian Fernández
```
jaumeortola, line 20027: This seems a misspelling in Spanish. Adrián Fernández: https://en.wikipedia.org/wiki/Adri%C3%A1n_Fern%C3%A1ndez

languagetool-core/src/main/resources/org/languagetool/resource/spelling_global.txt
```diff
@@ -19997,4 +19997,268 @@ DB Schenker
...
+John Walton
+John Wisker
+Jonas Falk
+José Marzán
+Joseph Bové
+Joseph Cilona
+Josué Montello
+José Blanc de Portugal
+José Mouzinho d'Albuquerque
+José Mouzinho de Albuquerque
```
jaumeortola, line 20132: Both `de` + `d'`?

languagetool-core/src/main/resources/org/languagetool/resource/spelling_global.txt
```diff
@@ -19997,4 +19997,268 @@ DB Schenker
...
+Michael Carlin
+Michael Harner
+Michael Morton
+Michael Norman
+Michael Steger
+Michael Tejera
+Michael Warner
+Miguel Burnier
+Mike Penner
+Mikhail Yuryevich
```
jaumeortola, line 20165: Avoid Russian names in global_spelling.txt because different languages use different spellings. 
Is this one? https://pt.wikipedia.org/wiki/Mikhail_L%C3%A9rmontov

languagetool-core/src/main/resources/org/languagetool/resource/spelling_global.txt
jaumeortola: Yes. That's true. Don't worry about them. We can fix them afterward.

PR #9984, opened 2023-12-23
[pt] Improved/rewrote rule ID:BÁSICO_ELEMENTAR

marcoagpinto:

I have rewritten the rule.

Now, instead of two subrules, all is done in one rule.

The words it accepts are there, making it easier to add or remove words.

This means, that if one of you detects false positives, it is just a matter of deleting the specific word from the rule.


p-goulart (approved):
A more cautious approach, but I think this is wiser in this case, esp. for `academic`-tagged rules.

PR #9991, opened 2023-12-27
[pt] Added rule ID:QUE_SE_VERBO_PART_PASSADO

marcoagpinto:

After three or four days working on this rule, it is ready.

Only yesterday I noticed that there was already a rule doing a similar thing, but this one is more powerful.

I need to search for the older rule and see if it can be deleted.

Thanks!

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -12209,6 +12209,99 @@ USA
     <category id='SHORTEN_IT' name="Linguagem concisa" type="style">
 
 
+        <rule id='QUE_SE_VERBO_PART_PASSADO' name="Simplificar: 'Que se Verbo' → Verbo Part. Passado" default="temp_off">
+            <!-- IDEA shorten_it -->
+
+           <antipattern>
+                <token regexp='no' inflected='yes'>ser</token>
+                <token postag='(SPS00:)?D[AI].+' postag_regexp='yes'/>
+                <token postag='NC.+' postag_regexp='yes'/>
+                <token>que</token>
+                <token>se</token>
+                <token postag='V.+' postag_regexp='yes'>
+                    <exception scope='next' postag_regexp='yes' postag='SPS00:DI.+|RM|RG'/>
+                    <exception scope='next' regexp='yes'>a|em|entre</exception> <!-- SPS00 -->
+                    <exception scope='next' regexp='yes'>ao|n[ao]s?</exception> <!-- SPS00:DA.+ -->
+                </token>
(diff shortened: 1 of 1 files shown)


marcoagpinto:
“Lightweight”:

```
Portuguese (Portugal): 59 total matches
Portuguese (Portugal): 811129 total sentences considered
```

[63stable.txt](https://github.com/languagetool-org/languagetool/files/13777780/63stable.txt)

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
```diff
@@ -12209,6 +12209,99 @@ USA
...
+                </token>
+                <token>
+                    <exception postag_regexp='yes' postag='DA.+|Z0.+'/>
+                    <exception scope='next' postag_regexp='yes' postag='DA.+|CS'/>
+                </token>
+            </pattern>
+            <filter class="org.languagetool.rules.pt.AdvancedSynthesizerFilter" args="lemmaFrom:5 lemmaSelect:V.* postagFrom:2 postagSelect:NC(.)(.).* postagReplace:VMP00[\b2][\b1].*" />
+            <message>&simplify_msg;</message>
+            <suggestion>{suggestion}</suggestion>
+            <example correction="levado">O Rui calcula o tempo <marker>que se leva</marker> a percorrer cada cidade.</example>
```
p-goulart, line 12289: This is ungrammatical in `pt-BR`. 'O tempo levado' sounds like 'naughty time'.

p-goulart (changes requested):
If it *is* a more powerful rule, then I think we should see more *positive* examples. We see many *negative* ones (also in the APs), but I'm not convinced we don't need more *semantic* constraints, as well as syntactic.

Another nitpick I have with this is that the examples are *very long*, to the point that it takes me a while to even find the specific fragment that contains the relevant construction. I'd shorten them so that they preserve the context but don't have dozens of unrelated words around.

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
marcoagpinto: @p-goulart 

Maybe make it pt-PT only?

p-goulart: If we can't find any examples of this being grammatical in `pt-BR` but we can in `pt-PT`, then I suppose we could make it dialect-specific, yes.

marcoagpinto:
@p-goulart 
Tonight I will close this pull request and move it to pt-PT and open another pull request.

❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

marcoagpinto:
Moved it to pt-PT and created a new pull request.

PR #10016, opened 2023-12-29
[de] Better German ss -> ß message

danielnaber:
Affected errors (examples):
* Aussenstände -> Außenstände
* gross -> groß

Old: `Möglicher Tippfehler gefunden.`

New (examples):
* `Nach einem Doppellaut aus zwei Vokalen (hier: ei) schreibt man 'ß' statt 'ss'.`
* `Nach einem lang gesprochenen Vokal (hier: o) schreibt man 'ß' statt 'ss'.`


languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -61,7 +61,7 @@ public class GermanSpellerRule extends CompoundAwareHunspellRule {
 
   private static final String adjSuffix = "(affin|basiert|konform|widrig|fähig|haltig|bedingt|gerecht|würdig|relevant|" +
     "übergreifend|tauglich|untauglich|artig|bezogen|orientiert|fremd|liebend|hassend|bildend|hemmend|abhängig|zentriert|" +
-    "förmig|mäßig|pflichtig|ähnlich|spezifisch|verträglich|technisch|typisch|frei|arm|freundlich|feindlich|gemäß|neutral|seitig|begeistert|geeignet|ungeeignet|berechtigt|sicher|süchtig|verträglich)";
+    "förmig|mäßig|pflichtig|ähnlich|spezifisch|verträglich|technisch|typisch|frei|arm|freundlich|feindlich|gemäß|neutral|seitig|begeistert|geeignet|ungeeignet|berechtigt|sicher|süchtig)";
```
danielnaber, line 64: This change is unrelated and shouldn't be in this PR

languagetool-core/src/main/java/org/languagetool/rules/spelling/hunspell/HunspellRule.java
```diff
@@ -244,17 +244,14 @@ public RuleMatch[] match(AnalyzedSentence sentence) throws IOException {
           ruleMatch.setType(RuleMatch.Type.UnknownWord);
           String cleanWord2 = cleanWord.substring(dashCorr);
           if (userConfig == null || userConfig.getMaxSpellingSuggestions() == 0 || ruleMatches.size() <= userConfig.getMaxSpellingSuggestions()) {
-            ruleMatch.setLazySuggestedReplacements(() -> {
```
danielnaber, line 247: The laziness is lost, didn't find a simple way to keep it and still have a match-specific message

danielnaber:
@donnerpeter This PR reverts a change you made a 2-3 years ago, namely using `setLazySuggestedReplacements`. If you have an idea how to get match-specific messages and still keep the lazy suggestions, please let me know.

donnerpeter:
> @donnerpeter This PR reverts a change you made a 2-3 years ago, namely using `setLazySuggestedReplacements`. If you have an idea how to get match-specific messages and still keep the lazy suggestions, please let me know.

Great messages! Maybe check if `replaceFirst("ss", "ß")` is a valid word and then produce these messages?

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanTools.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package org.languagetool.rules.de;
+
+public class GermanTools {
+
+  private GermanTools() {
+  }
+
+  public static boolean isVowel(char c) {
+    return (c == 'a' || c == 'e' || c == 'i' || c == 'o' || c == 'u' || c == 'y' ||
```
donnerpeter, line 27: A shorter way: `"aeio...".indexOf(c) >= 0`

danielnaber:
> Maybe check if replaceFirst("ss", "ß") is a valid word and then produce these messages?

That shouldn't be needed, as we're comparing to the suggestion, which should always be a valid word.

> A shorter way: "aeio...".indexOf(c) >= 0

Thanks, I'll use that.

donnerpeter:
> > Maybe check if replaceFirst("ss", "ß") is a valid word and then produce these messages?
> 
> That shouldn't be needed, as we're comparing to the suggestion, which should always be a valid word.

I meant such a check could work without any suggestions, thus preserving their laziness.

danielnaber:
> I meant such a check could work without any suggestions, thus preserving their laziness.

I see... or maybe I don't. It seems I'm not familiar enough with the code, I somehow cannot get the message set after the suggestions have been calculated. If you could provide some code, that would be great.

donnerpeter:
I'm on vacation now, so I could look at the code after 8.01. But I'm not very familiar with the code either :)
And it seems that the laziness still isn't really used by IntelliJ (where it was supposed to be used), so its loss might not be that bad after all.

PR #10019, opened 2023-12-30
[pt-PT] Moved category from grammar.xml to style.xml, and AO45 category to inside the real AO45 category

marcoagpinto:

I have moved the “ERROS DE TRADUÇÃO” category from inside style.xml into grammar.xml, just like it happens in the main PT file.

Also, there was a category with an AO45 rule whose rule I moved to inside the main category AO45 and turned the rule off because it needs to be rewritten.

Somewhere during the week, I must also create a “SHORTEN IT” category in the style.xml.

This is all regarding pt-PT.

Thanks!

Happy New Year!


p-goulart (approved):

marcoagpinto:
@p-goulart 

Are you working on the holiday?

Only now I saw the approval because I haven't been looking much at the mails today.

Thanks!

PR #10023, opened 2024-01-02
[en] change the encoding of NZ spelling dictionary

jaumeortola:
The encoding was wrong. It was not much relevant because there are few words with diacritics in English (only foreign terms). 

I don't know if there will be some undesired effects after this change. Do we need more testing?

The encoding was changed here: https://github.com/languagetool-org/languagetool/commit/8bb96ad05b0060158a93504337eb2dcfcecaa15f, but it was probably wrong.

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/hunspell/en_NZ.info
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/hunspell/en_NZ.info
@@ -5,7 +5,7 @@
 #
 
 fsa.dict.separator=+
-fsa.dict.encoding=utf-8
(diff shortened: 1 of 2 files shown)


danielnaber:
> I don't know if there will be some undesired effects after this change. Do we need more testing?

I think it should be fine. I suggest you manually test again on the live system after deployment.

danielnaber (approved):

AzadehSafakish:
Pedro did mention to me a few weeks ago that he needed to change the encoding for the PT dict to Latin-1 to make it work, so perhaps it's a viable solution for English as well.

My test for "quinceañera" passed; I don't foresee any major issues.

AzadehSafakish (approved):

PR #10035, opened 2024-01-03
[fr] CONFUSION_MEMES_MEME[4] adding exceptions

LucieSteib:
Fix https://github.com/languagetooler-gmbh/languagetool-premium/issues/6772.

This solution is only a partial fix, obviously.
- CONFUSION_MEMES_MEME[4] has a great apply rate above 75%
- there's not easy and obvious fix, maybe Entities in a futur try?


GillouLT (approved):
LGTM!

PR #10044, opened 2024-01-04
[en] add words 1/03

AzadehSafakish:
routine dictionary additions; PRing because I'm not sure why local mvn tests fail


danielnaber:
> I'm not sure why local mvn tests fail

`SpellingCheckRuleTest.testIgnoreSuggestionsWithMorfologik:57 » IndexOutOfBounds Index 0 out of bounds for length 0` -> you have added `Michaels`and the test in line 57 fails now because that used to be a typo that had a suggestion, not it's not a typo and thus doesn't have a suggestion

AzadehSafakish:
>you have added Michaels

I see, thank you. 
It's the name of a large arts and crafts chain in the US, as well as a common surname.
Oddly enough, I just noticed that we already have a disambiguation rule to ignore the spelling...
I will remove the test case.

PR #10046, opened 2024-01-04
[en] move dictionaries to a external dependency

jaumeortola:
These are the current dictionaries, with no changes, moved to an external dependency.


jaumeortola:
This works now. There is only a minor issue with the tagging of good/well/best. @AzadehSafakish 

I will create a diff with this branch. We cannot merge it without checking the diff.

AzadehSafakish:
@jaumeortola 

I've looked over the diffs [in this comment](https://github.com/languagetooler-gmbh/languagetool-premium/issues/6846#issuecomment-1879794493), and I don't see anything major.
I'm approving the PR.

However, to be safe, I'm not sure how I feel about merging this right before the Premium deployment.
Could it be merged _after_ the greenlight for EN deployment tomorrow, so it doesn't go out to Premium users right away?

Also, I just added some words right before reviewing this; I resolved the conflict by adding the words in `added` and `spelling` to the now empty files, but please let me know if I should revert the merge commit and move the new additions somewhere else.

AzadehSafakish (approved):

jaumeortola:
> However, to be safe, I'm not sure how I feel about merging this right before the Premium deployment.

I will merge the PR after the Premium sync.

PR #10047, opened 2024-01-05
External file for rule confidence

danielnaber:
Load rule confidence from am external file. This is limited to one confidence per match/rule, even though the match might have more than one suggestion. They will all end up with the same confidence and the client needs to interpret that...

### DIFF
--- a/languagetool-language-modules/nl/src/main/java/org/languagetool/rules/nl/CompoundAcceptor.java
+++ b/languagetool-language-modules/nl/src/main/java/org/languagetool/rules/nl/CompoundAcceptor.java
@@ -1099,7 +1099,7 @@ public class CompoundAcceptor {
     }
   }
 
-  private DutchTagger dutchTagger = DutchTagger.INSTANCE;
+  private final DutchTagger dutchTagger = DutchTagger.INSTANCE;
 
   public CompoundAcceptor() {
   }
--- a/languagetool-server/src/main/java/org/languagetool/server/HTTPServerConfig.java
+++ b/languagetool-server/src/main/java/org/languagetool/server/HTTPServerConfig.java
@@ -73,6 +73,7 @@ public class HTTPServerConfig {
   protected int textCheckerQueueSize = 8;
   protected Mode mode;
   protected File languageModelDir = null;
+  protected File ruleIdToConfidenceFile = null;
   protected boolean pipelineCaching = false;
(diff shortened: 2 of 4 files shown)


languagetool-language-modules/nl/src/main/java/org/languagetool/rules/nl/CompoundAcceptor.java
```diff
@@ -1099,7 +1099,7 @@ public class CompoundAcceptor {
     }
   }
 
-  private DutchTagger dutchTagger = DutchTagger.INSTANCE;
+  private final DutchTagger dutchTagger = DutchTagger.INSTANCE;
```
danielnaber, line 1102: sorry, unrelated commit

languagetool-server/src/main/java/org/languagetool/server/HTTPServerConfig.java
```diff
@@ -1503,4 +1512,10 @@ public void setDbMaxConnections(int dbMaxConnections) {
     this.dbMaxConnections = dbMaxConnections;
   }
 
+  /**
+   * @since 6.4
+   */
```
SteVio89, line 1517: Add @Nullable annotation

languagetool-server/src/main/java/org/languagetool/server/V2TextChecker.java
```diff
@@ -39,9 +45,40 @@ class V2TextChecker extends TextChecker {
...
+  private void loadConfidenceMap() throws IOException {
+    if (config.getRuleIdToConfidenceFile() != null) {
+      logger.info("Loading confidence map for rules from " + ruleIdToConfidence);
+      List<String> lines = Files.readAllLines(Paths.get(config.getRuleIdToConfidenceFile().getAbsolutePath()), UTF_8);
+      for (String line : lines) {
+        if (line.startsWith("#")) {
+          continue;
+        }
+        String[] parts = line.split(",");
+        if (parts.length >= 2) {
```
SteVio89, line 70: Why `>= 2` and not `== 2`?

danielnaber: there might be more columns for better debugging, but we don't use them here  - I'll be adding that as a comment

SteVio89 (approved):

PR #10050, opened 2024-01-05
Fr/grammar/a plein temps rule creation

LucieSteib:
We have a rule that adds a dash in "plein temps" - which is good when it's a noun.
Otherwise, which is majority of time, there should not be any, and for the moment there's no rule doing so. 

It is a known difficulty of French: https://www.larousse.fr/dictionnaires/francais/plein-temps/61603#difficulte.


LucieSteib:
@GillouLT merging now (it's temp_off anyway) so I can see the diff next week :)

PR #10051, opened 2024-01-05
[all][fr] Dictionnary additions

LucieSteib:
Adding names to fr dictionnary files
One addition also to spelling_global. 
(after https://github.com/languagetool-org/languagetool/commit/9b4f348ee5fee61ad5282a144e5dabc8f0d474cc)

### DIFF
--- a/languagetool-core/src/main/resources/org/languagetool/resource/spelling_global.txt
+++ b/languagetool-core/src/main/resources/org/languagetool/resource/spelling_global.txt
@@ -20402,3 +20402,4 @@ Horst Görtz
 SteamOS
 WW International
 Jens Rommel
+David Copperfield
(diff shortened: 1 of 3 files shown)


jaumeortola (approved):

languagetool-core/src/main/resources/org/languagetool/resource/spelling_global.txt
```diff
@@ -20401,4 +20401,5 @@ Bear Stearns
 Horst Görtz
 SteamOS
 WW International
-Jens Rommel
\ No newline at end of file
+Jens Rommel
+David Copperfield
```
jaumeortola, line 20405: It's okay. Now it is a duplicate, and it will be removed when we check for duplicates.

jaumeortola (approved):

Issue #9819, opened 2023-12-04
Swedish grammar - dates


walseb:
Hello!

The rules for Swedish grammar appear quite rudimentary.

One simple rule that's missing is for week days. Week days should be written with uncapitalized letters.

Source:
http://dinsvenska.se/stor-bokstav/ska-manader-och-veckodagar-inledas-med-stor-eller-liten-bokstav/

Just thought I would add this issue as I don't know the syntax for adding new rules.

danielnaber:
Indeed, nobody is working on LT's support for Swedish currently. If you'd like to contribute, here's a documentation about how to write simple rules: https://dev.languagetool.org/development-overview

ljo:
I have some rules coming for this. I just got stuck with some dictionary extension and remodelling during the autumn. Please assign it to me.

PR #10072, opened 2024-01-09
Adding rules for names of weekdays and months which fixes request in issue #9819.

ljo:
Adding rules for names of weekdays and months as first part of (many) rules for SSK chapter 8 'Stor eller liten bokstav?'. Fixes request in issue #9819.


danielnaber:
Thanks!

PR #10088, opened 2024-01-11
Remove misspelled ai suggestions

jaumeortola:
This can be quite drastic, but it seems to be the desired outcome: remove all matches with misspelled suggestions.


jaumeortola:
Do we need to add rules QB_EN.* to the filter? @AzadehSafakish @maphjo 

@fabrichter Can you review the PR? Any potential problems? We would like to know the exact content of the configuration.

fabrichter (approved):

maphjo:
> Do we need to add rules QB_EN.* to the filter?

I am unsure at the moment because we know that the en-US spelling dictionary is not very complete (good suggestions could be lost), unless @AzadehSafakish or @evan-defran-lt say that QB_EN provides a lot of nonsense suggestions.
But we can add QB_EN also in a second step.

PR #10097, opened 2024-01-15
[de] DE_MULTITOKEN_SPELLING_TWO_4 

Luisa-LT:
Improvements (added words  + exceptions + examples) + changed messages

Diff: https://regression.languagetoolplus.com/via-http/2024-01-15/de-DE_full/result_grammar_DE_MULTITOKEN_SPELLING_TWO%5B4%5D.html

@jaumeortola could you please review the changes (especially the exceptions added to the rules)? Thanks!

### DIFF
--- a/languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/multitoken-suggest.txt
+++ b/languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/multitoken-suggest.txt
@@ -110,6 +110,7 @@ Hackescher Markt
 Hackeschen Markts
 Hackeschen Marktes
 Hackeschem Markt
+Hackeschen Markt
 Rosenthaler Platz
 Rosenthaler Platzes
 Hamburgischen Staatsoper
@@ -129,7 +130,7 @@ Kleene'schen Hüllen
 New Yorkerin
 New Yorkerinnen
 San Martín
-Schimon Stein
+Shimon Stein
 Truman Plaza
 Union Berlin Stadion
 Musée de la Mode et du Textile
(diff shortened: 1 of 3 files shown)


languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/multitoken-suggest.txt
```diff
@@ -129,7 +130,7 @@ Kleene'schen Hüllen
 New Yorkerin
 New Yorkerinnen
 San Martín
-Schimon Stein
+Shimon Stein
```
jaumeortola, line 133: In de.wikipedia.org, Schimon Stein. In en.wikipedia.org, Shimon Stein. Different transcriptions? Or a misspelling?

jaumeortola: Now I noticed that you keep both.

jaumeortola:
It looks good to me.
Some comments: 
* The problems with the characters `©|\+|TM` should be solved with a change in the tokenizer or in the speller. That would be a different issue.  
* About the exception `.*s` for genitive. There is already an exception in the filter for the genitive. It works for "Stephen Hawkings" because "Stephen Hawking" is in the dictionary, but not for "Stephan Hawkings". I don't have a solution for that. Maybe we could generate a suggestion with the s: Stephan Hawkings ⇾ Stephen Hawkings (?).

Luisa-LT:
Thank you for your comments and approval @jaumeortola! Perhaps we can have a talk about the tokenizer regarding special characters later. Regarding the genitive -s: I was already speaking with @AgnesKleinhans that we want to ask you for help to add the line expanders /E, /N, /S, etc. also to the multitoken files (like in spelling and ignore.txt). Let's move this to slack :-) But yes, for the moment, I will also add Stephen Hawkings to the spelling file :-)

maphjo:
> * The problems with the characters `©|\+|TM` should be solved with a change in the tokenizer or in the speller. That would be a different issue.
--> this is already mentioned in https://github.com/languagetooler-gmbh/languagetool-premium/issues/6787
> * About the exception `.*s` for genitive. There is already an exception in the filter for the genitive. It works for "Stephen Hawkings" because "Stephen Hawking" is in the dictionary, but not for "Stephan Hawkings". I don't have a solution for that. Maybe we could generate a suggestion with the s: Stephan Hawkings ⇾ Stephen Hawkings (?).
--> maybe add to the issue or create a new one for the umbrella topic.

PR #10100, opened 2024-01-16
[es] add rule AI_ES_HYDRA_LEO_CP_SITILDE_SI

BellaDiekmann:
@jaumeortola: Adding the subrule 21 from SI_TILDE_PREMIUM with its antipatterns as antipattern to AI_ES_HYDRA_LEO_CP_SITILDE_SI hopefully fixes the loop between LOOP by AI_ES_HYDRA_LEO_CP_SITILDE_SI and SI_TILDE_PREMIUM[21] of our issue [6652](https://github.com/languagetooler-gmbh/languagetool-premium/issues/6652)

### DIFF
--- a/languagetool-language-modules/es/src/main/resources/org/languagetool/rules/es/remote-rule-filters.xml
+++ b/languagetool-language-modules/es/src/main/resources/org/languagetool/rules/es/remote-rule-filters.xml
@@ -751,6 +751,74 @@ To ignore a remote rule match, set the <marker> so that it exactly covers the te
         </rulegroup>
         <rulegroup id="AI_ES_HYDRA_LEO_CP_SITILDE_SI" name="">
             <rule>
+                <antipattern>
+                    <token inflected="yes" skip="7" regexp="yes">preguntar|decir</token>
+                    <token>que</token>
+                    <token>si</token>
+                </antipattern>
+                <antipattern>
+                    <token>que</token>
+                    <token skip="-1">si</token>
+                    <token>?</token>
+                </antipattern>
+                <antipattern>
+                    <token>si</token>
+                    <token min="0" max="2" postag="PP3..A.*|PP[123]C[SP][0D]00|P0[123].*" postag_regexp="yes"/>
+                    <token skip="-1" postag="V.I.*" postag_regexp="yes"/>
+                    <token>,</token>
+                    <token postag="V.N.*" postag_regexp="yes"/>
+                </antipattern>
+                <antipattern>
+                    <token skip="10" regexp="yes">más|menos|tantos?|tantas?</token>
+                    <token>que<exception postag="LOC_CS"/></token>
+                    <token>si</token>
+                </antipattern>
+                <antipattern>
+                    <token skip="2" regexp="yes">más|menos|tantos?|tantas?</token>
+                    <token skip="10">que<exception postag="LOC_CS"/></token>
+                    <token>si</token>
+                </antipattern>
+                <antipattern>
+                    <token regexp="yes">igual|vaya|toma</token>
+                    <token>que</token>
+                    <token>si</token>
+                </antipattern>
+                <antipattern>
+                    <token regexp="yes">qué|que</token>
+                    <token>tal</token>
+                    <token>si</token>
+                </antipattern>
+                <antipattern>
+                    <token>en</token>
+                    <token>cuanto</token>
+                    <token>a</token>
+                    <token>que</token>
+                    <token>si</token>
+                </antipattern>
+                <pattern>
+                    <token postag="CS|LOC_CS|PR.*" postag_regexp="yes" regexp="yes">que|porque|pues|aunque</token>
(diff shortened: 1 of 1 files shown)


jaumeortola:
Okay. This could work. But we need to modify si->sí in this pattern. 

We want to prevent the rule AI_ES_HYDRA_LEO_CP_SITILDE_SI from making the change sí -> si. We need a pattern with "sí".

jaumeortola:
This is what we want to avoid: 
image

BellaDiekmann:
> Okay. This could work. But we need to modify si->sí in this pattern.
> 
> We want to prevent the rule AI_ES_HYDRA_LEO_CP_SITILDE_SI from making the change sí -> si. We need a pattern with "sí".

Of course! Sorry, I still need to get used to the reversed logic. I'll take care of it tomorrow.

BellaDiekmann:
@jaumeortola: I replaced "si" with "sí" in the pattern of the rule, but left "si" in the antipatterns. Is that okay or do I still get it wrong?

jaumeortola:
Hmm. The logic is difficult to grasp. I would say that we need "sí" (with accent) in the APs. If the model wants to change sí → si in those cases, it is right, and we would rather not prevent the match.

BellaDiekmann:
> Hmm. The logic is difficult to grasp. I would say that we need "sí" (with accent) in the APs. If the model wants to change sí → si in those cases, it is right, and we would rather not prevent the match.
> 
> We could even change the "correct" examples (si -> sí) because we want the model to detect an error. We could check if our rules or models detect errors in those sentences.

Aaah, I see. Okay, let's try that.

jaumeortola:
>  Is it different for punctuation?

No. It is not different. What we did for the comma rule is incomplete.

BellaDiekmann:
@jaumeortola: Maybe you could check the last commit whether the logic is coherent now. 

I replaced "si" with "sí" in the antipatterns and corresponding example sentences for the new rule added to AI_ES_HYDRA_LEO_CP_SITILDE_SI. 

I also removed the comma from the antipatterns and corresponding example sentences added to AI_ES_HYDRA_LEO_MISSING_COMMA. 

In the case of example sentences with parenthesis, I removed the comma between the noun and the verb since the antipatterns come from a rulegroup that corrects wrong commas between subject and predicate (e.g.: "el objetivo, dice, es salvar, vidas." -->  "El objetivo dice, es salvar, vidas."). 

For the same reason, I only removed the first comma of this antipattern,: `<antipattern><token>,</token><token>es</token><token regexp="yes">lógico|claro|obvio|evidente</token><token>,</token></antipattern>`

I was not sure which comma to remove in this example sentence, though: "Los usuarios, además, pueden publicar comentarios en las historias." What do you think?

languagetool-language-modules/es/src/main/resources/org/languagetool/rules/es/remote-rule-filters.xml
```diff
@@ -459,25 +454,25 @@ To ignore a remote rule match, set the <marker> so that it exactly covers the te
...
+                <example>Louis vienen tus amigos.</example>
+                <example>Tom hay algo que quiero darte.</example>
+                <example>Mira se le rompió la fuente a esa señora.</example>
+                <example>Tom te llama tu mamá.</example>
+                <example>Mary venga, deja de protestar.</example>
+                <example>Ricardo ora por mi.</example>
+                <example>María me gustan mucho las enumeraciones.</example>
+                <example>Tom te llama mamá.</example>
+                <example>Claro le duele.</example>
+                <example>Tom no es necesario que te disculpes.</example>
```
jaumeortola, line 475: Okay. This means: if the model tries to add a comma to these sentences, the rule in remote-rule-filter.xml will not prevent the model from doing it.

languagetool-language-modules/es/src/main/resources/org/languagetool/rules/es/remote-rule-filters.xml
```diff
@@ -809,14 +804,14 @@ To ignore a remote rule match, set the <marker> so that it exactly covers the te
...
-                <example>Vaya que si ha sido dramático.</example>
+                <example>Dijo que sí puedo detener una Señal, que la cadena se romperia</example>
+                <example>Le digo a La Heredera que sí podemos jugar juntas.</example>
+                <example>de modo que sí observamos alguna incomodidad o alguna reacción exagerada, poder parar para atender sus necesidades</example>
+                <example>Prestaban más atención que sí gritaba sin parar. </example>
+                <example>Es más agradable a la vista que sí lo tachas concienzudamente.</example>
+                <example>Son menos que las posibles opciones sí aplicamos las 27 letras del alfabeto.</example>
+                <example>No es igual que sí tienen algo que ocultar.</example>
+                <example>Vaya que sí ha sido dramático.</example>
             </rule>
```
jaumeortola, line 815: The same. If the model tries to change sí -> si in these sentences, the rule in remote-rule-filter.xml will not prevent the model from doing it.

jaumeortola (approved):

PR #10107, opened 2024-01-17
[de] refactor ignorePotentiallyMisspelledWord

affemitkaraffe:


languagetool-language-modules/de/src/test/java/org/languagetool/rules/de/GermanSpellerRuleTest.java
```diff
@@ -81,24 +81,24 @@ public void testIgnoreMisspelledWord() throws IOException {
     assertTrue(rule.ignorePotentiallyMisspelledWord("Wölkchenbildung"));
     assertFalse(rule.ignorePotentiallyMisspelledWord("Abschlussgruße"));  // probably "...grüße"
     assertTrue(rule.ignorePotentiallyMisspelledWord("Offenlegungsfrist"));
-    assertFalse(rule.ignorePotentiallyMisspelledWord("Offenlegungsfirst"));
-    assertFalse(rule.ignorePotentiallyMisspelledWord("Dachfrist"));
+    //assertFalse(rule.ignorePotentiallyMisspelledWord("Offenlegungsfirst"));
+    //assertFalse(rule.ignorePotentiallyMisspelledWord("Dachfrist"));
```
AgnesKleinhans, line 85: Dass diese beiden rausfallen ist schade, aber nach der Logik valide compounds 👍

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2247,80 +2260,65 @@ protected boolean ignorePotentiallyMisspelledWord(String word) throws IOExceptio
...
-    }
+  }
+
+  private boolean isValidWordLength(String word) {
+    return word.length() <= MIN_WORD_LENGTH || word.length() >= MAX_WORD_LENGTH;
+  }
+
+  private boolean isValidPartLength(String part1, String part2) {
     // don't assume very short parts (like "Ei") are correct, these can easily be typos:
-    if ((hasInfixS || part1.endsWith("s")) && part1.length() >= 4 /* includes 's' */ && part2.length() >= 3 && startsWithLowercase(part2)) {
```
AgnesKleinhans, line 2297: Warum fällt das hier raus? Dadurch, dass nicht geprüft wird, ob part2 bereits mit uppercase reingegeben wird, funktioniert in der GermanSpellerRuleTest.java dieser case nicht mehr: 
`assertFalse(rule.ignorePotentiallyMisspelledWord("HaltungsSchäden"));`

AgnesKleinhans (changes requested):

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
affemitkaraffe: Dort habe ich letzte Woche aufgehört. Da es bisher keine systematische Lösung für Bindestriche gibt, habe ich solche Fälle vorerst weggelassen. Denn abhängig davon, ob es einen Bindestrich gibt, können Komposita korrekt sein ("Haltungs-Schäden" vs "HaltungsSchäden")

affemitkaraffe: Es können natürlich striktere Bedingungen an die Bindestrichsetzung gelegt werden, aber das besprechen wir ja noch

affemitkaraffe: CamelCase für zweiteilige Komposita failt wieder

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2247,80 +2266,76 @@ protected boolean ignorePotentiallyMisspelledWord(String word) throws IOExceptio
...
-      ) {
-      System.out.println("compound: " + part1 + " " + part2 + " (" + word + ")");
-      return true;
-    }
-    return false;
+    return part1.length() >= 3 && part2.length() >= 4;
+  }
+
+  private String removeInfixS(String part1) {
+    return part1.replaceFirst("s?$", "");
```
p-goulart, line 2330: ❓ This method deletes the **first** instances of a string-final `s` (or does nothing). Does it account for words that end in an \<s\> that is **not** a Fugen-S? As it stands, it seems like it'd transform `Pass` into `Pas`, right?

Specifically with nouns ending in \<ss\>, this may be frequent enough that you could be wasting a lot of time investigating this branch, if you know what I mean? Maybe assume much earlier that nouns that end in \<ss\> don't have a Fugen-S?

That's not an issue **in itself** if the method is only called once you have 100% made sure that the string you're passing has a Fugen-S, but you might at least write a comment here to make its working crystal clear, lest future people try to use it to remove Fugen-S and end up with \*`Pas` (not a Fugen-S) or `Rechtsanwalt` (Fugen-S not removed).

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2247,80 +2266,76 @@ protected boolean ignorePotentiallyMisspelledWord(String word) throws IOExceptio
...
+  }
+
+  private String removeHyphen(String part1) {
+    return part1.replaceFirst("-?$", "");
+  }
+
+  private String removeInfixSAndHyphen(String part1) {
+    return removeInfixS(removeHyphen(part1));
```
p-goulart, line 2338: 🥜 Two replaces, when you could just be replacing `s?-?$` with `""` once in a single call. The performance cost is negligible (so from that point of view it's a nitpick), but do note that if there's a problem with `removeHyphen`, `removeInfixS` may not apply as expected.

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2339,6 +2354,231 @@ private boolean isOnlyPluralNoun(String word) throws IOException {
...
+  private boolean processTwoPartCompounds(String part1, String part2) throws IOException {
+    // Expects two parts of a compound and checks
+    //  if their POS tags are correct,
+    //  if an infix s is missing or
+    //  if an infix s is incorrect
+    String part1uc = uppercaseFirstChar(part1);
+    String part2uc = uppercaseFirstChar(part2);
+    String part1WithoutHyphen = removeHyphen(part1);
+    boolean part2ucIsNoun = isNoun(part2uc);
+    boolean part2ucIsMisspelled = isMisspelled(uppercaseFirstChar(part2uc));
```
p-goulart, line 2366: 🥜 `uc` is not really transparent, maybe just use `upcased` for better readability.

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2339,6 +2354,231 @@ private boolean isOnlyPluralNoun(String word) throws IOException {
...
+    }
+    if (isNounNom(part1) && isVerbStem(part2) && isNoun(compound2)) {
+      // e.g. "Wein" + "kühl" + "schrank"
+      return true;
+    }
+    return false;
+  }
+
+  private boolean needsSometimesInfixS (String part1, String part2) throws IOException {
+    // Check if *part1* requires or prohibits infix s depending on *part2*
```
p-goulart, line 2431: ❓ The name of the method and the descriptive comment are a little confusing to me.

If you pass `Arbeit` and `platz`, how does it **sometimes** need a Fugen-S?

I understand what the method **does**, but I think the name and the comment (as well as the comment when it's actually called) don't really make it apparent to someone reading the code immediately.

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2339,6 +2354,231 @@ private boolean isOnlyPluralNoun(String word) throws IOException {
...
+      return true;
+    }
+    if (part1.equals("Rechts") && RECHT.matcher(lowercaseFirstChar(part2_lemma)).matches()) {
+      // e. g. "Rechtsfertigung"
+      return true;
+    }
+    return false;
+  }
+
+  private String findLemmaForNoun(String word) throws IOException {
```
p-goulart, line 2456: 🥜 I don't know much about Java, so it **could** very well be that this is the preferred way of doing this, but I'd think just making this method `Nullable` instead of returning an empty string for the lemma would be more efficient? I could be wrong, though.

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2339,6 +2354,231 @@ private boolean isOnlyPluralNoun(String word) throws IOException {
...
+      }
+    }
+    return lemma;
+  }
+
+  private boolean isValidCamelCase(String input) {
+    // Check if the string contains any instance of camel casing
+    boolean containsCamelCase = input.matches(".*(\\p{Ll}\\p{Lu}|\\p{Lu}{2,}\\p{Ll}).*");
+
+    // Add exceptions here
```
p-goulart, line 2471: ❓ Should we be expecting exceptions?

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2339,6 +2354,231 @@ private boolean isOnlyPluralNoun(String word) throws IOException {
...
+    List<AnalyzedTokenReadings> readings = getTagger().tag(singletonList(uppercaseFirstChar(word)));
+    for (AnalyzedTokenReadings reading : readings) {
+      if (reading.hasPosTagStartingWith("SUB")) {
+        lemma = reading.getReadings().get(0).getLemma();
+      }
+    }
+    return lemma;
+  }
+
+  private boolean isValidCamelCase(String input) {
```
p-goulart, line 2467: 🔍 Is it possible that this is already implemented in `StringTools::isCamelCase`? Is there a reason you can't use that one?

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
```diff
@@ -2339,6 +2354,231 @@ private boolean isOnlyPluralNoun(String word) throws IOException {
...
+    // Check if the string contains any instance of camel casing
+    boolean containsCamelCase = input.matches(".*(\\p{Ll}\\p{Lu}|\\p{Lu}{2,}\\p{Ll}).*");
+
+    // Add exceptions here
+
+    return !containsCamelCase;
```
p-goulart, line 2473: 🥜 If you **do** keep this method, instead of banging the boolean, maybe just `return !input.matches("...")` directly instead?

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
affemitkaraffe: You're right. I adopted this naming convention when I started working on this task^^

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
affemitkaraffe: Didn't know this, but will leave it that way for the time being as a precaution

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
affemitkaraffe: I mean, it's German. There are exceptions all the time^^'

p-goulart: Oh, you mean exceptions to the rule. I thought we were expecting some exception handling, I mean, some error to be thrown.

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/GermanSpellerRule.java
affemitkaraffe: Renamed the method to `removeTrailingS`

p-goulart (approved):
✨ 🌈

mark-baas (approved):
Looks good to me!

PR #10108, opened 2024-01-17
SentenceAnnotator: use a sentence ID

jaumeortola:


languagetool-http-client/src/main/java/org/languagetool/remote/SentenceAnnotator.java
```diff
@@ -115,14 +115,11 @@ private static void runAnnotation(AnnotatorConfig cfg) throws IOException, NoSuc
         break;
       }
       String[] partsLine = line.split("\t");
-      String sentence;
-      String sentenceHash = "Undefined";
+      String sentence = partsLine[0];
```
p-goulart, line 118: In the new GEMA files the `SentenceID` is the *first* column, not the second. So it should be the other way around.

languagetool-http-client/src/main/java/org/languagetool/remote/SentenceAnnotator.java
```diff
@@ -115,14 +115,11 @@ private static void runAnnotation(AnnotatorConfig cfg) throws IOException, NoSuc
...
+      String sentenceID = "N/A";
+      String sentenceHash = md5FromSentence(sentence);
       if (partsLine.length == 2) {
-        sentence = partsLine[1];
-        sentenceHash = partsLine[0];
-      } else {
-        sentence = partsLine[0];
-        //sentenceHash = md5FromSentence(sentence);
+        sentenceID = partsLine[1];
```
p-goulart, line 122: In the new GEMA files the `SentenceID` is the *first* column, not the second. So it should be the other way around.

p-goulart (commented):
Just to be sure, won't we also need to do something about line 173:
```java
        case "r":
          sentence = line;  // here?
          fpMatches.clear();
          cfg.outStrB = new StringBuilder();
          break;
```

languagetool-http-client/src/main/java/org/languagetool/remote/SentenceAnnotator.java
jaumeortola: This is the input, not the output.

According to https://github.com/languagetooler-gmbh/languagetool-premium/issues/6978#issuecomment-1893440159
> SentAnn is able to accept, as input, a TSV with an ID as the second column;

languagetool-http-client/src/main/java/org/languagetool/remote/SentenceAnnotator.java
jaumeortola: Again. This is the input. 

I will change the output. 

Is the input correct as it is? Sentence in the first column, ID in the second one.

languagetool-http-client/src/main/java/org/languagetool/remote/SentenceAnnotator.java
p-goulart: Oh, blimey, well, I misspoke in that comment... As you can see from the [actual files](https://github.com/languagetooler-gmbh/GEMA/blob/main/Annotations/PT/Source/24A.tsv) I posted last week, the ID is in the first column. Sorry about the confusion.

p-goulart: As for the output, the [ERMA code](https://github.com/languagetooler-gmbh/errant-for-gema/pull/57/files) assumes the `SentenceID` is the second column, after the hash.

p-goulart (approved):

PR #10124, opened 2024-01-19
[en] amended hyphenated compound rule cream-colored

evan-defran-lt:
Requesting your review on this, as I changed `<token regexp="yes" chunk_re="I-NP-.*">colou?red</token>` to `<token regexp="yes" postag="JJ">colou?red</token>`. 

For some reason, our chunker chunks `coloured` as `B-VP` in the sentence _The cream coloured jumper looked particularly smart._

Not sure if `JJ` instead of `I-NP-.*` is going to have any unintended consequences...


AzadehSafakish (approved):
I don't see a problem, should be fine 🥞

Issue #10113, opened 2024-01-18
Unpaired brackets rule does not detect duplicate errors


FredKruse:
The rule does not detect duplicate errors if the opening character could also be a closing character. In German these are ", « and ». The error does not occur if opening and closing characters are unique.

Examples:
Guten Morgen", sagte sie. "Schön, dass sie da sind.    <-- NOT detected
Guten Morgen«, sagte sie. »Schön, dass sie da sind.    <-- NOT detected
Guten Morgen“, sagte sie. „Schön, dass sie da sind.     <-- detected

FredKruse:
Although the changes made resolve the above-mentioned error, the UnpairedBracketsRule has a fundamental problem with the handling of quotation marks. The rule treats all characters like parentheses, which can be nested any number of times. Quotation marks normally cannot be nested. There are usually only two levels here: double quotes and single quotes. This basically leads to mishandling if the number of opening and closing characters is the same. Example: “Text1 “Text2” Text3” would be recognized as correct. But two errors should be displayed.
​ @danielnaber: 
My suggestion: I write a GenericUpairedQuotesRule specifically related to handling quotes only. I could install them for German and English and adapt the German and English BracketRules accordingly (handle only brackets). But I don't think I can do so with the other languages.

danielnaber:
@FredKruse Feel free to work on this, but please don't activate anything without a lot of testing (and only via PR, so others can have a look). My time for testing and supporting with code review is also very limited for the time being.

PR #10129, opened 2024-01-19
UnpairedBracketsRule: solves issue #10113

FredKruse:
The request solves issue  #10113


danielnaber:
Could you also add a test case for the change?

PR #10167, opened 2024-01-22
[de] added words to various fiels + multitoken APs + disambig rules

Luisa-LT:


languagetool-core/src/main/resources/org/languagetool/resource/spelling_global.txt
```diff
@@ -20889,4 +20889,13 @@ Tico Tico
 Tico Tico no Fubá
 Tico-Tico no Fubá
 Schott Music
-Georges Catoire
\ No newline at end of file
+Georges Catoire
```
Luisa-LT, line 20892: @jaumeortola I added some words to the global-spelling.txt - just FYI, please let me know if any changes are needed

languagetool-language-modules/de/src/main/resources/org/languagetool/rules/de/grammar.xml
```diff
@@ -83767,6 +83768,17 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
             </rule>
         </rulegroup>
         <rulegroup id="DE_MULTITOKEN_SPELLING_TWO" name="Rechtschreibfehler in Eigennamen (2 Token)">
+            <antipattern>
+                  <token postag="UNKNOWN"/>
+                  <token postag="UNKNOWN" regexp="yes">.+ies</token>
+                  <example>Wir mögen alle Short Stories, Smart Cities und die Westchester Counties.</example>
```
jaumeortola, line 83774: Instead of a particular exception, wouldn't it be better to add Smart Cities to the lists of words (besides Smart City)?

languagetool-language-modules/de/src/main/resources/org/languagetool/rules/de/grammar.xml
```diff
@@ -83801,7 +83813,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
                   <example>Sirius B ist der lichtschwache Begleiter von Sirius A.</example>
                 </antipattern>
                 <antipattern>
-                  <token postag="UNKNOWN">São</token>
+                  <token postag="UNKNOWN" regexp="yes">São</token>
                   <token postag="UNKNOWN"/>
                   <example>São Francisco</example>
```
jaumeortola, line 83818: `São Francisco` could be in global_spelling.txt. There are many places named `São Francisco`, and it cannot be confused with other langauges.

languagetool-language-modules/de/src/main/resources/org/languagetool/rules/de/grammar.xml
Luisa-LT: On the one hand, it's true, because in the end all the words that are now ignored by the multitoken rule due to the antipattern are still not accepted by the speller. On the other hand, I created an "ignore_spelling" rule in the disambiguation file to avoid speller matches. I hope that works. Nevertheless, adding them to spelling is a good idea.

languagetool-language-modules/de/src/main/resources/org/languagetool/rules/de/grammar.xml
Luisa-LT: Yes, thank you! I will add it.

PR #10174, opened 2024-01-23
[ast] new rules, updating hunspell licence dates and adding updated hunspell files

esbardu:
With this pull request I add new rules in Asturian `grammar.xml` file, and I update Hunspell files with the last version. I also adapt create_dict.sh file from [de] Daniel Naber in order to get a binary copy of spelling tool (not yet achieved to date).
It is my first contribution in GitHub, I hope I'm doing it properly. I beg your pardon if not.


esbardu:
I've just seen the warning from ci/circleci. I've fixed some detected typos, but the rest of fails are grammatically correct. They are all of kind "errors expected: 1 - errors found: 0". They occur when is needed to swich two words by only one word (contractions). I don't know how to make it work. What does the script expects in this lines?

danielnaber:
I don't have time to debug this right now, but I think `infllected="yes"` cannot be used with Asturian because it doesn't have a synthesizer. You can interactively test your rules at https://community.languagetool.org/ruleEditor/expert, so you don't need to wait for the tests to run.

esbardu:
Great, thanks Daniel! This evening I'll check the rules. Asturian has a POStagger, but is very incomplete. Perhaps errors are because of not finding some words into it? I'm constructing a new POStagger, but meanwhile perhaps it is better not to write rules that use the tagger, isn't it?

esbardu:
I found the origin of the errors:
1) I'm using the new POStag flags with the current POStag file (has different flags and configuration).
2) Current POStag file has not ALL inflections of the verbs.
Both issues caused to fail the check. For the moment I've fixed postag item to match the old flags, and put in quarantine the three rules that uses inflections of verbs not inflected until new POStagger is ready and operative.
I upload all changes.

danielnaber:
Just to be sure: the license files are still correct? I'm asking because the changes are quite big.

Issue #10182, opened 2024-01-24
[en] Rule assumes, first letter is pronounced as a vowel, but despite being one, it's pronounced like a consonant


shaedrich:
The example even lists the opposite case "an hour", which indicates, that it's not the literal letter, that's important, but the actual pronounciation is `ˈwidʒə`, so it actually is pronounced starting with a `ˈw`, which is a consonant.

![grafik](https://github.com/languagetool-org/languagetool/assets/11225821/2989e6b2-0d45-4499-adb4-b2812dabab36)

evan-defran-lt:
Hi @shaedrich, thanks for flagging this to us. I've added an exception to the rule. Please allow 24-36 hours before you see the fix implemented in LT.

shaedrich:
Thanks for the quick fix! 👍🏻

PR #10183, opened 2024-01-24
[en] add exception EN_A_VS_AN

evan-defran-lt:


evan-defran-lt:
https://github.com/languagetool-org/languagetool/issues/10182

PR #10191, opened 2024-01-26
[fr] Moving rules from grammar to style

LucieSteib:
Discussions with the WV french data team led to move these rules from grammar to style.


GillouLT (approved):
LGTM!

Issue #7013, opened 2022-08-12
[ua]: Vulgar language suggested 


MarvinErdner:
Can we remove the first suggestion, please? 

Text seems to be in Russian, not in Ukrainian. 

![image (6)](https://user-images.githubusercontent.com/88431679/184366946-81209483-f24b-4dcf-93d0-7ec1026ba45f.png)

PR #10212, opened 2024-01-29
[fr] PLACE_ADJ 2 AP added

LucieSteib:
To fix [#7013](https://github.com/languagetooler-gmbh/languagetool-premium/issues/7013)

> La marque grand public née à Roubaix en 1984 est brutalement placée en liquidation judiciaire

Added grand public as AP, chances of FN coming from that are low.


GillouLT (approved):
Nice! LGTM :)

PR #10233, opened 2024-01-31
new rule GenericUnpairedQuotesRule - applied to English and German - …

FredKruse:
The changes solve issue #10113.
A new rule "GenericUnpairedQuotesRule" was created to resolve issue #10113. The "GenericUnpairedBracketsRule" has not been changed, but the English and German rules have been modified to separate the bracket rules from the quotation mark rules.
The same was done with the test cases.
All tests involving quotation marks that had to satisfy "UnpairedBracketsRules" are now also satisfied by the "UnpairedQuotesRules" (especially the special cases for apostrophe and inch symbol). Additionally, the cases described in issue #10113 were included in the testing.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/GenericUnpairedQuotesRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/GenericUnpairedQuotesRule.java
@@ -158,6 +158,10 @@ public class GenericUnpairedQuotesRule extends TextLevelRule {
   private boolean isOpeningQuote(AnalyzedTokenReadings[] tokens, int i) {
     for (int j = 0; j < startSymbols.size(); j++) {
       if (startSymbols.get(j).equals(tokens[i].getToken())) {
+        if ((tokens[i - 1].isSentenceStart() || tokens[i].isWhitespaceBefore())
+            && (i >= tokens.length -1 || tokens[i + 1].isWhitespaceBefore())) {
+          return false;
+        }
         if (endSymbols.contains(startSymbols.get(j))) {
           return (tokens[i - 1].isSentenceStart()
               || tokens[i].isWhitespaceBefore()
@@ -175,6 +179,10 @@ public class GenericUnpairedQuotesRule extends TextLevelRule {
   private boolean isClosingQuote(AnalyzedTokenReadings[] tokens, int i) {
     for (int j = 0; j < endSymbols.size(); j++) {
       if (endSymbols.get(j).equals(tokens[i].getToken())) {
+        if ((tokens[i - 1].isSentenceStart() || tokens[i].isWhitespaceBefore())
+            && (i >= tokens.length -1 || tokens[i + 1].isWhitespaceBefore())) {
+          return false;
+        }
         return true;
       }
     }
--- a/languagetool-core/src/main/java/org/languagetool/rules/GenericUnpairedQuotesRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/GenericUnpairedQuotesRule.java
@@ -27,7 +27,7 @@ import java.util.*;
 import java.util.regex.Pattern;
 
 /**
- * Rule that finds unpaired quotes, brackets etc.
+ * Rule that finds unpaired quotes
  * 
  * @author Fred Kruse
  * @since 6.4
--- a/languagetool-core/src/main/java/org/languagetool/rules/GenericUnpairedQuotesRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/GenericUnpairedQuotesRule.java
@@ -34,9 +34,11 @@ import java.util.regex.Pattern;
  */
 public class GenericUnpairedQuotesRule extends TextLevelRule {
 
-  private static final Pattern OPENING_BRACKETS = Pattern.compile("[(\\[{]");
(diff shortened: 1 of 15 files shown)


languagetool-core/src/main/java/org/languagetool/rules/GenericUnpairedQuotesRule.java
```diff
@@ -0,0 +1,261 @@
...
+import java.util.*;
+import java.util.regex.Pattern;
+
+/**
+ * Rule that finds unpaired quotes, brackets etc.
+ * 
+ * @author Fred Kruse
+ * @since 6.4
+ */
+public class GenericUnpairedQuotesRule extends TextLevelRule {
```
danielnaber, line 35: How much of this is duplicated from `GenericUnpairedBracketsRule`? We should avoid code duplication.

FredKruse: > How much of this is duplicated from `GenericUnpairedBracketsRule`? We should avoid code duplication.

I started with a duplicate of the GenericUnpairedBracketsRule, but I had to change nearly everything. So there is no relevant duplicate code.

FredKruse: Something else: This change makes my other two pull requests unnecessary. How can I cancel it?

danielnaber: > How can I cancel it?

There's a "Close pull request" button at the bottom of e.g. https://github.com/languagetool-org/languagetool/pull/10129

FredKruse: @danielnaber 
Hi Daniel, I would like to test the new rules against the Wikipedia dataset, but the self-compiled languagetool-wikipedia-6.4-SNAPSHOT.jar doesn't work (Error: java.lang.NoClassDefFoundError: org/languagetool/tools/JnaTools). The new rules are not included in the nightly. Please give me a hint.

danielnaber: `languagetool-wikipedia-6.4-SNAPSHOT.jar` doesn't contain dependencies, but it will find them if you put it into the LT root directory and call it from there (the directory which has a `libs` folder with all the `.jar` files in it).

FredKruse:
The rules were tested in German and English against the latest wiki articles. No false positives were found. The rule was adapted to avoid false positives when testing against a Tatoeba data set (German).
Note: The rule is very sensitive to the use of mixed quotation mark symbols (e.g. mix between English and German quotation marks). But that's intentional (and was also in the GenericUnpairedBracketsRule).

danielnaber:
Is there debugging output in the message? I get `Zeichen ohne sein Gegenstück: ‚„(e2)‘ scheint zu fehlen` for: `Und da es im Englischen so schön heißt “sharing is caring” - ...`

FredKruse:
> Is there debugging output in the message? I get `Zeichen ohne sein Gegenstück: ‚„(e2)‘ scheint zu fehlen` for: `Und da es im Englischen so schön heißt “sharing is caring” - ...`

The String "(e2)" is only for debugging. I removed it in my last change. Is this an output before the last change?

danielnaber:
> The String "(e2)" is only for debugging. I removed it in my last change. Is this an output before the last change?

Yes, this wasn't the latest version, sorry.

I wonder whether we can avoid false alarms for cases where the characters are used like this:

> ...chen und steigern » Datenbankperformance steigern » Tipps zur Performance-Verbesserung

FredKruse:
> und steigern » Datenbankperformance steigern » Tipps zur Performance-Verbesserung

Did you really mean this example or:
> und steigern » Datenbankperformance steigern « Tipps zur Performance-Verbesserung

In my opinion, in a proper text, your example is a typical case of two correct alarms.

FredKruse:
Incidentally, both versions of the example generate a number of other alarms (a false blank) that overlay the Quotes rule alarms.
I wonder if we should consider this example.

FredKruse:
@danielnaber 
I just had the idea that we could interpret all characters surrounded by whitespace, not as quotation marks. 
What do you think?

danielnaber:
> I just had the idea that we could interpret all characters surrounded by whitespace, not as quotation marks. What do you think?

Sounds like a good idea.

danielnaber:
Sorry, I lost track of this. How many sentences has this been tested against now? I could also send you a more diverse test set than just Wikipedia + Tatoeba.

FredKruse:
> How many sentences has this been tested against now? 

I tested the English version against the English Wikipedia and the German against the German Wikipedia and a German Tatoeba dataset.
> I could also send you a more diverse test set than just Wikipedia + Tatoeba.

Yes, please send me that dataset.

FredKruse:
Based on the tests with the extended data set, I have included exceptions that primarily have an effect in descriptions of functions (missing spaces between characters). As far as I can see, given the huge data set with plenty of errors, there are no more false positives. However, the function (like the previous Brackets rule) is sensitive to the use of incorrect but very similar characters for the quotation marks.
I think the rule should be merged now.

PR #10313, opened 2024-02-14
[nl] accept specific EN words in PT

mark-baas:


p-goulart (approved):

p-goulart:
I do want to load the second dictionary for Portuguese, though, as I don't simply want to ignore correct English words, I want to correct the wrong ones. The English dictionary was never needed for disambiguation, but for spellchecking.

PR #10317, opened 2024-02-15
[nl] accept specific EN words in FR

mark-baas:


GillouLT (approved):
Great! Looks indubitably good to me :)

mark-baas:
We'll be using another method to accomplish this, without having to load the second dictionary.

PR #10318, opened 2024-02-15
[nl] accept specific EN words in ES

mark-baas:


mark-baas:
We'll be using another method to accomplish this, without having to load the second dictionary.

PR #10320, opened 2024-02-15
[ca] ignore english words

jaumeortola:
https://github.com/languagetooler-gmbh/languagetool-premium/issues/7061

### DIFF
--- a/languagetool-language-modules/ca/pom.xml
+++ b/languagetool-language-modules/ca/pom.xml
@@ -54,11 +54,6 @@
             <artifactId>catalan-pos-dict</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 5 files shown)


mark-baas (approved):
That looks great, should we be using the same method (adding the tags to multiwords) for all languages?

jaumeortola:
> should we be using the same method (adding the tags to multiwords) for all languages?

Not necessarily. Here, we are adding "ignore spelling" in the disambiguator. The tags in multiwords.txt are not essential. The method would work the same without these tags.

p-goulart (approved):

jaumeortola:
replaced by https://github.com/languagetool-org/languagetool/pull/10335

PR #10335, opened 2024-02-20
[es,ca] ignore English words

jaumeortola:
New version of https://github.com/languagetool-org/languagetool/pull/10320
This code can be easily shared by other languages, even to ignore Spanish words in English.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/tagging/disambiguation/AbstractDisambiguator.java
+++ b/languagetool-core/src/main/java/org/languagetool/tagging/disambiguation/AbstractDisambiguator.java
@@ -72,7 +72,9 @@ public abstract class AbstractDisambiguator implements Disambiguator {
     }
     if (lang != null) {
       spellingRule = lang.getDefaultSpellingRule();
-      tagsToIgnore = tagsToBeIgnored;
+      if (tagsToBeIgnored != null) {
+        tagsToIgnore = tagsToBeIgnored;
+      }
       if (languageCode.startsWith("en")) {
         commonWords = commonEnglishWords;
       }
--- a/languagetool-core/src/main/java/org/languagetool/tagging/disambiguation/AbstractDisambiguator.java
+++ b/languagetool-core/src/main/java/org/languagetool/tagging/disambiguation/AbstractDisambiguator.java
@@ -1,6 +1,6 @@
-/* LanguageTool, a natural language style checker 
+/* LanguageTool, a natural language style checker
  * Copyright (C) 2017 Daniel Naber (http://www.danielnaber.de)
- * 
+ *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
  * License as published by the Free Software Foundation; either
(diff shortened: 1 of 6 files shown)


mark-baas (approved):

languagetool-core/src/main/java/org/languagetool/tagging/disambiguation/AbstractDisambiguator.java
```diff
@@ -32,4 +42,84 @@ public AnalyzedSentence preDisambiguate(AnalyzedSentence input) {
...
+    "later", "part", "under", "well", "being", "including", "them", "through", "area", "both", "make", "name",
+    "before", "called", "could", "fine", "people", "second", "until", "while", "will", "i'll", "against", "age",
+    "city", "go", "may", "number", "population", "season", "several", "team", "these", "work", "born", "early",
+    "family", "film", "now", "same", "series", "so", "use", "what", "album", "based", "four", "it's", "life",
+    "released", "since", "state", "began", "century", "each", "end", "following", "found", "game", "high", "located",
+    "town", "i've", "around", "because", "can't", "come", "day", "did", "didn't", "government", "group", "here",
+    "home", "large", "like", "love", "music", "my", "named", "no", "often", "system", "that's", "three", "too", "up",
+    "us", "won", "yes", "you're", "along", "built", "career", "former", "if", "include", "left", "line", "local",
+    "long", "off", "own", "place", "small", "still", "those", "took", "band", "due", "form", "held", "its", "last",
+    "major", "member", "members", "much", "don", "didn", "doesn", "wasn", "weren", "'s", "'t", "'ll");
```
p-goulart, line 62: The one thing I think we will need is language-specific lists of words (ideally kept in txts that are then read by the disambiguator), as some of these words are also valid words in other languages and shouldn't be used to make judgements about Englishiness.

p-goulart (changes requested):

languagetool-core/src/main/java/org/languagetool/tagging/disambiguation/AbstractDisambiguator.java
p-goulart: Right, I think I understand the idea. I've pulled this locally and added some Portuguese tests, and I think my scepticism about having common Portuguese function words annotated as 'common English' words is justified:

```
# Correct
Sentence: "Um bom film." // Matches: 1
[filme, fim, fila, filo, file, filma, filé, filmá, filem, filó, filá, pium, FILs, Pium, fiam, fiem, filam, filmo, mium]

# Incorrect; I expect 'film' to be corrected, but because it is between very common words 'no' and 'as', it is skipped
Sentence: "No film as meninax discutem assuntos." // Matches: 1
[menina, meninas, minimax, meninão, memina, meminas, meminha, meminhas, meninez]
```

---

Furthermore, taking the examples above, the whole point of my original idea was to then run the spellchecking for **English**, but I'm getting Portuguese suggestions. The idea was that, when in close proximity to English words, incorrect words are also probably in English. E.g.:

```
Sentence: "Ele gosta de The Ride of the Valkiries" // Matches: 1
[Valquírias]
```

Here, since the whole thing is in English, I want `Valkyries`. `The Ride of the Valquírias` makes no sense.

jaumeortola:
I did a forced push. Did I remove anything by you inadvertently? @p-goulart

languagetool-core/src/main/java/org/languagetool/tagging/disambiguation/AbstractDisambiguator.java
jaumeortola: I see. 
Obviously, we can make adjustments to this method according to the needs of each language. Anything can be overwritten. Or you can use something entirely different. 

If you want to provide English spelling suggestions, you have to do it in the Portuguese spelling rule. This will be more language-dependent. Now, you know how to access the English speller rule without adding a dependency in pom.xml.

I am going to merge this for ES and CA. I want to see the nightly diffs and what happens in production.

p-goulart (approved):

languagetool-core/src/main/java/org/languagetool/tagging/disambiguation/AbstractDisambiguator.java
p-goulart: Fair, I'm curious to see what happens. We will prob. have to use a different list of English words and work on the suggestions.

PR #10342, opened 2024-02-22
Multilingual confidence map

danielnaber:
Tthe confidence map now supports multiple languages

**NOTE** do not merge without changing `ruleIdToConfidenceFile` in the server config first -- LT would crash on start-up otherwise!


languagetool-core/src/main/java/org/languagetool/tools/ConfidenceMap.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public int hashCode() {
+      return Objects.hash(lang, ruleId);
+    }
+
+    @Override
+    public String toString() {
+      return lang.getShortCodeWithCountryAndVariant() + "/" + ruleId;
+    }
+  }
```
SteVio89, line 68: Why do we need the `ConfidenceMap` class?

SteVio89 (approved):

PR #10349, opened 2024-02-25
[fr] Dictionary additions

LucieSteib:
Adding proper names, nouns and adj to the FR dict files.
I decided to add "_homoérotique_" and "_homoérotisme_" without dash or space, to align the form with "_homosexualité_".
These words are not (yet) in the Larousse, Robert or TLF. CNRTL provides this list of words with homo- as a prefix, most of them not having a dash: https://www.cnrtl.fr/definition/homo-


GillouLT (approved):
LGTM!

PR #10352, opened 2024-02-27
[pt] Add spelling hotfixes

p-goulart:
For `email`, I'm just adding a very specific grammar rule specifically telling people to add a hyphen there, since our suggestions were maybe getting in the way:

Screenshot 2024-02-27 at 5 58 05 PM

Since this is the most frequent addition, maybe it's worth being very explicit..?

---

For `stories`, I assume that's the Instagram thing... I think we should just support it. Adding it here as a hotfix since it's in our top 10 for the past month.

---

For `EAD`, I'm not adding speller support, since it is wrong, but I am adding a super simple grammar rule to suggest `EaD` when specifically referring to 'ensino a distância`, which has higher priority than the speller.

---

Screenshot 2024-02-27 at 5 36 29 PM


susanaboatto (approved):

maphjo:
yeay! Already making use of the new dashboard. :rocket:

PR #10358, opened 2024-02-29
[en] improve EN_A_VS_AN

AzadehSafakish:
If "a" or "an" is followed by "EUR", look to the _next token_ to determine the correct determiner.
I needed to add "100" to `det_a.txt` to make the 3rd example work:
`assertIncorrect("The company has also entered into an EUR 100 million ($88 million) debt");`

Not sure why I had to do this, but I do notice that:
>The company has also entered into an 100 million USD debt

is not caught by this particular rule either.
Maybe there's a blind spot with certain digits.

### DIFF
--- a/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/AvsAnRule.java
+++ b/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/AvsAnRule.java
@@ -95,6 +95,11 @@ public class AvsAnRule extends Rule {
       }
       if (equalsA || equalsAn) {
         Determiner determiner = getCorrectDeterminerFor(token);
+        if (token.getToken().equals("EUR") && i < tokens.length - 1) {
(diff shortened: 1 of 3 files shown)


languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/AvsAnRule.java
```diff
@@ -95,6 +95,11 @@ public RuleMatch[] match(AnalyzedSentence sentence) {
       }
       if (equalsA || equalsAn) {
         Determiner determiner = getCorrectDeterminerFor(token);
+        if (token.getToken().equals("EUR") && i < tokens.length - 1) {
+          determiner = getCorrectDeterminerFor(tokens[i + 1]);
+        } else if (token.getToken().equals("EUR") && i == tokens.length - 1) {
```
danielnaber, line 100: There's no test for this case, or am I missing something?

PR #10366, opened 2024-03-05
[fr] ignore English words

jaumeortola:
A local regression test is to be done before merging.


LucieSteib:
Ah, cool! Excellent move, thanks @jaumeortola! 
If this pass; we should maybe include this commit to the weekly deploy, what do you think?

jaumeortola:
> If this pass; we should maybe include this commit to the weekly deploy, what do you think?

No. We still need to create a new diff to be certain of the results. And it is better to start with OS only.

LucieSteib:
Ah okay I though you might have done a local diff, I don't know why 🤦‍♀️ all good then, curious to see the results 😍

GillouLT:
Just pushed changes deleting words that could be problematic as discussed in https://github.com/languagetooler-gmbh/languagetool-premium/issues/7080#issuecomment-2009547578

PR #10369, opened 2024-03-06
fr: make example non-offensive

kudkuda:
Make the shortest example (shown in JetBrains IDE tooltips) non-offensive

https://youtrack.jetbrains.com/issue/IDEA-330919/Dont-show-sexist-examples-in-the-grammar-checker-tooltips


GillouLT:
That's great!

PR #10375, opened 2024-03-07
[pt] Use new POS tagging schema

susanaboatto:
⚠️ This change **requires** the [portuguese-pos-dict](https://github.com/languagetool-org/portuguese-pos-dict) dependency to be `v0.13` or later!

## tl;dr

This PR adds POS tags to the `pt` tagset for enclitic pronouns.

## What changes

Let's use `diz-me` ('tell me') to illustrate the changes.

|was|will be|
|---|---|
|three tokens (`diz`, `-`, `me`)|a single token (`diz-me`)|
|each token tagged separately|tagged as `dizer[VMM02S0:PP1CSO00]`|

### Consequences
- dictionary suggestions consider whole word forms (e.g. `ama-se` <=> `amasse`);
- forms that only appear in derived environments are no longer accepted by the dictionary in isolation (e.g. `amá` and `lo` from `amá-lo` should now both be flagged as spelling errors);
- working with mesoclitics is made significantly easier (e.g. `fá-lo-á` doesn't require **five** tokens, **three** of which shouldn't exist in isolation!);
- XML rules no longer require two extra (often optional) tokens for the hyphen and pronouns.

## Also in this PR

- tokenisation logic changes from defining breaking characters to defining word characters;
- extensive changes to XML rules in all dialects to account for new tokenisation and tagging schema;
- `EnclisisFilter` and `ProclisisFilter` to make synthesising verb forms with pronouns easier.

### DIFF
--- a/languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/MorfologikPortugueseSpellerRule.java
+++ b/languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/MorfologikPortugueseSpellerRule.java
@@ -197,6 +197,24 @@ public class MorfologikPortugueseSpellerRule extends MorfologikSpellerRule {
     return wordMap;
   }
 
+  private boolean isValidCliticVerb(String word) {
+    // Because of differences between the tagger and the speller when it comes to verbs with clitics, and the fact
+    // that, theoretically, it is possible for dialect issues to emerge, we need to check if the flagged word is
+    // actually a *valid* verb form (e.g. pt-PT struggles with 'diz-se').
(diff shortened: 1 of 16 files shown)


p-goulart:
@marcoagpinto I'm pinging you here so you're aware of it.

After this PR goes through, you should be able to go back to editing rules and the dictionary freely. Thank you for your patience!

p-goulart (approved):

jaumeortola (approved):

p-goulart:
@susanaboatto I hope you've had a chance to have a look. Since it is Friday, I want to wait until Monday to merge this to avoid surprises over the weekend.

PR #10378, opened 2024-03-08
[ast] dictionaries in external dependency

jaumeortola:
Everything seems okay here. We can do the merge. @esbardu


esbardu:
Thanks a lot for all work here, @jaumeortola. From now on, every time I want to update files (or have a group of changes to update them all together), I understand that I'll have to do it in a branch of my `ast-pos-dict` repository and, if everything goes well, I'll merge it with the main branch of my repository, right? And is this change automatic for LanguageTool or does a new release have to be made in Sonatype? In the latter case, should I contact you to make the update?

jaumeortola:
> does a new release have to be made in Sonatype?

A new release in Sonatype is needed. 

Usually, you do this:
- build a new binary file
- build a new version of the dependency and test it locally (changing the version in LanguageTool, but not committing yet)
- if all tests pass, publish the new version in Sonatype
- commit the new version number to LT

You can contact me to publish the new version on Sontaype. If you want to be more independent, you can open your account.

PR #10384, opened 2024-03-11
[fr] add exception for nouns in V.* tags - Update

Sharcoux:
Update based on the comments here: https://github.com/languagetool-org/languagetool/pull/9875

Basically, `postag="V.*"` was matching Nouns as well, so I added an exception to remove those from the detection.

For instance:
- **Classe** de neige: this one is incorrectly marked as Verb, triggering the rule for no reason.
- **Menace** de mort: this one could legitimately be both a verb or a noun, and we don't want to trigger the rule in such a case


Sharcoux:
The German CI failed? o0

My change doesn't change that part.

Sharcoux:
@jaumeortola I updated the description of this one too, added some examples and updated the title of the PR to match the patterns. If the CI succeeds, I think that it can be merged too.

PR #10385, opened 2024-03-11
Fix detection of nouns and verbes that are preceeded with "tu"

Sharcoux:
Update based on the discussion here: https://github.com/languagetool-org/languagetool/pull/9875

Fix the following issues in the current state:
- LanguageTool incorrectly detects some nouns to be a verb, like "Actes de naissance". This should probably be improved elsewhere too, but I don't have the competency to do it myself. Thus, in the current state, this patch is needed.
- The rule was marking regular usage of _2nd pers ind_ as wrong. This fix is essential for the rule to work as expected.


Sharcoux:
The German CI seems broken. There are no relation with this PR

Sharcoux:
I suggest that we keep track of this PR here.

I'll update the PR description to describe better the case handled here

Sharcoux:
I updated the description and added 2 examples to clarify the rule.

jaumeortola:
Okay. Let's see what happens in the diffs.

PR #10387, opened 2024-03-12
[fr] message for ggec rules that change only the casing

jaumeortola:
If the suggestion only changes the casing of the original, then adapt the message and other attributes of the rule or the match.
For testing, we could build a fake match and test it.


jaumeortola:
@GillouLT: Even better if you can try it locally.

GillouLT:
Thank you!! Tested locally with a fake rule as suggested by @jaumeortola, and it works :) I'll make some changes to the PR and merge later in the context of https://github.com/languagetooler-gmbh/languagetool-premium/issues/7072

PR #10392, opened 2024-03-12
[fr] Fix FP in CONFUSION_ER_E_PAR2

jaumeortola:


Sharcoux:
I suggest to close this MR and track to the benefit of https://github.com/languagetool-org/languagetool/pull/10385

Issue #10373, opened 2024-03-06
academic abbreviation


GuShaocheng:
Hi,

I run Languagetool in a docker.  I found it treats the following words (bold parts) as wrong spelling.

The article studies cyt. **c** in S. **cerevisiae**.

In fact, "cyt. c" and " S. cerevisiae" are correct words. I have many other abbreviation. May I ask a way to deal with this issue.
Thank you very much!

evan-defran-lt:
Hi @GuShaocheng, we will soon recognize some 800 Linnaean taxonomic abbreviations: https://github.com/languagetool-org/languagetool/commit/76580be18140f14d1eaf368e87078e36742116fa

I will also go ahead and add `cyt. c` to our English dictionary.

You can also add individual words to your personal dictionary. Simply click on the dictionary icon in the upper-righthand corner of the error card when the spelling suggestion pops up. This will add the word to your personal dictionary. 
<img width="332" alt="Screenshot 2024-03-19 at 15 15 32" src="https://github.com/languagetool-org/languagetool/assets/112640985/495a31ab-9c5c-4df7-b2e2-7091bd7cdf28">

We are always looking to improve our dictionaries, so if you have any suggestions, please let us know.

PR #10414, opened 2024-03-18
Add abbreviated scientific names to global spelling

p-goulart:
I noticed we accept full scientific names (e.g. `Helicobacter pylori`) but not the abbreviated ones (e.g. `H. pylori`), which are common in scientific literature. This PR addresses that.


LucieSteib:
OMG this is great ⭐

p-goulart:
(cf. [this post](https://languagetooler.slack.com/archives/C03UT0HD6CB/p1710778134237019?thread_ts=1710776506.122819&cid=C03UT0HD6CB) from devOps about the Circle failure)

evan-defran-lt:
See also this issue: https://github.com/languagetool-org/languagetool/issues/10373

jaumeortola:
We had some problems with short words with abbreviations `spp` or `sp`. For example: Shorea spp., Shorea sp., Agave spp., and so on. The solution was to move these names to disambiguation-global.xml so that they were accepted by the spelling rules but not used as suggestions.
If there are problems with the names in this PR, we could do the same.

jaumeortola (approved):

PR #10417, opened 2024-03-19
[pt] Add OQ_O_QUE_ORTHOGRAPHY rule

p-goulart:
Hotfix for `oq`. As you can see, it is the 10th most frequent add event, and both gGEC and our own speller fail to provide the appropriate correction.

---

Screenshot 2024-03-19 at 10 16 59 AM

Screenshot 2024-03-19 at 10 17 56 AM

Screenshot 2024-03-19 at 10 20 05 AM


susanaboatto (approved):

p-goulart:
@susanaboatto up to you whether you want to include this commit in the premium deployment; it's obviously not urgent.

PR #10437, opened 2024-03-25
Add prio field to rule

SteVio89:


fabrichter (changes requested):
Nitpick: I would have preferred to use priority instead of prio for consistency.

In the English and Catalan Language classes, we overwrite getRulePriority without calling super; so we either need to fix that or overwrite getPriorityForId instead (where they call the super method).

fabrichter (approved):

PR #10448, opened 2024-03-29
[fr] AI_FR_GGEC_MAIL_EMAIL

GillouLT:
To address the last item in: https://github.com/languagetooler-gmbh/languagetool-premium/issues/6700#issuecomment-2026809498 (see the last comment)
And part of: https://github.com/languagetooler-gmbh/languagetool-premium/issues/7072

Trying to join join matches in French.Java

WIP: Adding tests in `JLanguageToolTest.java`


GillouLT:
Adding tests in `JLanguageToolTest.java` is not working out, not even for AI_FR_GGEC_QUOTES which is currently working well in production, I might be doing something wrong, I'll double-check with @jaumeortola 

I'm merging the PR to see the changes live in the next deploy, the behavior is correct when running the main test with fake local rules.

PR #10456, opened 2024-04-01
[It] Repeat Rule: add passo and piano

albertuan:


languagetool-language-modules/it/src/main/java/org/languagetool/rules/it/ItalianWordRepeatRule.java
```diff
@@ -40,12 +40,18 @@ public String getId() {
 
   @Override
   public boolean ignore(AnalyzedTokenReadings[] tokens, int position) {
-    if (wordRepetitionOf("via", tokens, position)) {
```
albertuan, line 43: Moved down, for alphabetic order

PR #10477, opened 2024-04-09
[pt] Fix ACENTUAÇÃO_VOGAL_ÊNCLISE

p-goulart:
Some FPs in `vê-lo`, `lê-lo`, etc.


p-goulart:
Messed up the branch name, deleting. Check #10478 instead.

PR #10491, opened 2024-04-11
[pt] Fixed FP in rule ID:SER_CONSISTIR_RESIDIR

marcoagpinto:

I have fixed a FP in the rule after seeing the nightly results:
https://internal1.languagetool.org/regression-tests/via-http/2024-04-10/pt-PT/result_style_SER_CONSISTIR_RESIDIR%5B1%5D.html

I know both of you are terribly busy with the new verbs scheme, so now I will work on other rules during the weekend, when I will also try to add words to the dictionaries.

❤️


susanaboatto (approved):

marcoagpinto:
Thanks!

PR #10500, opened 2024-04-11
disable some French rules if model is available

jaumeortola:


LucieSteib:
Ah, that's so cool to see!

fabrichter (approved):

fabrichter:
It's not relevant yet with the hard-coded set of rule IDs.

PR #10510, opened 2024-04-15
[es] add expressions to multiwords.txt

BellaDiekmann:


BellaDiekmann:
@jaumeortola: Como es la primera vez que añado algo a multiwords.txt quería asegurarme de que todo estaba bien. ¿Podrías echar un vistazo al PR? ¡Gracias!

languagetool-language-modules/es/src/main/resources/org/languagetool/resource/es/multiwords.txt
```diff
@@ -11220,3 +11220,25 @@ ya será menos	LOC_ADV
...
 óp. cit.	Y
 ópera prima	NCFS000
+Fase ON;NCFS000
+Fases ON;NCFP000
+Fase OFF;NCFS000
+Fases OFF;NCFP000
+Fase "on";NCFS000
+Fases "on";NCFP000
+Fase "off";NCFS000
+Fases "off";NCFP000
```
jaumeortola, line 11230: "fase" and "fases" should be in lower-case. It will work for both capitalized and uncapitalized forms. 

But there are different kinds of quotation marks. So it would be better to do this in disambiguation.xml, to ignore the spelling of ON/OFF in several contexts, not only with "fase".

jaumeortola (changes requested):

languagetool-language-modules/es/src/main/resources/org/languagetool/resource/es/multiwords.txt
BellaDiekmann: Thank you! Can fase ON/OFF without quotation marks stay?

jaumeortola: They are not needed. This will cover most cases: 
https://github.com/languagetool-org/languagetool/commit/5cfa97c3741331ab44a011b3c9b58cd163764593

Unless we provide a suggestion, I think we have to accept these words, even in more contexts.

jaumeortola (approved):

PR #10512, opened 2024-04-17
Spelling: lower priority for short words from user dicts

jaumeortola:


p-goulart (approved):
lgtm ✨

btw, I was looking into this and **testing** the speller using custom words from the user dictionary is actually not that trivial... I do think adding a test for this would be ideal, but yeah...

PR #10514, opened 2024-04-17
[pt] Fix SENT_END bug with PortugueseProclisisFilter

p-goulart:
Fix bug described [here](https://languagetooler.slack.com/archives/C02R689CQ1W/p1713361355114399).

That filter was being too naïve about verb tags. Now it's more restrictive. Added an example to test it — should work.

@SteVio89 & @susanaboatto — I suppose you can decide if this is worth a redeploy.

### DIFF
--- a/languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/PortugueseProclisisFilter.java
+++ b/languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/PortugueseProclisisFilter.java
@@ -21,7 +21,7 @@ public class PortugueseProclisisFilter extends RuleFilter {
     HashSet<String> suggestions = new HashSet<>(Collections.emptyList());
     for (AnalyzedToken at : encliticVerbTokenReadings.getReadings()) {
       String posTag = at.getPOSTag();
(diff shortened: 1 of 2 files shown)


languagetool-language-modules/pt/src/main/java/org/languagetool/rules/pt/PortugueseProclisisFilter.java
```diff
@@ -23,8 +21,10 @@ public RuleMatch acceptRuleMatch(RuleMatch match, Map<String, String> arguments,
     HashSet<String> suggestions = new HashSet<>(Collections.emptyList());
     for (AnalyzedToken at : encliticVerbTokenReadings.getReadings()) {
       String posTag = at.getPOSTag();
+      if (posTag == null || (!posTag.startsWith("V") && !posTag.contains(":"))) {
```
jaumeortola, line 24: Maybe OR instead of AND here to require both conditions?
>  if (posTag == null || !posTag.startsWith("V") || !posTag.contains(":")) {

p-goulart: Agreed, done at 42e0f0bd6f1

jaumeortola (approved):

PR #10518, opened 2024-04-19
[fr] IS_URL-postag-fix

GillouLT:
The diff shows new matches with URLs (FPs) since `IS_URL` has been added to the disambiguation file yesterday, this PR is a quick fix for that.


LucieSteib:
Thank you Gilles 😍

PR #10519, opened 2024-04-20
[pt] Created rule ID:QUE_SER_ESTAR_PARTPASSADO_V2 to replace rule ID:QUE_SER_ESTAR_PARTPASSADO

marcoagpinto:

After some six days working on it almost non-stop, it is ready.

I have created a rule to replace the obsolete rule ID:QUE_SER_ESTAR_PARTPASSADO .

This rule is a lot more accurate than the obsolete one it replaces.

The code looks huge, but that happens because of the commented examples (thus why I asked the other day how many examples were acceptable, and Pedro told me not to worry about aesthetics).

The code of this new rule is very simple and both of you can easily fix FPs by changing the last two antipatterns.

So, here is some useful information:

```
rule ID:QUE_SER_ESTAR_PARTPASSADO (old rule)
Portuguese (Portugal): 593 total matches
Portuguese (Portugal): 809459 total sentences considered
```
[0_ALL_FROM_OLD_RULE_ORIGINAL.txt](https://github.com/languagetool-org/languagetool/files/15047455/0_ALL_FROM_OLD_RULE_ORIGINAL.txt)

```
rule ID:QUE_SER_ESTAR_PARTPASSADO_V2 (new rule)
Portuguese (Portugal): 410 total matches
Portuguese (Portugal): 809459 total sentences considered
```
[_129.txt](https://github.com/languagetool-org/languagetool/files/15047457/_129.txt)

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -17098,6 +17098,410 @@ USA
         </rule>
 
 
+        <rule id='QUE_SER_ESTAR_PARTPASSADO_V2' name="Simplificar: Que + ser/estar + part. passado → Particípio passado" type='style' tags='picky' default="temp_off">
+            <!-- The rule has been completely rewritten.
+                 In my first rewrite, I used postag='VMI[IPS]3.+' for the verb "ser" and "estar".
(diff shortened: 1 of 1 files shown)


p-goulart (approved):
The rule itself looks okay. But all of those negative examples... for only *a single* positive example of the rule actually matching..? But anyway, I suppose we can see how it behaves.

PR #10529, opened 2024-04-23
[pt] Check English words in speller

p-goulart:
@susanaboatto I've sent you on Slack a sample diff showing the performance of the speller after tagging English words with `_english_ignore_`. If you need a different test set, do let me know.

@jaumeortola this logic seems to work for me. The one annoying thing about it is that, since we need the English tagger for the filter to work, I cannot add Java tests without making English a dependency. Any ideas? I'd really like to add some proper tests here.


jaumeortola:
@p-goulart You can add tests to https://github.com/languagetool-org/languagetool/blob/master/languagetool-standalone/src/test/java/org/languagetool/JLanguageToolTest.java
That's what we have done in other languages.

susanaboatto (approved):

jaumeortola (approved):

PR #10542, opened 2024-04-25
[LO extension / Standalone] add the possibility to define more than o…

FredKruse:
The change creates the ability to define more than one custom option per rule.
All existing rules with custom options have been adjusted. The options panel has been expanded accordingly.
The API and the HTTP server (ConfigurationInfo) have been adapted. Upward and downward compatibility between versions lower than 6.5 and higher than 6.4 was ensured.
A second option was added for the GermanFillerWordsRule as a test case.
All changes were tested.
To-do: Some more rules should be expanded to include additional options.
A merge can be performed.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/RuleOption.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/RuleOption.java
@@ -1,0 +1,81 @@
+/* LanguageTool, a natural language style checker 
+ * Copyright (C) 2005 Daniel Naber (http://www.danielnaber.de)
+ * 
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+package org.languagetool.rules;
+
+/**
+ * Class to configure rule option by option panel
+ * @author Fred Kruse
+ * @since 4.2
+ */
+public class RuleOption {
+  private final Object defaultValue;
+  private final String configureText;
+  private Object minConfigurableValue;
+  private Object maxConfigurableValue;
+  
+  public RuleOption(Object defaultValue, String configureText, Object minConfigurableValue, Object maxConfigurableValue) {
+    this.defaultValue = defaultValue;
+    this.configureText = configureText;
+    if (minConfigurableValue != null && maxConfigurableValue != null && 
+        (defaultValue instanceof Integer || defaultValue instanceof Character || 
+            defaultValue instanceof Float || defaultValue instanceof Double)) {
+      this.minConfigurableValue = minConfigurableValue;
+      this.maxConfigurableValue = maxConfigurableValue;
+    } else {
+      this.minConfigurableValue = 0;
+      this.maxConfigurableValue = 100;
+    }
+  }
+  
+  public RuleOption(Object defaultValue, String configureText) {
+    this(defaultValue, configureText, 0 , 100);
+  }
+
+  /**
+   * Get a default Value by option panel
+   */
+  public Object getDefaultValue() {
+    return defaultValue;
+  }
+
+  /**
+   * Get the Configuration Text by option panel
+   */
+  public String getConfigureText() {
+    return configureText;
+  }
+  /**
+   * Get the minimum of a configurable value
+   */
+  public Object getMinConfigurableValue() {
+    return minConfigurableValue;
+  }
+
+  /**
+   * Get the maximum of a configurable value
+   */
+  public Object getMaxConfigurableValue() {
+    return maxConfigurableValue;
+  }
+
+
+
+  
+  
+}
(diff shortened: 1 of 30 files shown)


languagetool-gui-commons/src/main/java/org/languagetool/gui/Configuration.java
```diff
@@ -1466,13 +1466,38 @@ private void parseUnderlineTypes(String typessString, Map<String, Short> underli
...
+          if (ruleAndValue.length != 2) {
+            throw new RuntimeException("Could not parse rule and value, colon expected: '" + ruleToValue + "'");
+          }
+          String[] values = ruleAndValue[1].split(";");
+          Object[] objects = new Object[values.length];
+          for (int i = 0; i < values.length; i++) {
+            String str = values[i].trim();
+            char c = values[i].charAt(0);
+            str = str.substring(1);
+            if (c == 's') {
```
danielnaber, line 1483: A comment with examples would be nice here

languagetool-core/src/main/java/org/languagetool/rules/RuleOption.java
```diff
@@ -0,0 +1,81 @@
...
+   * Get the maximum of a configurable value
+   */
+  public Object getMaxConfigurableValue() {
+    return maxConfigurableValue;
+  }
+
+
+
+  
+
```
danielnaber, line 80: empty lines

languagetool-office-extension/src/main/java/org/languagetool/openoffice/LORemoteLanguageTool.java
```diff
@@ -309,12 +310,37 @@ public void disableCategory(CategoryId id) {
...
     for (String rule : rules) {
-      String ruleValueString = rule + ":" + configurableValues.get(rule);
-      ruleValues.add(ruleValueString);
+      Object[] obs = configurableValues.get(rule);
+      if (obs != null && obs.length > 0) {
+        StringBuilder sbOV = new StringBuilder();
+        char c;
+        for (int i = 0; i < obs.length; i++) {
+          Object o = obs[i];
+          if (o instanceof Integer) {
```
danielnaber, line 323: Is this code duplicated from above? Can we avoid the duplication?

languagetool-office-extension/src/main/java/org/languagetool/openoffice/LORemoteLanguageTool.java
```diff
@@ -478,6 +505,48 @@ static class RemoteRule extends Rule {
...
+        ruleOptions = new RuleOption[rOptions.size()];
+        for (int i = 0; i < rOptions.size(); i++) {
+          String defaultType = (String) rOptions.get(i).get("defaultType");
+          String configureText = (String) rOptions.get(i).get("configureText");
+          Object defaultValue;
+          Object minConfigurableValue;
+          Object maxConfigurableValue;
+          if (defaultType.equals("Integer")) {
+            defaultValue = (int) rOptions.get(i).get("defaultValue");
+            minConfigurableValue = (int) rOptions.get(i).get("minConfigurableValue");
```
danielnaber, line 521: `minConfigurableValue` etc. should probably be a constant, as it is used several times

languagetool-office-extension/src/main/java/org/languagetool/openoffice/LORemoteLanguageTool.java
```diff
@@ -590,6 +648,47 @@ static class RemoteTextLevelRule extends TextLevelRule {
         setOfficeDefaultOff();
       }
       isDictionaryBasedSpellingRule = ruleMap.containsKey("isDictionaryBasedSpellingRule");
+      if (ruleMap.containsKey("ruleOptions")) {
+        Object o = ruleMap.get("ruleOptions");
+        @SuppressWarnings("unchecked")
+        List<Map<String,Object>> rOptions = (List<Map<String, Object>>) o;
+        ruleOptions = new RuleOption[rOptions.size()];
+        for (int i = 0; i < rOptions.size(); i++) {
```
danielnaber, line 656: This code is duplicated?

languagetool-server/src/main/java/org/languagetool/server/TextChecker.java
```diff
@@ -732,16 +732,37 @@ private Map<String, Integer> getRuleMatchCount(List<CheckResults> res) {
...
+        } else if (c == 'f') {
+          objects[i] = Float.parseFloat(str);
+        } else if (c == 'd') {
+          objects[i] = Double.parseDouble(str);
+        } else if (c == 'c') {
+          objects[i] = str.charAt(0);
+        } else if (c == 'i') {
+          objects[i] = Integer.parseInt(str);
+        } else {  //  compatible to old version 
+          objects[i] = Integer.parseInt(values[i]);
```
danielnaber, line 762: This is duplicated?

danielnaber (changes requested):

danielnaber:
@fabrichter Maybe you can quickly check whether the change in `UserConfig` is harmless?

FredKruse:
The requested changes was done

fabrichter:
@danielnaber 
For UserConfig, the change to an Array probably breaks `.equals` and thus pipeline caching because equals on arrays checks for reference equality. But this most likely isn't relevant in practice, since for the use case where pipeline caching is enabled, configurableRuleValues is always empty. So I think we don't need to change this, but a comment about it in `.equals` would be good to add.

I would also add a better getter for the configuration values like this:
```java

  public <T> T getConfigValueByID(String ruleID, int index, Class<T> clazz, T defaultValue) {
    Object value = configurableRuleValues.get(ruleID);
    if (value == null) return defaultValue;

    if (index >= value.length) return defaultValue;
    if (!clazz.isInstance(value[index])) return defaultValue;

    return (T) value;
  }

```

This should be used instead of the existing manual accesses and checks to inner values of `configurableRuleValues`, as the checks are repetitive and not done consistently as-is.

FredKruse:
The requested changes from @fabrichter and @danielnaber was done.

PR #10546, opened 2024-04-26
[pt-PT] Added AP to rule ID:FORMAL_T-V_DISTINCTION

marcoagpinto:

It is just an antipattern to fix FPs.


p-goulart:
@marcoagpinto you can safely merge simple `pt-PT`-only changes after the tests pass. Of course, if you would still like a review, I'll gladly have a look, but I feel like in this particular kind of pt-PT PR that might not be super necessary.

p-goulart (approved):

PR #10547, opened 2024-04-26
[pt-PT] Removed 'temp_off' from rules ID:AUTO-FALANTE and CRIAR_QUERER

marcoagpinto:
Hi @susanaboatto and @p-goulart ,

It is time to start removing 'temp_off' from the rules.

The two rules here:
AUTO-FALANTE has only one hit: https://internal1.languagetool.org/regression-tests/via-http/2024-04-25/pt-PT_full/result_grammar_AUTO-FALANTE%5B1%5D.html

CRIAR_QUERER has zero hits.

Thanks!


p-goulart:
For this kind of very simple `pt-PT`-only change, I think you can just go ahead and merge once the tests pass.

PR #10549, opened 2024-04-27
[pt] Added AP to rule ID:CONCORDANCIA_DETERMINANTES_MASCULINOS

marcoagpinto:

I was revising my thesis with the LibreOffice add-on and spotted a FP (all rules turned on).

Then, after I already had a solution (AP) planned, I noticed that the rule was “temp_off”, nevertheless, here is an AP for it.


p-goulart (approved):
LGTM; 'mesmo' is a pesky one 😩

PR #10596, opened 2024-05-15
Add some new rules to en-GB grammar.xml.

Tang567:


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/en-GB/grammar.xml
```diff
@@ -534,6 +534,55 @@ USA
...
+        <rule id="WORD_COLLOCATION_2" name="Inappropriate Word Collocation">
+            <pattern>
+                <token>fast</token>
+                <token>meal</token>
+            </pattern>
+            <message>Consider using <suggesiton>fast food</suggestion> instead of 'fast meal'. "Fast food" is a specific type of food that is prepared and served quickly, whereas "fast meal" is not commonly used.</message>
+            <example correction="fast food">Let's grab some fast meal.</example>
+        </rule>
+
+        <rule id="Singular_Subject_Verb_Agreement" name="Singular Subject-Verb Agreement">
```
evan-defran-lt, line 574: LT already has rules for subject verb agreement. Additionally, as currently written, this rule will result in some false positives, like in the sentence, `He let his dog run free in the field.`
If you want to refine this rule, you should find instances where LT does not successfully correct the issue (i.e., False Negatives) and refine our already-existing rules for those cases.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/en-GB/grammar.xml
```diff
@@ -534,6 +534,55 @@ USA
...
+            <pattern>
+                <token postag="DT" min="0" max="1"/>
+                <token postag="NN">dog</token>
```
evan-defran-lt, line 577: This rule will apply only when the subject is `dog` and the verb is `run`. Could you think of ways to make it more robust? (Just flagging this for you. As noted above, LT already has rules for subject-verb agreement issues, so it would be preferable to refine those rules than create new ones from scratch.)

pom.xml
```diff
@@ -31,6 +31,7 @@
         </developer>
     </developers>
     <modules>
+        <module>languagetool-client-example</module>
```
evan-defran-lt, line 34: Could you let me know why you added this here?

evan-defran-lt (changes requested):
Hi @Tang567, thanks for contributing to LT! 

I've reviewed your PR, and I've left some comments for you. Some things will need to be amended before this PR could be approved. 

Some other things to note: 
- none of these rules are specific to en-GB, so these rules would have more impact in the general EN file: `languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml`. 
- please make sure all new rules have the attribute `default="temp_off"`. This ensures that rules do not go into production before we've had the chance to test them fully.
- before committing, it's good practice to run a Maven test in your IDE terminal to ensure that there are no errors that would break the build. Here's the command for English: `mvn --projects languagetool-language-modules/en --also-make clean test`. 
- I'm not sure if you already know this, but you can also test your rules in LT's "expert" RuleEditor here: https://community.languagetool.org/ruleEditor/expert?devMode=true. It will run your rule over a large corpus of sentences, allowing you to see how your rule performs, which means you can easily identify TPs and FPs. It will also flag some errors that would break the build. It's not perfect – you cannot test rule groups with it, for instance – but it is a helpful resource to start!

Thank you again for contributing to LT. If you have any questions, please let me know. 
Evan

PR #10600, opened 2024-05-16
[pt] Improve behaviour of compound measurements

p-goulart:
For measurements that include multiplication signs, our disambiguator is not adding an `ignore_spelling` flag, which results in stuff like this:

Screenshot 2024-05-16 at 11 41 01 AM

From the speller dashboard:

Screenshot 2024-05-16 at 11 49 57 AM

<sup>Note that here they **look** like proper multiplication signs, but if you look at the regex and inspect the values, you'll see they are really just `x`. Probably the browser being clever when rendering stuff.</sup>

---

This PR:
- adds `X_AS_MULTIPLICATION_SIGN_IGNORE` to the disambiguator (this the most important thing);
- adds tests to the speller and tokeniser (just to be on the safe side);
- add an example to `MULTIPLICATION_SIGN` picky rule (to make sure the `x` can properly be changed to a `×` even before unspaced units);
- improves the messages for `UNITS_OF_MEASURE_SPACING` picky rule (chapter 7,103 of 'our messages all sound like `pt-PT`).

Note that, because of our current tokenisation logic, unit spacing will **not** be fixed for tokens like `500x500mm`. For it to work, one would first need to apply the `MULTIPLICATION_SIGN` suggestion. (IMO that's probably fine, since these are picky rules. The most crucial thing here was no longer flagging stuff like `5x5` and `5x5mm` as spelling errors.)

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
@@ -3473,6 +3473,13 @@
         <disambig action="ignore_spelling"/>
     </rule>
 
+    <rule id="DICE_ROLL_NOTATION">
+        <pattern>
+            <token regexp="yes">\d+d(4|6|8|10|12|20)</token>
+        </pattern>
+        <disambig action="ignore_spelling"/>
+    </rule>
+
     <rule id="NONSTANDARD_DURATION_IGNORE">
          <pattern>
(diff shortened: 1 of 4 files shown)


p-goulart:
There was a bit of a peak this week, but note how this is nothing new.

Screenshot 2024-05-16 at 11 57 50 AM

p-goulart:
4de7888 adds `DICE_ROLL_NOTATION` to ignore stuff like `1d6` (roll one 6-sided die), `3d20` (roll three 20-sided dice), etc.

susanaboatto (approved):

PR #10601, opened 2024-05-20
[pt] Fix hexadecimal number bug

p-goulart:
This PR fixes a few FPs with valid hexadecimals like `0x5A` stemming from the multiplication rule<sup>1</sup>.

To achieve that, we:

- fix a logic bug in `RegexAntiPatternFilter` (@jaumeortola);
- add a `RegexAntiPatternFilter` to `MULTIPLICATION_SIGN`;
- add a couple of hexadecimal examples to `MULTIPLICATION_SIGN`;
- add `HEXADECIMAL_NOTATION` and `OCTAL_NOTATION` rules to the disambiguator so we can reliably ignore spelling for valid hexadecimal and octal numerals<sup>2</sup>;
- add tests to the speller rule to make sure we're ignoring what we should be ignoring.

---

<sup>**1**: this is not a new issue, but the matches appear as new because of the recent improvements to the disambiguator.</sup>

<sup>**2**: I decided to only ignore spelling of C-style hexadecimal and C-inspired octal numerals (i.e. `0x...` and `0o...`, respectively). Most other notations seemed like they'd conflict with legitimate spelling issues, and I didn't want to be too aggressive.</sup>

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
@@ -3466,6 +3466,20 @@
         <disambig action="ignore_spelling"/>
     </rule>
 
+    <rule id="HEXADECIMAL_NOTATION">
+        <pattern>
+            <token regexp="yes" case_sensitive="yes">0x[0-9A-Fa-f]+</token>
+        </pattern>
+        <disambig action="ignore_spelling"/>
+    </rule>
+
+    <rule id="OCTAL_NOTATION">
+        <pattern>
+            <token regexp="yes" case_sensitive="yes">0o[0-7]+</token>
+        </pattern>
+        <disambig action="ignore_spelling"/>
+    </rule>
+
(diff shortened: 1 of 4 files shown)


jaumeortola (approved):

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
```diff
@@ -3466,6 +3466,20 @@
         <disambig action="ignore_spelling"/>
     </rule>
 
+    <rule id="HEXADECIMAL_NOTATION">
```
jaumeortola, line 3469: The 'name' attribute is missing. I think it is required in some tests.

p-goulart: This is a rule within a rulegroup, though, and the rulegroup has both a `name` and an `id` attribute. Should be fine..?

jaumeortola: Probably yes. If the tests pass, it is fine.

PR #10621, opened 2024-05-28
[pt] Added/improved APs in rule ID:MORRER_PERECER_FALECER

marcoagpinto:

My improvements in this rule yesterday removed:
https://internal1.languagetool.org/regression-tests/via-http/2024-05-27/pt-BR/
MORRER_PERECER_FALECER[1]: 76 FPs
MORRER_PERECER_FALECER[2]: 3 FPs

Now it will remove tons more.

I guess I will look at the tonight's results and tomorrow remove the "temp_off".

Thanks!

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -3650,10 +3650,10 @@ USA
                     <example>Tom está morrendo de rir.</example>
                 </antipattern>
                 <antipattern>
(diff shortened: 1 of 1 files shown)


marcoagpinto:
> LGTM, but I think this rule is a great example of why we need wordnet-style semantic categories.
> 
> Where are you getting these exceptions from? Are you using a specific corpus?

@p-goulart 

I used the trick in the wordlist of adding "12 34 56." in front of each line since:
`Portuguese (Portugal): 11546 input lines ignored (e.g. not between 10 and 300 chars or at least 4 tokens)`

Now I got more 40 000 or 50 000 valid sentences to check, and a lot more than the ones in the nightly results.

Now I can finally make deep testing of the rules, although it takes a very long time.

If next year I am able to buy an i9 computer, I will join both PT-BR and the PT-PT wordlists in one and have hundreds of thousands of more sentences to check.

Anyway, in a shorter reply: for example, if using "gato" removed FPs, then I added "cão" and "cachorro", which is one of the tricks I used in this rule.

PR #10627, opened 2024-05-29
[de] ignore english words de

jaumeortola:


St-ac-y:
@jaumeortola I think we can close this PR?

PR #10640, opened 2024-06-05
Protect remote rules from unwanted traffic

SteVio89:


fabrichter (changes requested):
There's a println statement that should be removed; if needed, you can log the information there as part of the existing request logs.

You can look into making the tests actually runnable for other people and as part of CI, by using the `TestRemoteRule` which is added like this in `org.languagetool.Language#getRelevantRemoteRules` when an ID starts with `TEST`

``` configs.stream()
      .filter(config -> config.getRuleId().startsWith("TEST"))
      .map(c -> new TestRemoteRule(this, c))
      .forEach(rules::add);
```

Instead of local files for server.properties and remote-rule.json, you can generate temporary files as part of the setup for the test suite with the required content and use those files when setting up the server.

fabrichter (approved):

PR #10647, opened 2024-06-10
Refactor LanguageDependentRuleMatchFilter.java

jaumeortola:
What are we doing here?
— use only one filter for rule matches, named LanguageDependentRuleMatchFilter.java
— each language implements the method `filterRuleMatches()`

To-do:
— adjustments on Premium


AzadehSafakish (approved):
The changes for EN look okay, but on renaming the OS method vs. moving it to Premium:

We've opted to rename it and call `super.filterRuleMatches()` in Premium, but then the [update in Premium](https://github.com/languagetooler-gmbh/languagetool-premium-modules/pull/1348) can't be tested until `filterRuleMatches()` is implemented (merged) in OS.

I suppose we can continue like this, but in hindsight do you think it would be cleaner to move the logic in `filterRuleMatches()` to Premium so we don't have to use `super`? Or do we run into this issue either way? I'm not entirely sure.

jaumeortola:
The code in `adaptSuggestions()` (now `filterRuleMatches()`)  in OS is needed for this test:
```
List<RuleMatch> matches = tool.check("Whatever their needs, we doesn't never disappoint them.");
assertEquals(matches.get(0).getSuggestedReplacements().toString(), "[n't,  never]"); 
```

GillouLT (approved):

PR #10653, opened 2024-06-13
[pt] Revert English corrections

p-goulart:
This PR restores most of the reverted changes from #10652, but, crucially, **not** the English speller calls from the PT Morfologik class. For now, I am backlogging that.

- the tests I ran locally failed to account for some things, mostly proper nouns;

- we ended up with a lot of semi-valid corrections on proper nouns, which... although technically correct, since those words *aren't* recognised by the English speller, we shouldn't be introducing;

- I *could* limit it to words starting with an uppercase letter, but that would need some further testing.

---

This means we keep `v1.1.0` of the dictionary and the new simple replace orthography rule, as well as the `ok/OK` grammar rule, but we lose the English corrections.


p-goulart:
(I'm fixing tests locally and pushing in parallel to save time, so some CircleCI tests needed to be cancelled.)

susanaboatto (approved):

PR #10657, opened 2024-06-14
[pt] Improve multi-token foreign word checks

p-goulart:
This PR contains improvements to the multi-token spelling corrections.

- now that we actively ignore possible English words with the `_english_ignore_` tag, the entity containing multi-token exceptions could be significantly simplified;
- instead of FPs with `the`, we were getting tonnes of **FNs** — e.g.:
  - _Ela lê o jornal the guardian toda manhã._
  - the two-token sequence _the guardian_ is tagged (comme il faut) with `_english_ignore_`;
  - because 'the' was present in the `exceptions_multitoken_spell` entity, 'the guardian' was **not** matched by `PT_MULTITOKEN_SPELLING_TWO`;
  - this left the user with an **uncorrected** lowercase 'the guardian'.

Removing those English words from `exceptions_multitoken_spell` ensures that we correct the letter case of global multi-token entites beginning with words like 'the' and 'my'.

A couple more improvements:
- tweaked to the logic of the `IGNORE_ENGLISH_WORDS` disambiguation rule so we no longer have words with **multiple** `_english_ignore_` tags;
- added an AP to `PT_MULTITOKEN_SPELLING_TWO` to block matching of 3+ tokens;
- added relevant examples<sup>1</sup>.

---

The results look okay — we will probably gain a few FPs, but I think that will far be outweighed by the decrease in FNs. That said, it's probably better to only merge next week, I don't want to push hotfixes at the weekend.

<sup>**1:** note that, since the pattern rule tests run **without** the English module, those examples rely on `NULL` tagging, rather than `_english_ignore_` tagging; the effect should be the same, but I did make sure to test those sentences in a live environment just to be sure they'd behave the same way.</sup>

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
@@ -43514,7 +43514,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 
     <category id="MULTITOKEN_SPELLING" name="Erros ortográficos em nomes próprios" type="misspelling" default="on">
         <rulegroup id="PT_MULTITOKEN_SPELLING_FOUR" name="erros ortográficos em nomes próprios (4 tokens)" default="on">
-            <rule>
+            <rule> <!-- #1 -->
                 <pattern>
                     <token postag_regexp="yes" postag="UNKNOWN|_english_ignore_"><exception regexp="yes">&particles_of;</exception></token>
                     <token postag_regexp="yes" postag="UNKNOWN|_english_ignore_"><exception regexp="yes">&exceptions_multitoken_spell;|&exceptions_multitoken_spell2;|&exceptions_multitoken_sensitive;</exception></token>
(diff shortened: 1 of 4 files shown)


jaumeortola (approved):

languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/grammar.xml
```diff
@@ -43735,6 +43743,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
                 <example correction="García Márquez"><marker>Garcìa Mraquez</marker>.</example>
                 <example correction="Lluís Llach"><marker>Lluís Lach</marker>.</example>
                 <example correction="SÃO PAULO"><marker>SAO PAULO</marker></example>
+                <example correction="The Guardian">Ele lê o jornal <marker>the guardian</marker> toda manhã.</example>
+                <example>Ele disse the guardian is in the garden.</example>
```
jaumeortola, line 43747: Maybe this test won't pass when testing PT alone because the 'ignore English' feature needs the English dictionary.

p-goulart: I tested this one, and the behaviour here should be the same for `NULL` and `_english_ignore_`. There were _some_ examples I could indeed not add, as they would break in the absence of English.

susanaboatto (approved):

PR #10688, opened 2024-06-26
[pt] Added AP to rule ID:VERB_QUE_É_VERB_SER

marcoagpinto:

Here is an antipattern to fix false positives.

Notice that the rule could be completely rewritten, it is in my TODO list.

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -15451,6 +15451,17 @@ USA
       Eu acho que é perigoso subir aquela montanha sozinho. → Eu acho ser perigoso subir aquela montanha sozinho.
       -->
             <antipattern>
+                <token>que</token>
+                <token>é</token>
+                <token postag='VMSF3.+' postag_regexp='yes'/>
+                <example>Agora faço o que mais gosto, que é usar o meu conhecimento nos projetos open-source.</example>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
```diff
@@ -15450,6 +15450,17 @@ USA
             <!--
       Eu acho que é perigoso subir aquela montanha sozinho. → Eu acho ser perigoso subir aquela montanha sozinho.
       -->
+            <antipattern>
+                <token>que</token>
+                <token>é</token>
+                <token postag='VMSF3.+' postag_regexp='yes'/>
```
p-goulart, line 15456: This postag is only for the 3sg future subjunctive. It will match on `que é *fizer`, but not `que é fazer`. You probably meant to use the infinitive tag, didn't you?

marcoagpinto: ohhhh… right. The original idea was “fazer”. Give me a few minutes and I will fix it.

marcoagpinto:
@p-goulart 

Fixed! 😛 😛 😛 😛 😛 😛 😛 😛

marcoagpinto:
@p-goulart 

I added it again for review, I wasn't sure if I would just do the fix and press the merge button after the checks.

p-goulart (approved):

PR #10700, opened 2024-07-02
[en] added rule {und=>and}

evan-defran-lt:


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar-l2-de.xml
```diff
@@ -223,5 +223,24 @@ USA
             </rule>
         </rulegroup>
     </category>
+    <!-- ====================================================================== -->
+    <!-- Possible typo -->
+    <!-- ====================================================================== -->
+    <category id="TYPOS" name="Possible Typo" type="misspelling">
+        <rule id="UND_AND" name="He likes water und (and) lemonade" default="temp_off">
```
maphjo, line 230: If you add `ORTHOGRAPHY` to the rule ID, the rule will be shown as a spelling rule with a red underline. The default is grammar with a yellow underline.
@evan-defran-lt

PR #10704, opened 2024-07-03
perf: various optimizations for hot spots in profiling data

fabrichter:
avoid creating regexes via Pattern.split and String.replace
 remove unneccessary lock in wordsToAdd
cache data in HunspellRule initialization

### DIFF
--- a/languagetool-language-modules/pt/src/main/java/org/languagetool/tokenizers/pt/PortugueseWordTokenizer.java
+++ b/languagetool-language-modules/pt/src/main/java/org/languagetool/tokenizers/pt/PortugueseWordTokenizer.java
@@ -162,34 +162,32 @@ public class PortugueseWordTokenizer extends WordTokenizer {
   /* Splits a word containing hyphen(-) if it doesn't exist in the dictionary. */
   private List<String> wordsToAdd(String s) {
     final List<String> l = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


languagetool-language-modules/pt/src/main/java/org/languagetool/tokenizers/pt/PortugueseWordTokenizer.java
```diff
@@ -162,34 +162,32 @@ public List<String> tokenize(final String text) {
   /* Splits a word containing hyphen(-) if it doesn't exist in the dictionary. */
   private List<String> wordsToAdd(String s) {
     final List<String> l = new ArrayList<>();
-    synchronized (this) { // speller is not thread-safe
```
SteVio89, line 165: can we remove the synchronized in `EnglishWordTokenizer` and `SpanishWordTokenizer` as well?

SteVio89 (approved):

languagetool-language-modules/pt/src/main/java/org/languagetool/tokenizers/pt/PortugueseWordTokenizer.java
fabrichter: I'll take a look; those didn't show up in the profiling data

PR #10718, opened 2024-07-11
[pt] Added numbers to ENTITY numero_por_extenso

marcoagpinto:

“milhar” and “milhares” were missing.


p-goulart:
This entity is not actually used anywhere but `pt-PT` files... where it is _redefined_. You may want to remove its redeclaration in `pt-PT/grammar.xml` and `pt-PT/style.xml` and import the entity from the proper file.

marcoagpinto:
@p-goulart 

Does that mean I can merge it?

Pedro, thank you for telling me that it is only used in pt-PT as I didn't know.

The entities to be used on pt-PT: on Saturday or Sunday, I will “import” them to pt-PT.

I am scared to do it so soon because I need you and Susana “looking over my shoulder” on this improvement.

It is too major an improvement, and I am too scared to fail.

It is better for me to dedicate a reasonable number of time to doing it.

p-goulart:
You should be free to merge it, since it should only affect `pt-PT` and no tests fail, but it won't have **any** effect until you actually import it in the relevant files. So I don't see why you would.

marcoagpinto:
I will import it to the relevant files on Saturday or Sunday.

I need to dedicate a reasonable time to it to make sure it will be well done, and you and Susana to look over my shoulder on this one.

PR #10729, opened 2024-07-13
[pt-PT] Imported entities into style.xml

marcoagpinto:

Importing the entities in styles.xml (pt-PT) doesn't break the rules, but doing the same for grammar.xml does.

Anyway, I will wait for the tests to pass and then press merge since it is pt-PT.


marcoagpinto:
All tests passed, pressed “Merge”.

Issue #10721, opened 2024-07-11
[en] For British English, the LT spell checker does not find upper-case COLOR, but it finds lower-case


MikeUnwalla:
Snapshot 2024-07-07.
Spell checker finds COLOR and VAPOR, but not CENTER.
Test sentences:
1.u) THE COLOR IS WRONG; IT'S WHITE LIKE VAPOR.
1.l) The color is wrong; it's white like vapor.
2.u) MOVE THE CONTROL STICK BACK TO THE CENTER, AND THEN MOVE IT FORWARD AGAIN.
2.l) Move the control stick back to the center, and then move it forward again.
3.u) WITH THE NOSE WHEEL AT 30 DEG. TO THE CENTER LINE, SET THE LEVER TO "UP".
3.l) With the nose wheel at 30 deg. to the center line, set the lever to "UP".
4.u) ADJUST THE CLINOMETER UNTIL THE BUBBLE IS IN THE CENTER.
4.l) Adjust the clinometer until the bubble is in the center.
5.u) ALIGN THE PISTON WITH THE CENTER OF THE SLEEVE.
5.l) Align the piston with the center of the sleeve.
6.u) SET THE CONTROLS TO THE CENTER POSITION.
6.l) Set the controls to the center position.
7.u) IF THERE ARE CRACKS IN THE CENTER PLY, REPLACE THE PANEL.
7.l) If there are cracks in the center ply, replace the panel.
8.u) REMOVE THE BOLT THAT IS FARTHEST FROM THE CENTER.
8.l) Remove the bolt that is farthest from the center.
9.u) SEND THE DEFECTIVE COVER, WITH THE OIL SAMPLES, TO THE REPAIR CENTER.
9.l) Send the defective cover, with the oil samples, to the repair center.
10.u) WHEN YOU LIFT THE AIRCRAFT ON JACKS, KEEP THE CENTER OF GRAVITY BETWEEN THESE LIMITS:
10.l) When you lift the aircraft on jacks, keep the center of gravity between these limits: 
11.u) THE MODIFICATIONS CAN CHANGE THE CENTER OF GRAVITY COORDINATES.
11.l) The modifications can change the center of gravity coordinates.
12.u) THE ROD MUST TOUCH THE CENTER OF THE STRIP.
12.l) The rod must touch the center of the strip.
13.u) THE CENTER OF GRAVITY MOVES IN RELATION TO THE LOADS ON THE WINGS.
[...]

PR #10734, opened 2024-07-14
[en] Add POS and spelling for BrE for #10721

MikeUnwalla:
All verified with https://www.collinsdictionary.com/

@jaumeortola, I added the terms to spelling-eng-GB.txt to prevent confusion (the POS and the spelling agree).


MikeUnwalla:
Sorry. My stupidity. Ignore this pull request. Most terms are already in LT. I used the wrong snapshot (refer to my comment in #10735).

PR #10742, opened 2024-07-15
[pt] Update dictionary binaries

p-goulart:
Monthly update (a little late, I know). Not much to see here, just more words. No logic improvements.


susanaboatto:
I take it that checks aren't supposed to pass yet?

susanaboatto (approved):

PR #10745, opened 2024-07-16
[de,en] rules for 'JD Vance'

tiff:


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -64615,6 +64615,16 @@ USA
             <url>https://languagetool.org/insights/post/possessive-apostrophe/</url>
             <example correction="Schindler's List"><marker>Schindlers List</marker> is a 1993 American historical drama film.</example>
         </rule>
+        <rule id="JD_VANCE" name="Correct spelling for 'J. D. Vance'">
```
evan-defran-lt, line 64618: Vance's own Senate webpage supports the "JD" spelling: https://www.vance.senate.gov/about/

tiff: Thanks! I only checked Wikipedia and some news publishers. Will remove the rule.

PR #10746, opened 2024-07-16
[de] make entries in prohibit.txt work when followed by dot

affemitkaraffe:


danielnaber:
Could you add a test case?

PR #10771, opened 2024-07-22
[en] improve NEITHER_NOR_SUPERFLUOUS_COMMA

AzadehSafakish:
For the internal feedback here:
https://languagetooler.slack.com/archives/CG7ULA9T6/p1721383643012519

>But neither deployment, nor the local development flow, is containerized yet.

I believe there are two solutions here, and I'm not sure if my change is the best one.
Either an AP to block the match for the first subrule (which is what I did), or expand the second subrule to remove the second comma (after 'flow') and write an AP to block the HydraLeo match.

Do you have a preference or another idea?


evan-defran-lt:
Honestly, my preference would be for the second suggestion. 

Leaving the sentence as `But neither deployment, nor the local development flow, is containerized yet.` would results in 2 FNs.

AzadehSafakish:
> Is that a lot of additional extra work?

Nope!
I'll fix it and re-request a review when I'm done.

evan-defran-lt (approved):
lgtm!

Issue #10770, opened 2024-07-22
[en] replace.txt for British/American words


jaumeortola:
(Related to https://github.com/languagetool-org/languagetool/issues/10721)

There is a file that suggests replacements for American -> British words in the en-GB variant: https://github.com/languagetool-org/languagetool/blob/master/languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/en-GB/replace.txt

```
automaker=carmaker
garbage=rubbish
```
But it works only partially. It works only for words that are allowed by the en-GB speller (garbage, but not automaker).
![image](https://github.com/user-attachments/assets/354d60ff-bb3f-420e-9132-6919e4906e6e)

Possible solutions: 
1. Add the American words to the British spelling dictionary. 
2. Give higher priority to the rule `EN_GB_SIMPLE_REPLACE_` (prio=-50) so that it is not hidden by `MORFOLOGIK_RULE_EN_GB` (prio=-10).

@MikeUnwalla @AzadehSafakish @evan-defran-lt

jaumeortola:
The replacement rule doesn't include inflected forms (movies -> films). We need to add them.
![image](https://github.com/user-attachments/assets/412ab027-b818-4446-b7a5-cd2e8b0859b1)

MikeUnwalla:
replace.txt is a crude solution because it does not account for POS. Refer to https://github.com/languagetool-org/languagetool/issues/10344.

replace.txt does not know about semantics. For example 'garbage' is correct BrE (https://www.ldoceonline.com/dictionary/garbage). We use 'rubbish' (not 'garbage') when the word refers to 'stuff to throw away'. Similar problem with 'stroller', which is a person who is stolling.

I suggest that you review the contents of replace.txt, and remove all entries that can be used in BrE (garbage, stroller, movie,...)

Your option 2 (higher priority for EN_GB_SIMPLE_REPLACE is a safer option than adding the AmE words to the BrE spelling dictionary.

PR #10772, opened 2024-07-22
[en] gb/us lexical variants, update rule priority

jaumeortola:
Updates suggested here: https://github.com/languagetool-org/languagetool/issues/10770


AzadehSafakish (approved):
Good catch with the priority

PR #10787, opened 2024-07-26
[pt-PT] Rewrote rule ID:SIMPLIFICAR_QUE_VERBO_A_VERBOINFINITIVO

marcoagpinto:

I have rewritten another pt-PT rule, I will merge it after the checks since it is pt-PT.


marcoagpinto:
@p-goulart @susanaboatto 

I have been rewriting the pt-PT rules since the code was written years ago, and most of it is too complicated (a mess).

PR #10788, opened 2024-07-26
[fr] Adding words

LucieSteib:
See https://app.asana.com/0/1205948788428908/1207821748298280/f


GillouLT (approved):
LGTM! We'll need to remove attigé/attigées for the next update

PR #10805, opened 2024-08-02
[de] swiss german salutation rule

St-ac-y:
A rule that deletes the comma after salutation in Swiss German (as recommended) Is there a way to implement this as style (because the German variant is not super uncommon)?

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/CategoryId.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/CategoryId.java
@@ -25,8 +25,10 @@ import java.util.Objects;
  * @since 3.3
  */
 public class CategoryId {
-  
   private final String id;
+
+  // Predefined category ids
+  public static final CategoryId PUNCTUATION = new CategoryId("PUNCTUATION");
 
   public CategoryId(String id) {
     Objects.requireNonNull(id, "Category id must not be null.");
--- a/languagetool-language-modules/de/src/main/java/org/languagetool/language/SwissGerman.java
+++ b/languagetool-language-modules/de/src/main/java/org/languagetool/language/SwissGerman.java
@@ -34,13 +34,13 @@ import org.languagetool.tagging.Tagger;
 import org.languagetool.tagging.de.SwissGermanTagger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.languagetool.rules.de.SwissGermanSalutationRule;
 
 import java.io.IOException;
 import java.util.*;
 
 @SuppressWarnings("deprecation")
 public class SwissGerman extends German {
-
   private static final Logger logger = LoggerFactory.getLogger(SwissGerman.class);
 
   @NotNull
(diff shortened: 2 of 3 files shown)


languagetool-core/src/main/java/org/languagetool/rules/CategoryId.java
```diff
@@ -25,9 +25,11 @@
  * @since 3.3
  */
 public class CategoryId {
-  
   private final String id;
 
+  // Predefined category ids
+  public static final CategoryId PUNCTUATION = new CategoryId("PUNCTUATION");
```
danielnaber, line 31: This already exists in class CategoryId**s**

languagetool-language-modules/de/src/main/java/org/languagetool/language/SwissGerman.java
```diff
@@ -91,14 +94,17 @@ public List<Rule> getRelevantRemoteRules(ResourceBundle messageBundle, List<Remo
 
   @Override
   public List<RuleMatch> filterRuleMatches(List<RuleMatch> ruleMatches, AnnotatedText text, Set<String> enabledRules) {
-    //First, use the filter in German.java
+    // First, use the filter in German.java
```
danielnaber, line 97: Please don't re-format the code in the same commit that makes the actual change

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/SwissGermanSalutationRule.java
```diff
@@ -0,0 +1,75 @@
+package org.languagetool.rules.de;
```
danielnaber, line 1: Please add the license header (see other .java files)

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/SwissGermanSalutationRule.java
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  private static final String SALUTATIONS_REGEX = String.join("|", SALUTATIONS);
+  private static final Pattern SALUTATION_PATTERN = Pattern.compile(
+    "(?i)\\b(?:"
+      + SALUTATIONS_REGEX
+      + ")\\b,\\n"
+  );
+
+  private static final Category PUNCTUATION_CATEGORY = new Category(CategoryId.PUNCTUATION, "Punctuation");
```
danielnaber, line 36: see above

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/SwissGermanSalutationRule.java
```diff
@@ -0,0 +1,75 @@
...
+
+  public SwissGermanSalutationRule(ResourceBundle messages) {
+    super(messages);
+    setCategory(PUNCTUATION_CATEGORY); // Set the PUNCTUATION category
```
danielnaber, line 40: The comment isn't needed here, as the code shows what's happening

languagetool-language-modules/de/src/main/java/org/languagetool/rules/de/SwissGermanSalutationRule.java
```diff
@@ -0,0 +1,75 @@
...
+  private static final String SALUTATIONS_REGEX = String.join("|", SALUTATIONS);
+  private static final Pattern SALUTATION_PATTERN = Pattern.compile(
+    "(?i)\\b(?:"
+      + SALUTATIONS_REGEX
+      + ")\\b,\\n"
+  );
+
+  private static final Category PUNCTUATION_CATEGORY = new Category(CategoryId.PUNCTUATION, "Punctuation");
+
+  public SwissGermanSalutationRule(ResourceBundle messages) {
```
danielnaber, line 38: Could you add a test case for this class?

danielnaber (changes requested):

danielnaber:
>  Is there a way to implement this as style (because the German variant is not super uncommon)?

You can call `setCategory(Categories.STYLE.getCategory(messages));` for that, I think

PR #10828, opened 2024-08-14
[fr] PERSOS_fix

GillouLT:
Simple, quick and safe fix for: https://app.asana.com/0/1207342200148885/1208022931754948/f

PERSOS[1] rule in the disambiguation file was changing "chéris" POS to Noun automatically, causing issues when it's used as a verb.


LucieSteib:
That's so cool!
Very elegant and efficient 🦦 ❤️

PR #10829, opened 2024-08-15
[pt] Removed "temp_off" from rule ID:ESTAR_AQUI_PARA_INF_VIR_INF

marcoagpinto:

I have removed the "temp_off".

https://internal1.languagetool.org/regression-tests/via-http/2024-08-14/pt-BR/result_style_ESTAR_AQUI_PARA_INF_VIR_INF%5B1%5D.html

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/rules/pt/style.xml
@@ -3634,7 +3634,7 @@ USA
         </rule>
 
 
-        <rule id='ESTAR_AQUI_PARA_INF_VIR_INF' name="'Estar aqui para' Inf. → 'Vir Inf.'" tone_tags='formal' default="temp_off">
+        <rule id='ESTAR_AQUI_PARA_INF_VIR_INF' name="'Estar aqui para' Inf. → 'Vir Inf.'" tone_tags='formal'>
             <pattern>
                 <marker>
                     <token inflected='yes'>estar</token>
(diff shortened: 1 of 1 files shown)


susanaboatto:
Note that in pt-BR, `vimos ajudar` sounds unnatural and would likely not be accepted, the past form `viemos ajudar` being preferred. In fact, the conjugation of this verb in the third-person plural is avidly avoided in the present tense, often being replaced with the singular `a gente` or other alternatives (likely due to confusion with the verb `ver`. I think this entire rule would sound better in pt-BR if the verb "vir" was converted into its past perfect form when the original verb "estar" is in the simple present:

`Eu estou aqui para ajudar → Vim ajudar.`

Is this also the case for pt-PT? If not, I would suggest moving this rule to the pt-PT grammar.

Also, this example:

`Nós sempre estamos aqui para ajudar. `

does not seem to work well for any given pronoun — you might want to consider adding exceptions for sentences with adverbs like `sempre`, as they imply a change of meaning when this rule is applied.

marcoagpinto:
Sure… I will do it tonight or at 5am and tomorrow when both you and Pedro arrive at work I will have the rule fixed, and I will readd the temp_off.

Thanks for the feedback.

marcoagpinto:
@susanaboatto 

I have just fixed the rule.

I changed the rule like you suggested, changing the verbal form it shows as output.

Here is the resulting file:

[2.txt](https://github.com/user-attachments/files/16628114/2.txt)

susanaboatto (approved):

PR #10833, opened 2024-08-18
[pt] Added ENTITY colloquialism_msg to messages.ent

marcoagpinto:

I have added this message because of the dozens of pt-PT rules such as:

```
        <rule>
            <pattern>
                <token inflected='yes'>o</token>
                <token>nem-nem</token>
            </pattern>
            <message>Esta é uma expressão oral. Reveja.</message>
            <url>https://pt.wiktionary.org/wiki/a_nem-nem</url>
            <short>Coloquialismo</short>
            <example correction=''><marker>a nem-nem</marker></example>
        </rule>
```

Also, would it be a good idea for me to remove all `<short>` messages in pt-PT, since they are terrible? A person is revising a document, and it shows a suggestion with just one or two words which is completely unclear and not useful at all… I prefer longer explaining messages.

`<short>Coloquialismo</short>`


susanaboatto:
The short messages don't necessarily replace the longer message. They can coexist.

susanaboatto (approved):

PR #10836, opened 2024-08-20
[de] fix 'Fugpassagieren'

affemitkaraffe:


languagetool-language-modules/de/src/main/resources/org/languagetool/resource/de/verb_stems.txt
```diff
@@ -8215,7 +8215,7 @@ fräs
 frühstück
 fuchs
 fuchtel
-fug
+#fug
```
danielnaber, line 8218: Maybe also add a comment why this is problematic

PR #10845, opened 2024-08-27
[es] AI_ES_GGEC disable all punctuation marks at the sentence end

jaumeortola:


BellaDiekmann (approved):
Thanks for tackling this issue, Jaume. As far as I can tell (I'm not an expert on Java), it looks good.

PR #10861, opened 2024-09-03
[pt] disambiguation.xml fix

marcoagpinto:
Heya, @susanaboatto , @p-goulart and @jaumeortola ,

I have fixed rare verbs detecting:
```
Quero duas garotas.
Quero duas partes.
```

"garotas" and "partes" no longer appear as verbs.

But it will impact in the whole core of PT.

That is why I am always scared to change the disambiguator.

But if you believe it is okay, then it is a major improvement, and it will fix tons of false positives in our rules.

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
@@ -226,6 +226,15 @@
   </rule>
   <rulegroup id="RARE_POS" name="rare POS tags">
     <rule>
+      <pattern>
+        <token postag="Z0FP0" postag_regexp="no"/>
+        <marker>
+          <token postag="VMIP2S0" postag_regexp="no"/>
+        </marker>
+      </pattern>
+      <disambig action="remove" postag="V.*"/>
+    </rule>
+    <rule>
       <antipattern>
         <token>damos</token>
         <token postag="NC.+"/>


marcoagpinto:
Ahhhh... it now also fixes:
```
Quero as duas militares.
Quero os dois militares.
```
"militares" no longer appears as a verb.

marcoagpinto:
Sorry!!!!!!!

I have just fixed it:
`Para as duas matares.`

If the words are not also nouns, it won't remove them as verbs.

marcoagpinto:
Now it is working again with:
```
Quero duas garotas.
Quero duas partes.
Quero as duas militares.
Quero os dois militares.
Para as duas matares.
```

p-goulart (approved):
Seems okay to me.

marcoagpinto:
@p-goulart 

Pedro, I am so stressed!

😛 😛 😛 😛 😛 😛 😛 

Thanks!

p-goulart:
The rule seems tightly scoped, I doubt this will have many undesired side effects. But do keep an eye on the diff tomorrow morning.

@susanaboatto probably wiser not to include this in a premium deployment, if it does happen later today (I know there may be some changes to the schedule, but just in case.)

marcoagpinto:
Yes, at 5am I will look at the diff to check if some exceptions need to be added.

marcoagpinto:
@p-goulart @susanaboatto 

It worked better than expected:
https://internal1.languagetool.org/regression-tests/via-http/2024-09-03/pt-BR/
https://internal1.languagetool.org/regression-tests/via-http/2024-09-03/pt-PT/

With such a positive experience, it seems I will “play” more with the disambiguation from now on 😛 😛 😛 😛 😛 😛 😛 

I AM SO HAPPY!!!!!

PR #10868, opened 2024-09-08
[pt] Improved the disambiguation.xml

marcoagpinto:
Heya, @jaumeortola , @susanaboatto and @p-goulart ,

This enhancement fixes sentences like:

```
A equipe fundiu as duas como uma forma de meio-termo.
A equipe fundiu as suas duas como uma forma de meio-termo.
A equipe fundiu umas duas como uma forma de meio-termo.
```

“como” no longer appears as a noun nor as a verb.

It should be a low-impact change since it is a strict rule, but one must look at the nightly diffs.

😛 😛 😛 😛 😛 😛 😛 😛 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new disambiguation rule for the Portuguese language module to enhance grammatical accuracy by refining the handling of specific verb forms in context.

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
@@ -238,6 +238,17 @@
       <disambig action="remove" postag="V.*"/>
     </rule>
     <rule>
+      <pattern>
+        <token postag="V.+" postag_regexp="yes"/>
+        <token min='1' max='2' postag="[DP].+" postag_regexp="yes"/>
+        <token postag="Z0[FM]P0" postag_regexp="yes"/>
+        <marker>
+          <token postag="VMIP1S0" postag_regexp="no"/>
+        </marker>
+      </pattern>
+      <disambig action="remove" postag="[NV].*"/>
+    </rule>
+    <rule>
       <antipattern>
         <token>damos</token>
         <token postag="NC.+"/>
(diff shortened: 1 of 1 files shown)


languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
```diff
@@ -237,6 +237,17 @@
       </pattern>
       <disambig action="remove" postag="V.*"/>
     </rule>
+    <rule>
+      <pattern>
+        <token postag="V.+" postag_regexp="yes"/>
+        <token min='1' max='2' postag="[DP].+" postag_regexp="yes"/>
+        <token postag="Z0[FM]P0" postag_regexp="yes"/>
+        <marker>
+          <token postag="VMIP1S0" postag_regexp="no"/>
```
jaumeortola, line 246: If this rule is only for "como", you could write explicitly "como" there. It will be safer.

marcoagpinto: @jaumeortola 

I am not sure if it only works with “como”.

It also works with “logo”.

I based myself on the rule I was working on.

I was expecting to look at the night diff to see the results and later fix it in case there are false positives.

jaumeortola: Anyway, something is missing. If the word inside the `<marker>` is only verb (or verb and noun), all POS tags will be removed. I would add some other conditions: specific words (como, logo), or other POS tags (CS, PT, RG...)

marcoagpinto:
@jaumeortola 

Done!

languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
marcoagpinto: @jaumeortola 

I have done it hours ago.

Let me refresh the review.

jaumeortola (approved):

PR #10946, opened 2024-10-17
[fr] ignore adding punctuation at the sentence end

jaumeortola:
@GillouLT 
AI_FR_GGEC_MISSING_PUNCTUATION_PERIOD creates undesired matches, for example, in bulleted lists. We can rely on other rules, with proper exceptions, for this.


GillouLT (approved):
Sounds good! Thank you Jaume :)

PR #10947, opened 2024-10-17
Do not use unnecessary regular expressions

hurricup:
- Replaced replaceAll with replace wherever pure text is used instead of pattern. Since Java 9 it does not use patterns and works faster.
- Replaced literal patterns usages with simple replace
- Rplaced `replaceFirst` with constant patterns with `StringUtils.replaceOnce`from apache commons
- Special case is: `AbstractAdvancedSynthesizerFilter.getCompositePostag`seem the business-logic prevented from matching and `groupStr` being null. But technically it can be null and `replaceAll` first tries to find, then do something with argument, but `replace` first tries to do something with argument and fails if there is `null` even if we didn't find anything. So we need to check manually for null now.

## Summary by CodeRabbit

- **New Features**
	- Enhanced string replacement methods across various classes for improved performance and readability.
  
- **Bug Fixes**
	- Adjusted handling of specific characters and patterns in multiple language modules to streamline processing.

- **Refactor**
	- Replaced regex-based string manipulations with simpler string replacement methods in several classes, improving code clarity and efficiency.

- **Documentation**
	- Minor updates to comments and documentation for clarity across various classes.


hurricup:
I have [more adjustments](https://github.com/hurricup/languagetool/commit/620206f2fb94316cf564b8917b0623e3eb1d5b3c) but probably will submit with the next PR

hurricup:
@danielnaber sorry, i'm new here. Do I need to do anything else to be merged? Or just wait? Should I put additional adjustments into this PR or better to create another?

danielnaber:
@hurricup No, just wait for the reviewers to finish. Might take another week or so.

SteVio89 (approved):

fabrichter (approved):
LGTM, thanks!
Sorry for the delay

PR #10964, opened 2024-10-24
Feature/ext config

wernerjuelich:
## Summary by CodeRabbit

- **New Features**
	- Added support for environment variable configuration, allowing runtime adjustments for public access and port settings.
	- Updated default server binding to accept connections from any network interface.

- **Chores**
	- Introduced a new dependency for `lettuce-core` and a plugin for Azure deployment to enhance build configuration.
	- Minor formatting improvements in plugin configurations for consistency.

### DIFF
--- a/languagetool-server/pom.xml
+++ b/languagetool-server/pom.xml
@@ -67,11 +67,6 @@
             <version>6.5.0.RC1</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


danielnaber:
I must admit I'm not a fan of using env variables here. We already have command line arguments and a property file. Is there a use case where env variables are needed and the command-line parameters don't work?

PR #11027, opened 2024-11-16
Memory and performance optimizations

hurricup:
We did get a report that when the French language is active, a lot of memory consumed. I did investigate and there are a lot of things that could be optimized. I made some improvements and will split them into several PRs to make it easier to review.

In this PR:
- French language is a singleton (`getInstance` should be used to obtain it). API prevents you from the instantiating yourself.
- German language is a singleton (`getInstance` should be used to obtain it). API prevents you from the instantiating yourself.
- Fixed significant performance flaw when `Languages` instantiated all languages twice for no reason on class loading.

My intent is to do the same for all languages in LT. This improves the loading timing and reduce the memory footprint. Currently, LT has a lot of unnecessary duplicated objects while working.

Here is the synthetic test I am using for the snapshots:
```
class MeasureMemory {
  public static void main(String[] args) throws Exception {
    List<JLanguageTool> tools = new ArrayList<>();
    Arrays.asList(
      GermanyGerman.getInstance(),
      French.getInstance()
    ).forEach(lang -> {
      System.err.println("LOADING: " + lang);
      try {
        tools.add(loadLanguage(lang));
      } catch (IOException e) {
        throw new RuntimeException(e);
      }
    });

    HotSpotDiagnosticMXBean mxBean = ManagementFactory.newPlatformMXBeanProxy(ManagementFactory.getPlatformMBeanServer(),
      "com.sun.management:type=HotSpotDiagnostic",
      HotSpotDiagnosticMXBean.class);
    String file = "/home/hurricup/tmp/lt-heap-" + LocalDateTime.now().format(DateTimeFormatter.ofPattern("KK-mm-d.MM")) + ".hprof";
    mxBean.dumpHeap(file, true);

    System.out.println("Done: " + tools + ", " + file);
  }

  @NotNull
  private static JLanguageTool loadLanguage(Language language) throws IOException {
    language.setChunker(new NoopChunker());
    JLanguageTool tool = new JLanguageTool(language);
    tool.check("Hello Hello great great Dolly");
    return tool;
  }
}
```

Total results of all changes I made with (English, German, French):
Before:
image

After:
image

## Summary by CodeRabbit

- **New Features**
	- Introduced singleton access for `GermanyGerman` and `French` language instances across various components, improving resource management and consistency.
  
- **Bug Fixes**
	- Updated assertions in tests to reflect changes in expected outcomes, ensuring accuracy in language processing validations.

- **Documentation**
[...]

PR #11043, opened 2024-11-20
Revert "Memory and performance optimizations"

SteVio89:
Reverts languagetool-org/languagetool#11027 → caused issues with premium need to fix those first

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/Languages.java
+++ b/languagetool-core/src/main/java/org/languagetool/Languages.java
@@ -24,12 +24,12 @@ import org.languagetool.noop.NoopLanguage;
 import org.languagetool.tools.MultiKeyProperties;
 import org.languagetool.tools.StringTools;
 
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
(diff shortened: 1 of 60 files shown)


hurricup:
@SteVio89 how can I investigate this premium thing? And why there are no tests for it in common repo? Or isn't it possible to test without some close source code?

hurricup:
Maybe you could set up some CI I could use for testing? Still may be tricky to fix without tests and/or code. Btw, it is possible things that failing are those direct field accesses I replaced with getInstance

SteVio89:
@hurricup Unfortunately you cannot test this by yourself and this case also required the close source code.

I will update your code, hopefully by Friday. We have CI to tests against the premium, but only for our master branch, and the test results are only available with access to the premium branch.

Here is the new PR https://github.com/languagetool-org/languagetool/pull/11044 I will update this with my changes ASAP.

hurricup:
How do we proceed with the next changes? Because there are more like this one.

SteVio89:
Just create the PR as usual and I will check everything against premium before the merge :)

PR #11119, opened 2024-12-17
3d part of optimizations

hurricup:
This is the last bundle of changes I made back then.
Would be nice if you would not squash them. Each commit has its own semantics and would be nice to keep it that way

### DIFF
--- a/languagetool-language-modules/en/src/main/java/org/languagetool/language/BritishEnglish.java
+++ b/languagetool-language-modules/en/src/main/java/org/languagetool/language/BritishEnglish.java
@@ -19,12 +19,11 @@
 
 package org.languagetool.language;
 
-import java.io.IOException;
-import java.util.*;
-
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.languagetool.GlobalConfig;
 import org.languagetool.Language;
+import org.languagetool.Languages;
 import org.languagetool.UserConfig;
 import org.languagetool.languagemodel.LanguageModel;
 import org.languagetool.rules.Rule;
@@ -33,7 +32,27 @@ import org.languagetool.rules.en.MorfologikBritishSpellerRule;
 import org.languagetool.rules.en.UnitConversionRuleImperial;
 import org.languagetool.rules.spelling.SpellingCheckRule;
 
+import java.io.IOException;
+import java.util.*;
+
 public class BritishEnglish extends English {
+  private static final String LANGUAGE_SHORT_CODE = "en-GB";
+
+  private static volatile Throwable instantiationTrace;
+
+  public BritishEnglish() {
+    Throwable trace = instantiationTrace;
+    if (trace != null) {
+      throw new RuntimeException("Language was already instantiated, see the cause stacktrace below.", trace);
+    }
+    instantiationTrace = new Throwable();
+  }
+
+  /**
+   * This is a fake constructor overload for the subclasses. Public constructors can only be used by the LT itself.
+   */
+  protected BritishEnglish(boolean fakeValue) {
+  }
 
   @Override
   public String[] getCountries() {
@@ -85,5 +104,13 @@ public class BritishEnglish extends English {
       return prio;
     }
     return super.getPriorityForId(id);
+  }
+
+  public static @NotNull English getInstance() {
+    Language language = Objects.requireNonNull(Languages.getLanguageForShortCode(LANGUAGE_SHORT_CODE));
+    if (language instanceof English britishEnglish) {
+      return britishEnglish;
+    }
+    throw new RuntimeException("BritishEnglish language expected, got " + language);
   }
 }
--- a/languagetool-language-modules/en/src/main/java/org/languagetool/language/CanadianEnglish.java
+++ b/languagetool-language-modules/en/src/main/java/org/languagetool/language/CanadianEnglish.java
@@ -19,15 +19,11 @@
 
 package org.languagetool.language;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.ResourceBundle;
-
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.languagetool.GlobalConfig;
 import org.languagetool.Language;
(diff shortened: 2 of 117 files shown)


hurricup:
Side question. Did you consider introducing kotlin into the project?

languagetool-language-modules/en/src/main/java/org/languagetool/language/BritishEnglish.java
```diff
@@ -86,4 +105,12 @@ protected int getPriorityForId(String id) {
     }
     return super.getPriorityForId(id);
   }
+
+  public static @NotNull BritishEnglish getInstance() {
+    Language language = Objects.requireNonNull(Languages.getLanguageForShortCode(LANGUAGE_SHORT_CODE));
+    if (language instanceof BritishEnglish britishEnglish) {
```
SteVio89, line 111: This will not work with the close source part :( See my comments in German: https://github.com/languagetool-org/languagetool/blob/482657017ccde7123292c801ef9f332a4f6c9e80/languagetool-language-modules/de/src/main/java/org/languagetool/language/GermanyGerman.java#L90

This applies for those languages and their variants:

de
en
es
pt
fr
nl

languagetool-language-modules/en/src/main/java/org/languagetool/language/CanadianEnglish.java
```diff
@@ -67,4 +83,11 @@ public List<Rule> getRelevantLanguageModelCapableRules(ResourceBundle messages,
     return rules;
   }
 
+  public static @NotNull CanadianEnglish getInstance() {
+    Language language = Objects.requireNonNull(Languages.getLanguageForShortCode(LANGUAGE_SHORT_CODE));
+    if (language instanceof CanadianEnglish canadianEnglish) {
```
SteVio89, line 88: This will not work with the close source part :( See my comments in German: https://github.com/languagetool-org/languagetool/blob/482657017ccde7123292c801ef9f332a4f6c9e80/languagetool-language-modules/de/src/main/java/org/languagetool/language/GermanyGerman.java#L90

This applies for those languages and their variants:

de
en
es
pt
fr
nl

languagetool-language-modules/pt/src/main/java/org/languagetool/language/BrazilianPortuguese.java
```diff
@@ -54,4 +75,11 @@ public String[] getCountries() {
     return new String[]{"BR"};
   }
 
+  public static @NotNull BrazilianPortuguese getInstance() {
+    Language language = Objects.requireNonNull(Languages.getLanguageForShortCode(LANGUAGE_SHORT_CODE));
+    if (language instanceof BrazilianPortuguese brazilianPortuguese) {
```
SteVio89, line 80: This will not work with the close source part :( See my comments in German: https://github.com/languagetool-org/languagetool/blob/482657017ccde7123292c801ef9f332a4f6c9e80/languagetool-language-modules/de/src/main/java/org/languagetool/language/GermanyGerman.java#L90

This applies for those languages and their variants:

de
en
pt

languagetool-language-modules/pt/src/main/java/org/languagetool/language/PortugalPortuguese.java
```diff
@@ -86,4 +101,12 @@ public String getOpeningDoubleQuote() {
   public String getClosingDoubleQuote() {
     return "»";
   }
+
+  public static @NotNull PortugalPortuguese getInstance() {
+    Language language = Objects.requireNonNull(Languages.getLanguageForShortCode(LANGUAGE_SHORT_CODE));
+    if (language instanceof PortugalPortuguese portugalPortuguese) {
```
SteVio89, line 107: This will not work with the close source part :( See my comments in German: https://github.com/languagetool-org/languagetool/blob/482657017ccde7123292c801ef9f332a4f6c9e80/languagetool-language-modules/de/src/main/java/org/languagetool/language/GermanyGerman.java#L90

This applies for those languages and their variants:

de
en
pt

hurricup: Dam, forgot about that. sorry

hurricup:
force-pushed updated version with proper instance checking. the rest is the same

languagetool-language-modules/pt/src/main/java/org/languagetool/language/PortugalPortuguese.java
hurricup: fixed

languagetool-language-modules/en/src/main/java/org/languagetool/language/CanadianEnglish.java
hurricup: fixed

languagetool-language-modules/en/src/main/java/org/languagetool/language/BritishEnglish.java
hurricup: fixed

hurricup:
fixing

hurricup:
Fixed, local tests passed. sorry for the noise

SteVio89:
@hurricup thank you! I will update our close source and I think I can merge this tomorrow.

hurricup:
Could you please clarify what is this metric about? Maybe you could figure out some synthetic test to reproduce the issue so I could look into snapshots on my side?

SteVio89:
We measure how long it takes to handle a textChecking request (percentile P95).

This is just a rough indicator for us, but if it increases at the same time as the general server load and directly after deployment, it is almost always a fairly reliable indicator that something went wrong.

Unfortunately, I can't give you a test with which we can reproduce this well, but I just located another recursive update issue in English while debugging. I am currently in the process of adapting the implementations of the PartialPosTagFilter to the respective languages.

I am not sure if this could be partly responsible for the problems.

The cause lies between these commits:

0c70a6d...34183a0

It is also possible that the issue came from the closed code part. I will do some more tests today.

hurricup:
Let me know If you won't find the cause, I'll take a look.

hurricup:
Did you manage to find the issue? 
Also I'm curious if you have a memory-related metrics and what do they say after changes.

Issue #11189, opened 2025-01-15
[en]: False Friends : Criminal rule issue 


pelfins:
There is an issue with the following rule suggesting a change from 'criminal' to 'kryminalista' which, is far as I understand, is a Polish word. It certainly isn't in use in common British English. If the word 'kyryminalista' is used, LanguageTool correctly identifies this as a potential spelling mistake when set to English:

Category: FALSE_FRIENDS
Rule: CRIMINAL

```text
**False friend**

Hint: "criminal" (English) means "kryminalista" (English). Did you mean "kryminalista"?
```

**file**: `languagetool-core/src/main/resources/org/languagetool/rules/false-friends.xml`

```xml
<rulegroup id="CRIMINAL">
        <rule>
            <pattern lang="en">
                <token>criminal</token>
            </pattern>
            <translation lang="pl">przestępca</translation>
            <translation lang="en">kryminalista</translation>
        </rule>
```

pelfins:
Additionally, I've noticed that this only seems to be occurring through the API with the Obsidian plugin. I have checked the plugin settings, and they seem to be correct. The issue doesn't happen within Notebook or within the LT editor.

danielnaber:
Can you set your native language in the Obsidian add-on? These kinds of rules should only be active if LT gets a parameter with your native language. This is what I get for an English text when I set my native language to `pl`: 

> Message: Wskazówka: “criminal” (English) oznacza “przestępca” (Polish) Czy chodzi o “whodunit”, “prison”?

Would this make more sense?

pelfins:
Thank you for getting back to me. Yes, native language is set as `en-GB` within the add-on. 

I'm confused by the suggestion: 

> Hint: "criminal" (English) means "kryminalista" (English). Did you mean "kryminalista"?

As _kryminalista_ certainly isn't an English word, yet is being suggested as such, I'm wondering if the correction hint should be criminalist (which is quite an outdated English word) although the _internet_ tells me it is a translation of the Polish: kryminalista?

I also don't fully understand the `xml` file, but this line:

> `<translation lang="en">kryminalista</translation>`

This implies to me that the `<token>criminal</token>` word being checked by LT, has a potential false-friend, in English, of kryminalista? 

I could be fundamentally wrong about all of this, and it could of course just be something to do with my set-up!

Thank you :)

danielnaber:
[...]

PR #11194, opened 2025-01-16
fix more cases as the one in #11189 and add check to prevent this issue

danielnaber:
## Summary by CodeRabbit

- **New Features**
	- Added new false friend translation rules between English and multiple languages (French, Spanish, Portuguese, German, Polish, Italian, Galician, and Russian).
	- Introduced an option for test mode in the false friend rule handling, enhancing error handling and flexibility.

- **Bug Fixes**
	- Improved validation for translation language configurations to prevent incorrect language settings in false friend rules.

- **Tests**
	- Added a new test class to validate the syntax of the false friend rules XML file.

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/patterns/FalseFriendRuleHandler.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/patterns/FalseFriendRuleHandler.java
@@ -96,6 +96,9 @@ class FalseFriendRuleHandler extends XMLRuleHandler {
       if (Languages.isLanguageSupported(languageStr)) {
         Language tmpLang = Languages.getLanguageForShortCode(languageStr);
         currentTranslationLanguage = tmpLang;
+        if (currentTranslationLanguage == language) {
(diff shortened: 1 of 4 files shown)


languagetool-core/src/main/java/org/languagetool/rules/patterns/FalseFriendRuleHandler.java
```diff
@@ -96,6 +96,9 @@ public void startElement(String namespaceURI, String lName,
       if (Languages.isLanguageSupported(languageStr)) {
         Language tmpLang = Languages.getLanguageForShortCode(languageStr);
         currentTranslationLanguage = tmpLang;
+        if (currentTranslationLanguage == language) {
+          throw new RuntimeException("Translation language (" + currentTranslationLanguage + ") must not be the same as pattern language (" + language + ") for rule " + id);
+        }
```
SteVio89, line 101: Wouldn't it be better to test this in a test instead of throwing a RuntimeException (which prevents the server from starting)?

danielnaber: Yes, let me try it that way

SteVio89 (approved):

Issue #11295, opened 2025-03-19
Incorrect Handling of Spaces Before ".NET"


luizcarlosfaria:
### Description:
When the term ".NET" appears in text, the spellchecker incorrectly identifies the preceding space as an error, suggesting the removal of the space. This causes it to merge the previous word with ".NET" incorrectly. For example, "technology .NET" gets flagged, and the suggestion is to change it to "technology.NET".

The correct behavior is to recognize ".NET" as a proper noun, allowing a space before it without flagging errors or suggesting incorrect fusions.

### Expected Behavior:
The term ".NET" should be recognized as a standalone proper noun, and the spellchecker should not suggest removing the space before it.

### Current Behavior:
A space before ".NET" is flagged as incorrect, prompting the spellchecker to suggest combining the preceding word with ".NET".

Steps ### to Reproduce:

Type or paste a sentence containing a space before the term ".NET" into the editor.

Example: "I work with the technology .NET and Azure."

Run the spellchecker or grammar-checker.

Observe the incorrect suggestion to remove the space, combining "technology" and ".NET".

**Suggested Fix:**
Update the parsing rules or add an exception to recognize ".NET" as a proper noun, allowing a space before it without prompting incorrect fusion suggestions.

vinodjayachandran:
PR for the fIX : https://github.com/languagetool-org/languagetool/pull/11322

luizcarlosfaria:
Hi, the error is about the tech name .NET and not a domain name.

Phrases to explain:

- Do you are new on .NET platform?
- Maybe choosing .NET and C# you do this better.
- The .NET ecosystem enable to develop web, mobile, microservices, windows desktop and cloud apps.

Best regards

PR #11322, opened 2025-04-08
Incorrect Handling of Spaces Before .NET

vinodjayachandran:
### Bug

https://github.com/languagetool-org/languagetool/issues/11295 

### Description
`CommaWhitespaceRule` allows file extensions and website domains after a full stop. `.net` is allow listed as part of existing website domains. Making regular expression case **insensitive** fixes the issues. 

**Note** : Although **.NET** in this context isn't really a domain,  instead of introducing a new category of allow listing, opted to re-purpose the existing rule for `.net`

### DIFF
--- a/languagetool-core/src/main/java/org/languagetool/rules/CommaWhitespaceRule.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/CommaWhitespaceRule.java
@@ -191,7 +191,7 @@ public class CommaWhitespaceRule extends Rule {
   }
 
   private boolean isDomain(AnalyzedTokenReadings[] tokens, int i) {
-    return i < tokens.length && tokens[i].getToken().matches("(com|org|net|int|edu|gov|mil|[a-z]{2})");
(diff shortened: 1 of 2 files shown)


vinodjayachandran:
> @vinodjayachandran even without your change the sentence
> 
> > I work with the technology .NET and Azure.
> 
> will not have a match from the `CommaWhitespaceRule`. This is because the `isFileExtension` will catch this case. But it will not catch the case:
> 
> > I work with the technology .Net and Azure.

Yes. With my change, it will catch  **I work with the technology .Net and Azure.**

I have verified the same with unit test as part of the PR ` assertMatches("I work with the technology .NET and Azure.", 0);`

co-odw:
Despite the test cases looking valid. I believe this commit doesn't achieve what it intends to. The reason is that the Sentence tokenizer will never consider `I work with the technology .NET and Azure.` as a sentence. It will only see `I work with the technology .`

PR #11484, opened 2025-08-15
[pt] Replaced the two previous rules with fixed logic

marcoagpinto:
The logic of the previous rules was wrong.

Now they work properly.

## Summary by CodeRabbit

* Bug Fixes
  * Improved Portuguese style checks for “duas/três vezes mais” and variants, with better handling of “do que/em/que,” punctuation boundaries, and flexible sentence patterns to reduce false positives and improve suggestions.
* Documentation
  * Updated examples to reflect expanded detection and clearer corrective guidance.
* Refactor
  * Renamed related Portuguese style rules (V2), with no change to default activation or user-facing availability.


marcoagpinto:
Now they give ZERO hits, which means the precision is 100% or near.

PR #11554, opened 2025-10-07
[pt] Major verb/noun fix in disambiguation.xml (must see nightly diff)

marcoagpinto:
Very high-impact fix!

I must pay close attention to the diffs tonight.

## Summary by CodeRabbit

* **New Features**
  * Improves Portuguese analysis by adding a disambiguation rule to prevent infinitive verbs from being mis-tagged as nouns (including plural cases). This reduces false positives and increases accuracy of grammar checks, agreement detection, and suggestions in complex sentences, so users will see more reliable results where infinitives and nouns might be confused.


marcoagpinto:
@jaumeortola 

I am producing a second test to make sure that the 'next' AQ.+ exception fixes "militar" and others.

If it breaks premium, I will have to ask for your help.

The idea of the coderabitai of adding a min='0' max='1' blah blah won't be implemented yet, first I must see the nightly results since it is a massive fix.

PR #11567, opened 2025-10-14
[en] Fix Java rule to avoid false positive

AnnaRusalkina:
I fixed the rule ENGLISH_WORD_REPEAT_RULE as it was incorrectly triggered when the proper name "[Bora Bora](https://www.tahiti.com/island/bora-bora)" is used. Additionally, I added a test case to the test file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected repeat-word detection to allow proper noun repetitions like “Bora Bora,” preventing false positives for legitimate names and phrases.

* **Tests**
  * Expanded test coverage with a case validating “I have visited Bora Bora.” to ensure consistent behavior and guard against regressions.


AnnaRusalkina:
Hi @AzadehSafakish ,
I have made an update to one English Java rule as it was producing a false positive. Could you review my pull request?

PR #11568, opened 2025-10-15
Fix: add lemma immunity in WordCoherencyRule to prevent false positiv…

minizhiren:
🧩 Summary

This PR fixes a false positive in the English Word Coherency Rule where certain inflected forms of words triggered unwanted "mixed spelling variant" warnings.

🐛 Problem

The rule incorrectly flagged these word:

doggies  
doggier  
doggiest

as inconsistent with the base forms doggy / doggie.
During testing (EnglishTest.testLanguage), this caused an assertion failure in the coherency.txt consistency check.

⚙️ Root Cause

The rule compared surface tokens directly against the variant map
without considering that an inflected word’s lemma (its base form)
might itself be one of the allowed variants.

As a result, legitimate inflections (e.g., “doggies”) were incorrectly treated as mixed variant usages.

🔧 Fix

Added a lemma-based immunity check in WordCoherencyRule to skip reporting
when a token’s lemma belongs to the same variant set defined in coherency.txt.

For example, if doggy and doggie are coherent variants,
then their lemmas are now correctly excluded from triggering false alarms for doggies, doggier, etc.

if (!Collections.disjoint(lemmas, variants)) {
    // lemma itself is one of the coherent variants → inflected form → skip
    continue;
}

🧠 Additional Notes

I also checked another repository (english-pos-dict)
and found that there are multiple word pairs similar to doggie/doggy
that could potentially cause the same issue.
However, only doggie was actually included in coherency.txt.

✅ Scope

Modified file: languagetool-language-modules/en/WordCoherencyRule.java

No core or cross-language code was changed.

This fix only affects the English module.

🧪How to Reproduce 

mvn edu.illinois:nondex-maven-plugin:2.2.1:nondex     -Dtest=org.languagetool.rules.en.EnglishTest#testLanguage     -DnondexMode=ONE -DnondexSeed=933178     -Dsurefire.failIfNoSpecifiedTests=false

## Summary by CodeRabbit

* **New Features**
  * Improved detection of inconsistent word variants across sentences with cross-sentence consistency checks and actionable replacement suggestions.

* **Bug Fixes**
  * Reduced false positives in variant checks for more accurate alerts.

* **Improvements**
  * Enhanced messaging and smarter replacement suggestions to make corrections faster and clearer.

### DIFF
--- a/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/WordCoherencyRule.java
+++ b/languagetool-language-modules/en/src/main/java/org/languagetool/rules/en/WordCoherencyRule.java
@@ -1,26 +1,9 @@
-/* LanguageTool, a natural language style checker 
- * Copyright (C) 2005 Daniel Naber (http://www.danielnaber.de)
- * 
- * This library is free software; you can redistribute it and/or
- * modify it under the terms of the GNU Lesser General Public
- * License as published by the Free Software Foundation; either
- * version 2.1 of the License, or (at your option) any later version.
- *
- * This library is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
- * Lesser General Public License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


minizhiren:
@coderabbitai generate docstrings

PR #11574, opened 2025-10-15
[en] Add antipatterns and change the message in rules

AnnaRusalkina:
I added new antipatterns where I saw the rules were giving false positives. Additionally, I updated a message for 1 rule as it sounded irrelevant.
List of modified rules:

- EN_MULTITOKEN_SPELLING_TWO
- DT_PRP
- UKRAINE_NO_THE
- COMMA_COMPOUND_SENTENCE_2
- YOUR

## Summary by CodeRabbit

* New Features
  * Added targeted grammar checks: prohibition of "the Ukraine", a pattern for determiner+pronoun misuse with clearer guidance, detection for negation + past participle + trailing adverb patterns, and a check for phrases like "your meanings?"
  * Added several multiword proper-name entries (e.g., multi‑token organization names)

* Bug Fixes
  * Reduced false positives by introducing antipatterns and refining rule messaging and short descriptions

### DIFF
--- a/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/multiwords.txt
+++ b/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/multiwords.txt
@@ -8206,3 +8206,5 @@ Grigory Margulis	NNP
 Vladimir Drinfeld	NNP
 Maryna Viazovska	NNP
 Ahmed al-Sharaa	NNP
+NN Group	NNP
+IKS Health	NNP


languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -139195,6 +139212,14 @@ The accident victim died from her injuries.
             </rule>
         </rulegroup>
         <rulegroup id="EN_MULTITOKEN_SPELLING_TWO" name="Spelling mistakes in proper nouns (2 tokens)">
+            <antipattern>
+                <token>NN</token>
+                <token regexp="yes">Group('s)?</token>
+            </antipattern>
```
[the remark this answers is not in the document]
AnnaRusalkina: Here <token>NN</token> should actually match a string "NN" as it is part of a proper noun "[NN Group](https://www.nn-group.com/)"

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -109829,6 +109836,12 @@ The accident victim died from her injuries.
         </rulegroup>
         <rulegroup id="COMMA_COMPOUND_SENTENCE_2" name="comma between independent clauses">
             <url>https://languagetool.org/insights/post/types-of-sentences/#compound-sentence</url>
+            <antipattern>
+                <token regexp="yes">have|has|had</token>
+                <token regexp="yes" min="0">not|n't|n’t</token>
+                <token postag="VBN" skip="-1"/>
+                <token postag_regexp="yes" postag="RB.*">yet</token>
```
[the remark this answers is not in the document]
AnnaRusalkina: In documentation of LanguageTool I see the example of `skip` negative value: https://dev.languagetool.org/development-overview

AnnaRusalkina:
Hi @AzadehSafakish ,
I have made updates to English rules in grammar.xml so they produce fewer false positives. Could you review my pull request?

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -130142,6 +130155,10 @@ The accident victim died from her injuries.
         <rulegroup id="YOUR" name="your (you're)">
             <url>https://languagetool.org/insights/post/your-vs-youre/</url>
             <rule>
+                <antipattern>
+                    <token>your</token>
+                    <token regexp="yes">meanings?</token>
+                </antipattern>
```
AzadehSafakish, line 130164: I would change this to:
```
<antipattern>
  <token>your</token>
  <token regexp="yes">meanings?</token>
  <token><exception>to</exception></token>
</antipattern>
```

to ensure an example like "If your (-> you're) meaning to do that, you should try doing it the other way" is still matched.
The antipattern as written eliminates this true positive.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -139195,6 +139212,14 @@ The accident victim died from her injuries.
...
         </rulegroup>
         <rulegroup id="EN_MULTITOKEN_SPELLING_TWO" name="Spelling mistakes in proper nouns (2 tokens)">
+            <antipattern>
+                <token>NN</token>
+                <token regexp="yes">Group('s)?</token>
+            </antipattern>
+            <antipattern>
+                <token>IKS</token>
+                <token regexp="yes">Health('s)?</token>
+            </antipattern>
```
AzadehSafakish: I think these two antipatterns will work, but generally we want to avoid adding antipatterns for specific words to this rule and instead add them to [multiwords.txt](https://github.com/languagetool-org/languagetool/blob/master/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/multiwords.txt) (only use this file for terms that are 2 tokens or more).

Benefits:
- Not crowding this rule with endless antipatterns
- We add the POS tag to our tagger (you can find the tagset [here](https://github.com/languagetool-org/languagetool/blob/master/languagetool-language-modules/en/src/main/resources/org/languagetool/resource/en/tagset.txt); these would both be `NNP`)
- You won't have to add/specify the possessive inflection ("'s").

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
```diff
@@ -109829,6 +109836,12 @@ The accident victim died from her injuries.
         </rulegroup>
         <rulegroup id="COMMA_COMPOUND_SENTENCE_2" name="comma between independent clauses">
             <url>https://languagetool.org/insights/post/types-of-sentences/#compound-sentence</url>
+            <antipattern>
+                <token regexp="yes">have|has|had</token>
+                <token regexp="yes" min="0">not|n't|n’t</token>
+                <token postag="VBN" skip="-1"/>
+                <token postag_regexp="yes" postag="RB.*">yet</token>
+            </antipattern>
```
AzadehSafakish, line 109845: As a general note:

For more complex antipatterns, it's helpful to include an example of the false positive(s) that you want to target in the antipattern.
This makes it a lot easier to adjust the antipattern in case it's too broad and eliminates too many true positives.

AzadehSafakish (changes requested):

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
AnnaRusalkina: Okay, thanks, I will correct it.

languagetool-language-modules/en/src/main/resources/org/languagetool/rules/en/grammar.xml
AnnaRusalkina: Okay, I will add now the examples I tested it on.

AzadehSafakish (approved):

PR #11666, opened 2025-11-23
[pt] Removed rule impossible in the next disambiguator fix

marcoagpinto:
This rule and example are impossible after I fix the verbs that should be nouns (next commit).

There is no other way but removing them, since they don't work in correct Portuguese (they are nouns, not verbs).

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed a Portuguese style rule that produced incorrect suggestions for constructions like "fazer ... para ...", reducing false-positive rewrite hints.
  * Adjusted Portuguese disambiguation to be more precise for feminine noun/adjective forms and to limit which verb tags are altered, improving detection accuracy and fewer wrong verb disambiguations.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
+++ b/languagetool-language-modules/pt/src/main/resources/org/languagetool/resource/pt/disambiguation.xml
@@ -4083,7 +4083,7 @@ USA
       -->
   </rule>
 
-  <rule id="VERB_VERBVMIP_20251121" name="It is NC.+/AQ.+ not VMIP.S0">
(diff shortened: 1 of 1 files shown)


marcoagpinto:
@jaumeortola 

This is a major fix, fixing many nouns that appeared as verbs as well.

I had to remove a rule that would be impossible to work with the correct morphological information.

If it breaks Premium, I will need your help.

Thanks!

marcoagpinto:
@jaumeortola 

Ahhhhh... tomorrow I will make it work also with male nouns/adjectives.

I just noticed I forgot to fix that.

PR #11720, opened 2025-12-16
[nl] Fixed decapitalization of months and religion names

inesakochur:
WEEKDAGEN_MAANDEN: added missing months to the rule pattern
check_case.txt: added missing religion names, as they need to be corrected to lowercase letter

## Summary by CodeRabbit

* **Improvements**
  * Dutch month recognition expanded to cover all 12 months, improving detection of month-related capitalization issues.
  * Added eight Dutch religious terms to case-checking rules for more accurate capitalization handling.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


inesakochur:
@mark-baas can you review my PR

mark-baas:
Hi Inesa, I'd probably leave the rule WEEKDAGEN_MAANDEN as is, because adding those days and months will add more false positives. For example Goede Vrijdag, which is correct with a capital, will get excluded if we add vrijdag in the regex. The other file looks good though!

inesakochur:
> Hi Inesa, I'd probably leave the rule WEEKDAGEN_MAANDEN as is, because adding those days and months will add more false positives. For example Goede Vrijdag, which is correct with a capital, will get excluded if we add vrijdag in the regex. The other file looks good though!

I think it's related mostly to Vrijdag, as I've seen the tests failed. But those are quiet safe to add: Oktober|November|December.
So I can remove it from the regex formula.

PR #11844, opened 2026-02-11
[en] fix HAVE_CD_YEARS

AnnaRusalkina:
There was a problem that when you wrote "I have 25 years old", LanguageTool suggested "I are 25 years old", which is incorrect. So, I added an antipattern to the original rule, so it doesn't match "I" and instead added a separate rule for "I" in present.


AnnaRusalkina:
Hi @AzadehSafakish ,
Could you review my pull request and merge it?

PR #12022, opened 2026-06-06
Bump opennlp-tools from 1.9.4 to 2.5.9 to fix CVEs

gt37csz:
## Summary
- Updates Apache OpenNLP (`opennlp-tools`) from `1.9.4` to `2.5.9` to fix
  CVE-2026-42027, CVE-2026-40682 (Critical) and CVE-2026-42440 (High).
- Pins `POSTagFormat.PENN` in `EnglishChunker`: since OpenNLP 2.x, `POSTaggerME`
  maps the bundled model's native Penn Treebank tags to Universal Dependencies
  by default, which would otherwise break the chunker and all Penn-based English rules.

## Compatibility notes / request for guidance (RFC)
- The bundled 1.5-era models (`en-token.bin`, `en-pos-maxent.bin`,
  `en-chunker.bin` from `edu.washington.cs.knowitall`) still load fine under
  2.5.9 - no `InvalidFormatException`.
- With the `POSTagFormat.PENN` fix, POS-tagging output is byte-identical to 1.9.4.
- However, OpenNLP 2.x's `ChunkerME` produces slightly different chunk
  boundaries for the **same model and identical Penn POS input** (e.g.
  coordinated adjectives "X and Y" become `NP` instead of `ADJP`; comparative
  "as" becomes `SBAR` instead of `PP`). The default beam size is identical (10)
  in both versions, so this is an internal ML-decoding change and is not
  configurable.
- As a result, the embedded examples of ~18-19 chunk-dependent English rules
  change behavior (e.g. `SERIAL_COMMA_ON`, `IT_IS`, `IT_IS_2`,
  `COMPOUND_NNP_AGREEMENT`, `THERE_S_MANY`, `MD_BASEFORM`, `JJ_NN_JJ`,
  `AI_HYDRA_LEO_CP_YOUR_YOU`, ...).

I'd appreciate maintainers' guidance on the preferred path (retrain/replace the
chunker model vs. adjust the affected rules), since this touches rule quality.

## Test plan
- [x] `EnglishChunkerTest` passes with the `POSTagFormat.PENN` fix
- [ ] Full `language-en` test suite - currently fails on the chunk-dependent
      rules listed above (expected; see notes)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/languagetool-language-modules/en/src/main/java/org/languagetool/chunking/EnglishChunker.java
+++ b/languagetool-language-modules/en/src/main/java/org/languagetool/chunking/EnglishChunker.java
@@ -21,6 +21,7 @@ package org.languagetool.chunking;
 import opennlp.tools.chunker.ChunkerME;
 import opennlp.tools.chunker.ChunkerModel;
 import opennlp.tools.postag.POSModel;
+import opennlp.tools.postag.POSTagFormat;
 import opennlp.tools.postag.POSTaggerME;
 import opennlp.tools.tokenize.TokenizerME;
 import opennlp.tools.tokenize.TokenizerModel;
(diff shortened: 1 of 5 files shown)


jaumeortola:
Are there only  "~18-19 chunk-dependent English rules"? Post the complete list. I thought there were more.

PR #12059, opened 2026-07-07
[de] fix FEHLERHAFTES_KOMMA_ALLG rule

AnnaRusalkina:
The rule (FEHLERHAFTES_KOMMA_ALLG) produced a false positive because the part-of-speech tagger incorrectly classified the word _strukturierte_ as a past participle verb rather than as part of a coordinated noun phrase. I added an antipattern for such cases and an example.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved comma-checking in German grammar rules to better avoid false positives after a colon in specific list-like constructions.


AnnaRusalkina:
Hi @jaumeortola 
Could you merge my pull request? @inesakochur has reviewed and tested my PR.

PR #12064, opened 2026-07-09
[core] make regexp and SRX patterns Unicode-aware for JDK 19/21 migration

jaumeortola:
Migrating the build/runtime from JDK 17 to JDK 19/21. Between those JDK versions, the behavior of \b, \w, \d and \s next to non-ASCII characters changed unless the pattern is compiled Unicode-aware (UNICODE_CHARACTER_CLASS / the inline (?U) flag). Without it, many rules and the sentence segmentation behave differently across JDKs. The migration forces Unicode-aware matching everywhere, and then fixes the one place where that new \s semantics has an unwanted side effect.

1. **PatternRuleHandler.java** XML <regexp> rules are now compiled with Pattern.UNICODE_CHARACTER_CLASS added to the flags. This makes \b/\w/\d/\s Unicode-aware for every <regexp> rule, so word-boundary matching next to non-ASCII letters is stable across JDKs — without having to add (?U) by hand to each pattern.

2. **SrxTools.java** Updated `loomchild.segment` to 2.0.4,  and the patterns are compiled with UNICODE_CHARACTER_CLASS. This applies the same Unicode-aware matching to the thousands of sentence-tokenizer patterns without editing them individually.

3. **segment.srx** Side-effect fix for the above. Making \s Unicode-aware means it now also matches \u00A0 (non-breaking space). The ByTwoLineBreaks paragraph rule \r?\n\s*\r?\n[\t]* then greedily swallowed \u00A0 and merged paragraph breaks that must stay separate. That rule now uses an explicit ASCII-whitespace class ([ \t\n\x0B\f\r]) instead of \s, so paragraph-break detection is unaffected by the new Unicode-aware \s. Everywhere else in segment.srx, the broader \s (including \u00A0) is intentional. This restores SentenceRangeTest.testSpecialCase (and keeps all 5 SentenceRangeTest cases green).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Unicode-aware regex compilation for more consistent Unicode boundary behavior across pattern matching and tokenization.
  * Updated sentence/segment SRX rules so whitespace around line breaks and in split/abbreviation contexts is handled more precisely.
  * Tightened multiple language grammar and pattern rules to treat optional whitespace as space/tab rather than broader whitespace.
* **Documentation**
  * Corrected a Javadoc typo in the sentence tokenizer.
* **Chores**
  * Bumped the managed `segment` dependency version (2.0.3 → 2.0.4).

### DIFF
--- a/languagetool-core/src/main/resources/org/languagetool/resource/segment.srx
+++ b/languagetool-core/src/main/resources/org/languagetool/resource/segment.srx
@@ -4641,7 +4641,7 @@
 </languagerule>
 <languagerule languagerulename="ByTwoLineBreaks">
 <rule break="yes">
-<beforebreak>\r?\n\s*\r?\n[\t]*</beforebreak>
+<beforebreak>\r?\n[ \t\n\x0B\f\r]*\r?\n[\t]*</beforebreak>
 <afterbreak></afterbreak>
 </rule>
 <rule break="yes">
--- a/languagetool-core/src/main/java/org/languagetool/rules/patterns/PatternRuleHandler.java
+++ b/languagetool-core/src/main/java/org/languagetool/rules/patterns/PatternRuleHandler.java
@@ -780,7 +780,11 @@ public class PatternRuleHandler extends XMLRuleHandler {
         rule.setDistanceTokens(distanceTokens);
         rule.setXmlLineNumber(xmlLineNumber);
       } else if (regex.length() > 0) {
-        int flags = regexCaseSensitive ? 0 : Pattern.CASE_INSENSITIVE|Pattern.UNICODE_CASE;
+        // UNICODE_CHARACTER_CLASS (equivalent to the inline (?U) flag) makes \b, \w, \d and \s
+        // Unicode-aware. Without it the behavior of word boundaries next to non-ASCII letters
+        // depends on the JDK version (it changed between JDK 17 and JDK 21), which breaks many
+        // <regexp> rules. Setting it here avoids adding (?U) by hand to every <regexp> pattern.
+        int flags = Pattern.UNICODE_CHARACTER_CLASS | (regexCaseSensitive ? 0 : Pattern.CASE_INSENSITIVE|Pattern.UNICODE_CASE);
         String regexStr = regex.toString();
         if (regexMode == RegexpMode.SMART) {
           // Note: it's not that easy to add \b because the regex might look like '(foo)' or '\d' so we cannot just look at the last character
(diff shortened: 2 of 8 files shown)


jaumeortola:
@milekpl Take a look at the solutions in this PR. They are more general than what you are doing. 
This also resolves the problem in SentenceRangeTest.java, which was what blocked the changes the last time we tried.

languagetool-core/src/main/resources/org/languagetool/resource/segment.srx
```diff
@@ -4641,7 +4641,7 @@
 </languagerule>
 <languagerule languagerulename="ByTwoLineBreaks">
 <rule break="yes">
-<beforebreak>\r?\n\s*\r?\n[\t]*</beforebreak>
+<beforebreak>\r?\n[ \t\n\x0B\f\r]*\r?\n[\t]*</beforebreak>
```
jaumeortola, line 4644: @SteVio89 This is what solves the problem in SentenceRangeTest.java. 
We exclude non-breaking spaces. Is that correct/enough?

milekpl:
Well, I will review it thoroughly tomorrow but in general, my fixes already make tests pass on JDK 22. It was also more critical for System.exit(1) behavior for CLI, but I would think that your approach with assuming Unicode for regex is essentially fine, For segment.srx, I tried, but some tests break then (in English I think?).

jaumeortola:
@SteVio89 Most of these changes have been done in the past. The only new thing is that the tests in SentenceRangeTest.java pass. Is that enough? Do we need more tests?

SentenceRangeTest.testSpecialCase() has non-breaking spaces. What should happen with other characters (white spaces, tabs, and so on) in a similar test?

milekpl:
I tested this and will remove some of my hacks in a sec

jaumeortola:
@milekpl We shouldn't have merged yet. I wanted to warn you, but I'm late.  

There may be issues with the server in production. We have done this before, and we had to revert the changes. That's why I wanted @SteVio89 to take a look.

milekpl:
@jaumeortola ok, sorry... But I tested this on several JDKs, and my changes before went beyond this (sentence tokenizer but also command line, which required more sandboxing). If the server in production works on JVM 17, then it should be fine. The only thing that is tricky is that maven has to be refreshed to pick up new segment lib.

jaumeortola:
@milekpl I think that we will be fine. The main issue was `SentenceRangeTest.testSpecialCase()`, and now it passes.

PR #12194, opened 2026-09-18
Fix enabledOnly=true returning no rules when enabledRules and enabledCategories are both given

xavivars:
## Summary

`enabledOnly=true` combined with **both** `enabledRules` and `enabledCategories` activates no rules at all. The API documentation says "only the rules and categories whose IDs are specified with `enabledRules` or `enabledCategories` are enabled", i.e. the union, and each parameter on its own behaves that way. Together they produce an empty rule set.

## Symptoms

Text `Us espera. Obre la sortida. Cal tenir-ho.`, language `ca-ES-valencia`:

| parameters | matches |
|---|---|
| `enabledRules=EXIGEIX_VOS,EXIGEIX_INFINITIUS_INDRE&enabledOnly=true` | `EXIGEIX_VOS`, `EXIGEIX_INFINITIUS_INDRE` |
| `enabledCategories=ADAPTADOR_VALENCIA&enabledOnly=true` | 2 × `ADAPTADOR_VALENCIA_LEXIC` |
| both lists `&enabledOnly=true` | **none** |
| both lists, no `enabledOnly` | all four, plus every default rule |

The same happens with the command line (`-e`, `--enablecategories`, `--enabledonly`), since it goes through the same `Tools.selectRules()`.

## Root cause

`Tools.selectRules()` applied `enabledOnly` in two independent passes:

1. After enabling the categories: disable every rule whose category is not in `enabledCategories`, by **full id** (`EXIGEIX_VOS[1]`, …).
2. After enabling the rules: disable every rule whose id is not in `enabledRules`, again by full id.

Pass 2 removes all rules of the enabled categories. And the rules enabled by id don't survive either: `enableRule("EXIGEIX_VOS")` only removes the bare id from the disabled set, while pass 1 had added `EXIGEIX_VOS[1]`, which `JLanguageTool.ignoreRule()` still honours. Every single-`<rule>` also gets sub id `1`, so this is not limited to rule groups. Net result: nothing is active.

## Fix

Rule selection is now a single pass: after applying the disabled/enabled categories and rule ids as before, if `enabledOnly` is set and at least one of the two lists is non-empty, a rule is kept when it is enabled by id (id or full id) **or** by category, and disabled otherwise. Behaviour with only one of the lists is unchanged, as is the case where both are empty.

## Tests

- `ToolsTest.testSelectRulesWithEnabledOnlyAndBothRulesAndCategories` (demo language): a rule and a rule group enabled by id plus a category enabled by id, with `enabledOnly=true`. Fails on master (empty result), passes with the fix; the existing `testSelectRules` cases still pass.
- Verified against the real server with the requests in the table above: the combined request now returns the four expected matches and the two single-list requests are unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved grammar rule selection when “enabled only” mode is active.
  - Rules selected individually or through an enabled category are now consistently retained.
  - Rules that are not selected are disabled as expected.
  - Explicitly disabled rules remain disabled, even when selected through another configuration.


jaumeortola:
Improved here: https://github.com/languagetool-org/languagetool/commit/aece4dadeff9c0764bdc25c8e5192c97691926cb

